PR #51, opened 2018-07-20
add critical error field and store import failures

GabAlexandria:
fix type and place calend columns
add comment to parsers


GabAlexandria:
@rultor Merge

PR #60, opened 2018-08-30
Feature/display neighbourhood turf - needs front end sanity

crantok:
The markup I added to display the neighbourhood turf is rubbish. Would someone like to take that on or tell me what to do?

* The neighbourhood turf is successfully retrieved (and displayed) for events that have a Place.
* I'll do another feature to add neighbourhood turf for other types of location. It looks like that will interact with the calendar events import, which is not touched here.
* The styleguide for the Event component is working despite what I said on the chat yesterday. I edited the yaml to use OpenStruct objects.


crantok:
@rultor merge

PR #62, opened 2018-09-06
Replace neighbourhood turfs with geocoded admin wards

crantok:
Replaces use of Google geocoding with postcodes.io.
Makes several knock-on changes as a result.
Up for discussion.


app/models/address.rb
```diff
@@ -30,18 +31,46 @@ def full_address
...
+    # our own code to query postcodes.io given that using postcodes.io
+    # now defines how we use Geocoder?
+    # - Passing anything other than a postcode will cause Geocoder.search to
+    #   return an empty array.
+    # - We are accessing admin_ward directly through through
+    #   Geocoder::Result::PostcodesIo#data hash.
+
+    geo = Geocoder.search(postcode).first&.data
+
+    if geo
```
kimadactyl: `return unless geo`

app/models/address.rb
```diff
@@ -30,18 +31,46 @@ def full_address
...
+      # TODO? This works. Is there a more correct (railsy) way of doing this
+      # e.g. using write_attribute ? What is the trade off?
+      self.longitude= geo['longitude']
```
kimadactyl: don't think you need `self` in these, looks fine to me tho!

app/models/address.rb
```diff
@@ -30,18 +31,46 @@ def full_address
...
+  end
+
   class << self
     # location - The raw location field
-    # components - Array containing parts of an event's location field
+    # components - Array containing parts of an event's location field, excluding the postcode.
     def search(location, components, postcode)
+
+      # Find the first Address whose first address line contains any one of the
+      # address lines in the components argument.
```
kimadactyl: I think find_by is the correct method to use here, could be wrong https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by

app/models/address.rb
```diff
@@ -30,18 +31,46 @@ def full_address
...
+      # We were looking for an exact match of geocoding coordinates, but we are
+      # now using postcodes.io exclusively so a postcode match is now
+      # equivalent to a coordinate match.
+      # if @address.blank?
+      #   coordinates = Geocoder.coordinates(postcode)
+      #   @address ||= Address.where(latitude: coordinates[0], longitude: coordinates[1]).first
+      # end
 
       # Search by postcode if it is minimum length of a full postal code
+      # TODO? Are postcodes guaranteed to be in the same format? E.g A1 2BC
```
kimadactyl: I am not sure, we prob need a unit test for this?

app/models/address.rb
crantok: Your suggestion fits well with the coding style I've seen in the app so far. I prefer to minimise method exit points but that's probably old skool now. I'll do what you said

app/models/address.rb
crantok: Although I still wonder whether this is the actual operation we want to perform. If the first line of the new address were "University of Manchester" then an existing address which started with "Skiving Dept, University of Manchester" would be found and used instead.

kimadactyl: Yeah, will need updating at some point. On the phone I suggested that maybe one thing Places can do it list an address as `canonical` - i.e. that any match will normalise to that one if found?

app/models/address.rb
crantok: Could this be a validation step? Could we validate that all postcodes are stored in the same format so that comparisons like this are always a know quantity?

kimadactyl: Yep that would be ideal. Only q would be is there a method to know where the break is for outputting in the frontend? Postcodes do look weird with no space.

crantok: According to the Market Research Society, all incodes (second part) are NAA, only the outcodes change in number of digits. https://www.mrs.org.uk/pdf/postcodeformat.pdf

app/models/address.rb
crantok: We still need to define 'match' though.

app/models/address.rb
kimadactyl: Pefecto then. Let's store space-free and add a presenter function to separate it out again.

app/models/address.rb
crantok: Is that tool widely used by the community? Although I don't like the suggested style in this case, I can see the benefit of the community tending towards a single style.

app/models/address.rb
kimadactyl: Lat/lon check I think right @GabAlexandria?

app/models/address.rb
kimadactyl: Tends to be, you can config it how you like, but there is a sound justification for most of these things. I think in this case it's that indentention just reduces readability, esp in increasingly nested blocks. maybe something like `return false unless geo` is clearer?

app/models/address.rb
crantok: Also, we can choose whatever format we like for storing to DB, so we can just use the preferred front-end format. We don't lose anything. (Except paying one byte per stored postcode that we could have saved :smile: ).

app/models/address.rb
kimadactyl: Ahhhhh IIRC now I think you do need self if you want to write (as opposed to read). I don't think I'm going totally mad!

app/models/address.rb
crantok: That was a separate check (now commented out) in the original code. Also, by using postcodes.io, everything in the same postcode will have the same latlon.

app/models/address.rb
kimadactyl: OK, will leave up to you, as long as its documented :)

app/models/address.rb
crantok: Aside:
The linux people used to mandate that coders set indents to 8 spaces. They said that if you had more than 2 levels of nesting then your code was unreadable anyway so having all slide off the side of the screen was good wake up call :laughing:

crantok:
I think that's resolved everything except for one thing:
 - Imported events with new addresses will grab the first address that has the same postcode. Is this what we want?

kimadactyl:
Yep. For now. Tagging @GabAlexandria for her feelings.

crantok:
@rultor merge

PR #65, opened 2018-09-11
Added filtering of Events by Site

crantok:
* The Event index on a valid local Site, e.g. hulme.placecal.org , should now only show those events within the primary and secondary Turfs for that Site.
* This should not have affected events filtering on Place or Partner pages.


crantok:
Post deployment steps:

1.  rake addresses:update_all_neighbourhood_turfs
1.  rake events:set_missing_addresses_from_place

crantok:
Ah, that makes a lot of sense. So presumably most "rails" commands are now implemented as rake tasks.

crantok:
@rultor merge

PR #68, opened 2018-09-17
Killed redundant and confusing Event#location

crantok:
Removed Event#location.
* The only place it was called from was the EventComponent UI class.
* Changed EventComponent to use Event#address directly.


crantok:
@rultor merge

PR #70, opened 2018-09-18
Refactor turfs -> neighbourhoods + (interest) turfs

crantok:
* Minor refactoring of get/set_home_turfs
* Addition of home_turf as named param to EventComponent
* Examples of home turf in partners/places index


crantok:
TODO

* Extract neighbourhood turfs from turfs table to a neighbourhoods table
* Modify sites_turfs to use neighbourhoods table
* Rename turfs table to tags (or something more platable)
* Push consequent name changes through the code.

kimadactyl:
Better create an ADR for this too

PR #77, opened 2018-09-25
Added map + address to Event page when the Event does not have a Place

crantok:


crantok:
There's something I'd like input on in this PR: My code interacts with code in
* app/fields

and
* app/views/fields

both of which are concerned solely with addresses. I have changed code in those directories in order to fix test failures, but I don't have a clear idea of what the 'fields' code is for and whether my changes are sufficient.

Are fields a rails thing? It's difficult to google cus ActiveRecord fields are referred to a lot and this is plainly not that.

If the fields stuff is legacy and we're doing it a different way for everything else then I would be very happy to strip out the legacy stuff.

crantok:
@rultor merge

PR #78, opened 2018-09-25
Removed erd.pdf from repo.

crantok:
This file is already ignored but had not been removed from the repo


crantok:
@rultor merge

PR #86, opened 2018-10-02
Fixed admin subdomain redirecting to default site.

crantok:
Any action that called current_site on the admin site could trigger a redirect. Have now prevented this.


crantok:
Abandoned. Made same changes in the Mossley front end PR.

PR #89, opened 2018-10-03
Select correct home page template

crantok:
Fixed selection of home template. Refactored site-selection.


crantok:
@rultor merge

PR #96, opened 2018-10-07
require place if same as place selected, set is_working to false if c…

GabAlexandria:
…ritical error


kimadactyl:
Move the checkbox for is_a_place to below the Places dialogue so it's a bit clearer :)

kimadactyl:
@rultor merge

PR #102, opened 2018-10-12
allow users to add their facebook app tokens, generate user tokens fr…

GabAlexandria:
…om app creds


GabAlexandria:
@rultor merge

PR #104, opened 2018-10-15
Removed Places. Added their features to Partners.

crantok:
This is the same as #101 except rebased onto recent master as a squash commit. Will merge this as soon as final query from #101 is resolved.


crantok:
Just noticed that there is a db migration that postdates the 3 that I added. How will that work when this gets pushed to staging. Should I rename my migrations to make them later than the one I merged from master?

crantok:
@rultor merge

PR #107, opened 2018-10-16
Removed places index. Only one Partner index now.

crantok:
Removed all place / partner distinction. All Partners in the relevant geographical region are shown on the partners page.


kimadactyl:
@rultor merge

PR #111, opened 2018-10-22
Added rake task to download prod db and restore either to local or st…

crantok:
…aging.

Servers and db names are hard coded. This makes invocation easier but may
not be the best approach.

After implementing, realised that I could probably have reused existing rake
tasks rather than shelling out, although that would only have saved a small
amount of effort.

This is the main task:

rails db:dump_production_and_restore_other  # Download production DB dump and optionally use it to restore_on_local=1 and/or restore_on_staging=1

If its components are required, e.g. because one of them fails,
they can be called individually:

rails db:dump_production                    # Download production DB dump
rails db:restore_local                      # Restore db dump file db_dump_filename=[filename] to local dev DB
rails db:restore_staging                    # Restore db dump file db_dump_filename=[filename] to staging server DB

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -134,6 +134,67 @@ namespace :db do
     end
   end
 
+  DB_DUMP_ENV_KEY = 'db_dump_filename'
+
+  desc "Download production DB dump"
+  task dump_production: :environment do
+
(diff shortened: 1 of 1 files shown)


crantok:
@rultor

crantok:
@rultor merge

PR #127, opened 2018-11-01
Added contact info UI to admin calendar form.

crantok:
- Removed "Event manager" from admin partner form.
 - Changed calendars' contact columns from text to string.


crantok:
TODO: With the addition of the contact details fields to the calendar admin form, the various buttons are now in two different locations that don't really make sense in mobile view. It would take more brain power than I currently have to do something sensible about this.

crantok:
@rultor merge

PR #306, opened 2019-10-25
allow partners to edit themselves

GabAlexandria:


kimadactyl:
Nice changes, this might be a good time to put in some basic policy unit tests too - should theoretically catch a lot of issues?

PR #320, opened 2019-11-03
Migrate js to webpacker

kimadactyl:
Doing stylesheets and assets too is a total pain - too much faff


Gemfile
```diff
@@ -33,8 +33,9 @@ gem 'carrierwave'
 gem 'mini_magick'
 
 # Admin
-gem 'bootstrap-sass', '~> 3.4.1'
+gem 'bootstrap', '~> 4.3.1'
 gem 'cocoon'
+gem 'font-awesome-rails'
```
kimadactyl: Why font awesome?

PR #334, opened 2019-11-22
WIP: Begin work on improving form inputs for partners

kimadactyl:
TODO: Just gonna add these here from the Trello card so it's all in one place!

- [x] Store twitter handles without a @ with a before_save hook
- [x] Make sure error readouts are clear and legible
- [x] Update main site interface to reflect changes
- [x] Create database migration for existing facebook and twitter links
- [x] Add in a tel: linkto frontend if its a valid phone number but otherwise let's leave this alone
- [x] Force postcodes to be validated in the admin frontend and don't allow ones that don't pass to be stored
- [x] Get rid of "existing address" dropdown, instead just populate the new address form so it looks seamless to the end user


kimadactyl:
Just gonna add these here from the Trello card so it's all in one place!

- [ ] Store twitter handles without a @ with a before_save hook
- [ ] Make sure error readouts are clear and legible
- [ ] Update frontend to reflect changes
- [ ] Create database migration for existing stuff
- [ ] Add in a tel: link if its a valid number

PR #353, opened 2019-12-08
always redirect to admin on sign in

GabAlexandria:


kimadactyl:
Looks like an integration test is failing, can you add in one for this route too?

PR #450, opened 2020-01-15
Update user permissions

GabAlexandria:


kimadactyl:
Looks great, can you just add a test for each bullet point on the ticket? :)

PR #778, opened 2021-11-01
Create test.yml

alexandria-gfsc:
Committing without secrets because to be honest I'm not 100% sure what the secrets tarball contains?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
           bundler-cache: true # Automagically runs bundle install
       - name: Setup SQL
         run: |
-          psql -h postgres -d postgres_db -U postgres_user -c 'create database placecal_test;'
+          psql -h localhost -d postgres_db -U postgres_user -c 'create database placecal_test;'
         env:
           PGPASSWORD: foobar
       - name: Do Tests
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on: [push, pull_request, workflow_dispatch]
+
+# According to: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
+# Both Node and Yarn should be installed already
+# This would ideally be split up into dependent jobs maybe, but this is a direct conversion of the travis script for now
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
kimadactyl, line 11: maybe lets set this to the same version as server

alexandria-gfsc: Good call, I wasn't sure if there's a way to do this! But after doing a search for other people's code it looks like there is :)

alexandria-gfsc:
Everything works! The deploy isn't tested ATM so that needs to be done, which can only be done by merging with master :smiling_imp:

alexandria-gfsc:
Confirmed working on kim's machine too!

PR #796, opened 2021-11-18
Neighbourhoods (Resolves #791)

### LINKED ISSUE #791: Convert existing neighbourhood table to new structure
1. Pseudocode for new database structure using `ancestry` below. Note that we will just use existing data for now before we seed the UK data.
1. Make sure all tests pass. Might take a bit of a sweep of frontend stuff.
1. Probably [find_or_create_by](https://apidock.com/rails/v6.0.0/ActiveRecord/Relation/find_or_create_by) is your friend here
1. Is `unit` the right term? I picked it because it's a political designation, `area` sounds more geographical to me but open to alternatives.

```ruby
# Current structure 
# Political structure is saved as metadata (columns)
Neighbourhood {
  id: 48,
  name: "Coldhurst",
  name_abbr: "Coldhurst",
  ward: "Coldhurst",
  district: "Oldham",
  county: "Greater Manchester",
  region: "North West",
  WD19CD: "E05000723",
  WD19NM: "Coldhurst",
  LAD19CD: "E08000004",
  LAD19NM: "Oldham",
  CTY19CD: "E11000001",
  CTY19NM: "Greater Manchester",
  RGN19CD: "E12000002",
  RGN19NM: "North West"
}

# New structure (pseudocode)
# Political structure is stored as data (values)
Neighbourhood {
  # Our values
  id: 48,
  name: "Coldhurst",
  name_abbr: "Coldhurst",

  # Official values
  unit: "ward",
  unit_code_key: "WD19CD",
  unit_code_value: "E05000723" # Unique key,
  unit_name: "Coldhurst",

  Parent: Neighbourhood {
    # Our values
    id: XX,
    name: "Oldham",
    name_abbr: "Oldham",

    # Official values
    unit: "district",
    unit_code_key: "LAD19CD",
    unit_code_value: "E08000004",
    unit_name: "Oldham",

    Parent: Neighbourhood {
      # Our values
      id: XX,
      name: "Greater Manchester",
      name_abbr: "Greater Manchester",

      # Official values
      unit: "county",
      unit_code_key: "CTY19CD",
      unit_code_value: "E11000001",
      unit_name: "Greater Manchester",

      Parent: Neighbourhood {
        # Our values
        id: XX,
        name: "North West",
        name_abbr: "North West",

        # Official values
				unit_name: "region",
        unit_code_key: "RGN19CD"
        unit_code_value: "E12000002",
        unit_name: "North West"
      }
    }
  }
```

alexandria-gfsc:
Resolves #791


.github/workflows/test.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with: # ruby version specified in .ruby-version
           bundler-cache: true # Automagically runs bundle install
-          cache-version: 1    # Invalidate originally used corrupted cache
```
alexandria-gfsc, line 38: The old bundler cache will not contain the "ancestry" gem, so we need to do this so the new cache is correct.

app/controllers/pages_controller.rb
```diff
@@ -9,10 +9,8 @@ def home; end
   def find_placecal
     @grouped_sites = Site.where(is_published: true)
                          .joins(:primary_neighbourhood)
-                         .merge(Neighbourhood.order(district: :asc))
-                         .group_by{ |s| s.primary_neighbourhood.district }
-
-
+                         .merge(Neighbourhood.order(ancestry: :asc))
+                         .group_by { |s| s.primary_neighbourhood.parent.name }
```
alexandria-gfsc, line 13: This block changes the ordering on neighbourhoods. We no longer have the `.district` attribute column so it's not really possible to refer to the parent district via SQL anymore. Instead I'm asking it to filter by Ancestry, which means the resulting ordering is no longer alphabetical

app/views/admin/neighbourhoods/_form.html.erb
```diff
@@ -34,44 +33,13 @@
   <h3>Ward</h3>
   <div class="row">
     <div class="col-md-6">
-      <%= f.input :WD19CD, label: 'WD19CD', readonly: true, as: :vue_string %>
+      <%= f.input :unit_code_value, label: 'WD19CD', readonly: true %>
     </div>
     <div class="col-md-6">
-      <%= f.input :WD19NM, label: 'WD19NM', readonly: true, as: :vue_string %>
```
alexandria-gfsc, line 40: "vue_string" is unable to properly access this value for whatever reason, so it got removed from the property.

db/migrate/20211118145604_alter_structure_of_neighbourhoods.rb
```diff
@@ -0,0 +1,99 @@
+class AlterStructureOfNeighbourhoods < ActiveRecord::Migration[6.0]
```
alexandria-gfsc, line 1: The main migration, the gist of the update is "shove things into the new values". We create new districts, counties, and regions based on the data available in the previous neighbourhoods, but in some cases (i.e. `region`), the information was present as the `ward.region` attribute, but not in the code data. So we just use those attributes, as they seemed to be more accurate than the code attributes anyway.

db/migrate/20211118145604_alter_structure_of_neighbourhoods.rb
```diff
@@ -0,0 +1,99 @@
...
+                    name: ward.name }
+        next
+      end
+    end
+
+    return unless errors.any?
+
+    File.open('20211118145604_alter_structure_of_neighbourhoods.errors.txt', 'w') do |f|
+      f.write errors
+    end
```
alexandria-gfsc, line 72: This block and the formatting of the migration avoids exceptions happening during migrations and going unnoticed. If the file exists and has data, the migration failed.

db/migrate/20211118145604_alter_structure_of_neighbourhoods.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
```
alexandria-gfsc, line 74: This downgrade block has (obviously) not been tested whatsoever. I think this current way created data-loss of attributes for regions, districts, and counties. It could be refactored to properly skim wards, repackage the data, and destroy the parents, but at the time I was more focused on failing tests. This is semi-functional as-is.

test/controllers/partners_controller_test.rb
```diff
@@ -30,6 +35,7 @@ class PartnersControllerTest < ActionDispatch::IntegrationTest
   test 'should get index without subdomain' do
     get url_for controller: "partners", subdomain: false
     assert_response :success
+
```
alexandria-gfsc, line 38: None of this is interesting. Just some formatting changes while I was debugging the cause of a test failure.

app/models/neighbourhood.rb
```diff
@@ -1,44 +1,67 @@
...
+    (parent&.name if unit == 'ward') || ''
+  end
+
+  def county
+    (parent&.parent&.name if unit == 'ward') || ''
+  end
+
+  def region
+    (parent&.parent&.parent&.name if unit == 'ward') || ''
+  end
```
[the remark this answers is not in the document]
kimadactyl: hmm would `self.ancestors.where(unit: :district)` be a little clearer and more robust?

also already linking to `.name` here, do we want to return string or object?

test/factories/neighbourhood.rb
```diff
@@ -1,42 +1,86 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
+  factory :neighbourhood_region, class: 'Neighbourhood' do
```
kimadactyl, line 4: Add comment like "Neighbourhoods representing various levels of zoom from most zoomed out to most zoomed in"?

app/views/admin/neighbourhoods/_form.html.erb
kimadactyl: that is weird and could be an indication something is up with javascript builds?

app/models/neighbourhood.rb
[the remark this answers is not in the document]
alexandria-gfsc: Yeah, the commit I had forgotten to push was to change that to object and extract .name as I caught it while I was annotating the code here :) 

So for the "level missing" case we simply do what postcodes.io does, and create an empty-named ward. I think postcodes.io returns special unit codes for that (E99999...) so it shouldn't be too difficult to filter out those in the interface.

app/views/admin/neighbourhoods/_form.html.erb
alexandria-gfsc: hmm. I know that I didn't convert the Javascript picker, but specifically because that's defined in another task. Might be worth making a small ticket for it (?)

app/models/neighbourhood.rb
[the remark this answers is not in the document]
alexandria-gfsc: > hmm would `self.ancestors.where(unit: :district)` be a little clearer and more robust?

Ah I misread sorry. Yeah, that seems better.

test/models/neighbourhoods_test.rb
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env ruby
```
kimadactyl, line 1: not sure what this line is doing :) we don't have it elsewhere

app/controllers/admin/sites_controller.rb
```diff
@@ -58,7 +58,7 @@ def set_site
     end
 
     def set_variables_for_sites_neighbourhoods_selection
-      @all_neighbourhoods = policy_scope(Neighbourhood).order(:name)
+      @all_neighbourhoods = policy_scope(Neighbourhood).order(:name).where(unit: 'ward')
```
katjam, line 61: @alexandria-gfsc This is the one we talked about. Maybe all neighbourhoods should not filter for wards, but instead have another function that returns only wards - add params to this one (maybe even default to [ward]) that takes a list of units to scope the query?

Basically, it doesn't seem intuitive to call it all neighbourhoods but filter some out.

PR #804, opened 2021-12-01
Update to rails 6.1.4.1

alexandria-gfsc:
Fixes an issue where a failing test causes no error output and crashes
DRb fatally.

- User official crypt_keeper repository now that it supports Rails 6
- Remove redundant '.html.erb' from template renders
  (Removes "DEPRACATED" warning during tests)
- Make case sensitivity explicitly an argument of uniqueness
  (Removes warning during tests)
- Fix formatting of joins and fixup some double quotes to single quotes
  (Removes linter warnings)
- Change partner_test.rb:
  - test 'fields validate': Use new error object rather than map
  - test 'fields validate': Rename and split huge test into separate tests
  - Change double quotes to single quotes to remove linter warnings


app/models/partner.rb
```diff
@@ -33,8 +33,7 @@ class Partner < ApplicationRecord
   # Validations
   validates :name,
             presence: true,
-            uniqueness: true,
-            case_sensitive: false,
```
alexandria-gfsc, line 37: The rest of this file is just altering double quotes to single and changing some formatting to remove linter warnings.  This line is the one that fixes an error during tests. Essentially, `case_sensitive` has always been a specific argument of uniqueness, not a validator by itself. Rails now requires that to be explicit.

test/integration/sites_integration_test.rb
```diff
@@ -19,7 +19,7 @@ class SitesIntegrationTest < ActionDispatch::IntegrationTest
     assert_redirected_to 'http://lvh.me/'
     # If there's a match, display a custom homepage
     get 'http://hulme.lvh.me'
-    assert_template 'sites/default.html.erb'
```
alexandria-gfsc, line 22: The other render changes fix deprecation warnings about referring to templates with dots in them (Rails seems to infer the fact that it's an ERB template file). This fixes the test so that it matches the newly changed code.

test/models/partner_test.rb
```diff
@@ -14,44 +14,60 @@ class PartnerTest < ActiveSupport::TestCase
...
     partner = Partner.new
     partner.save
 
     # The only mandatory field is title (for now)
-    assert_equal ["can't be blank", "must be at least 5 characters long"], partner.errors[:name], "must have a name"
-    assert_equal 1, partner.errors.keys.length
+    assert error_map = partner.errors.where(:name, :too_short).first
+    assert_equal 'must be at least 5 characters long', error_map.options[:message]
+    assert_equal 1, partner.errors.attribute_names.length
+  end
```
alexandria-gfsc, line 25: Error maps are now error objects and should be treated like that, this fixes the error created by that, apparently the rest of the code doesn't trigger a warning or an error so I figure it's fine.

alexandria-gfsc: Also, I split these up into separate tests rather than one big monstrous test haha

kimadactyl (approved):
Looks good to me!

alexandria-gfsc:
Cool I'll squash and merge :)

PR #808, opened 2021-12-03
Fix database model and add other geography data.

### LINKED ISSUE #794: Add all other UK geography data
1. [Import all other data](geoportal1-ons.opendata.arcgis.com/datasets/cdcc46d656e84e3d997e4ab2cd77881a_0.csv) - rake task?
2. Make sure indexes etc are set up
3. Check picker widget performance

alexandria-gfsc:
Fixes #794 

Scotland was fallaciously considered a Region (sorry!) due to
fallacious data. This fixes that and extends the database so that
it covers all of the available data.

- Adds migration to correct and extend the data
- Replaces 'create from postcodesio response' function with 'find from
  postcodesio response' function
- Removes redundant test referencing obsolete function
- Adds 'country' method to Neighbourhoods.


alexandria-gfsc:
lol I ran rake earlier and it succeeded. amazing

db/migrate/20211203153422_correct_datamodel_by_importing_neighbourhoods_json.rb
```diff
@@ -0,0 +1,73 @@
...
+  def create_unit(json_unit, parent_neighbourhood, parent_unit)
+    structure = { name: json_unit['properties']['name'],
+                  unit: json_unit['properties']['unit'],
+                  unit_code_key: json_unit['properties']['unit_code_key'],
+                  unit_code_value: json_unit['properties']['code'],
+                  unit_name: json_unit['properties']['name'] }
+
+    # We do not want to recreate wards that already exist, but we don't want to waste our
+    # cycles finding unit levels that don't exist
+    # Ergo, if the parent unit is a district that means that we are at ward level,
```
katjam, line 12: And previously we only had word level data, so all other levels are just create?

test/models/neighbourhoods_test.rb
```diff
@@ -9,50 +9,9 @@ class NeighbourhoodsAncestryTest < ActiveSupport::TestCase
     assert @neighbourhood.region
     assert @neighbourhood.county
     assert @neighbourhood.district
+    assert @neighbourhood.country
 
     assert @ashton_neighbourhood.district.name == 'Tameside'
     assert @ashton_neighbourhood.district.unit == 'district'
   end
-
-  test 'create from postcodesio' do
```
katjam, line 17: Are there any tests we want to add besides checking for `.country`

katjam (approved):
This looks good and well documented. My only question is - any further test coverage we can think of adding? Like maybe the expected failure if we try to load a ward not in the system?

db/migrate/20211203153422_correct_datamodel_by_importing_neighbourhoods_json.rb
alexandria-gfsc: Yeah I explicitly remove all of the non-ward level units because the sourcing on it was "bad" -- i.e. we had Scotland listed as a region because the region name for the one ward in the original-original dataset was 'scotland'

test/controllers/admin/partners_controller_test.rb
```diff
@@ -49,7 +49,7 @@ class Admin::PartnersControllerTest < ActionDispatch::IntegrationTest
   it_allows_access_to_index_for(%i[neighbourhood_admin]) do
     get admin_partners_url
     assert_response :success
-    #assert_select 'a', 'Edit'
+    # assert_select 'a', 'Edit'
```
katjam, line 52: Are the changes in this file just linting?

alexandria-gfsc: Yeah pretty much lol

test/fixtures/neighbourhoods.yml
```diff
@@ -0,0 +1,17 @@
+# This is sufficient for like 90% of the application. We don't actually ever use region/whatever at the moment
```
katjam, line 1: Nice

alexandria-gfsc: I think any future tests {are going to/absolutely should} use fixtures so this is just a backwards compatibility thing! :)

kimadactyl:
This looks great to me. The only thing I would add is that we already have a little confusion over `name` and `name_abbr` if the latter is emptystring or a copy of `name`. I think this is another ticket though and something for `mountain_view` to worry about.

kimadactyl:
Oh and maybe bump the depends before merge too :)

alexandria-gfsc:
> This looks great to me. The only thing I would add is that we already have a little confusion over `name` and `name_abbr` if the latter is emptystring or a copy of `name`. I think this is another ticket though and something for `mountain_view` to worry about.

Yes, I think this leads to a general question of "Who is using the shortname Neighbourhoods method, and who is still using name, and who needs to use what". It's a mix at the moment in the codebase, unfortunately.

PR #812, opened 2021-12-07
Add release drafter config, resolves #810

### LINKED ISSUE #810: Set up proper releases and spring clean github config
**Is your feature request related to a problem? Please describe.**
We should probably start doing proper releases with changes and the like.

**Describe the solution you'd like**
Alice suggests mergify + dependabot and offers some sample configs [here](https://github.com/wheresalice/go_boilerplate/tree/main/.github) (altho not for rails)

**Describe alternatives you've considered**
Simply having no releases -- seems bad, it would be nice to get some proper sense of progress and to get us set for doing proper prod builds next year.

**Additional context**
Generally strip out the bits we want from [Alice's config](https://github.com/wheresalice/go_boilerplate/tree/main/.github). This includes automerging dependabot alerts using mergify so could be a big timesaver.

kimadactyl:


kimadactyl:
Ok according to [this issue](https://github.com/release-drafter/release-drafter/issues/398) it's failing because it needs to be on the main branch to work. OK if I merge this?

alexandria-gfsc:
What does the issue label assigner do? I feel like it should be obvious but my brain is glazing over lmao

.github/workflows/issue-label-assign.yml
```diff
@@ -0,0 +1,27 @@
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Naturalclar/issue-action@v2.0.2
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          title-or-body: 'title'
+          parameters: >
+           [
+           {"keywords":["chore", "refactor", "dependencies"], "labels": ["chore", "patch"]},
```
alexandria-gfsc, line 20: Do we use any of these labels these days? We mostly just use the ee/vv-related labels and then have 'bug'. But I'm not sure if we have the "chore" or "patch" labels, for example, unless im misunderstanding?

I'm also wondering how this fits into @katjam's github cleanup :)

kimadactyl: So these do the semantic versioning stuff. Patch is 0.0.x, minor is 0.x.0, major is x.0.0 :) we could ditch chore tho?

.github/workflows/issue-label-assign.yml
```diff
@@ -0,0 +1,27 @@
+name: "Set Issue Label and Assignee"
```
wheresalice, line 1: If you're planning on introducing mergify then that might be a better place to do the labeling and assigning of issues in order to reduce the number of GitHub actions that you need to run.  See https://github.com/geeksforsocialchange/meeting-mood/blob/main/.github/mergify.yml#L82

kimadactyl: Dependabot became a total nightmare for us on this repo and we just finished tidying it up so I don't think we're quite ready to re-add it yet which seems the main benefit of Mergify. Was planning on looking at it again when we relaunch publically in Feb/March time. Unless I'm missing some of the nice things Mergify can do?

wheresalice: Mergify can also apply labels based on content of PRs, and in fact that's how my setup handles major version number changes raised by dependabot. You don't need dependabot to run mergify, and you don't need to actually merge anything with mergify.

The way you have it here will work just fine, but you'll want to refactor this when you do start using Dependabot and Mergify together so that your automatic labels are being applied in one place.

kimadactyl: Ok got it. Maybe I'll take another look. Didn't have much luck turning up good Rails starter guides and kinda gave up! All the ones I could find were way too in depth or way too minimal. Thanks.

wheresalice: You might not need it to be rails-specific to get as far as creating releases. You've already got tests, you can copy most of the remaining config from imok and meeting-mood.

You could then continue deploying to staging from main, you just need to figure out how to deploy a release to prod. And that doesn't sound too difficult if it's just a git tag.

But, again, what you have here should work if you want to leave the mergify and dependabot configuration until later.

.github/workflows/issue-label-assign.yml
alexandria-gfsc: Yeah I don't think the "chore" label is necessarily required :)

PR #865, opened 2021-12-20
Ruby gemset config file

ivankocienski:
Just the configuration for rbenv-gemset (https://github.com/jf/rbenv-gemset)

### DIFF
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,0 +1,1 @@
+gfsc-place-cal


katjam (commented):
I might be missing something but without also publishing a gemset config `Id`ed by `gfsc-place-cal` how does this help? Does the current list of dependencies in gem lock file get installed globally per version of ruby? And this makes them be specifically loaded and stored with this instance env - in a folder called `gfsc-place-cal` ?

It feels like at least we'd need to commit something to the readme explaining how to harness the power and what it does. Otherwise it's maybe just a bit of code that helps in one local env in which case, maybe not track this config?

kimadactyl:
Yeah I agree with Katja, I think rvm or rbenv is up to the individual developer unless we Dockerize the lot. Happy to reopen if there's a convincing argument otherwise. Appreciate the suggestion though!

PR #875, opened 2022-01-05
Repair broken data migration / find-placecal endpoint broken

### LINKED ISSUE #874: /find-placecal endpoint currently broken [blocks production]
The find-placecal endpoint has recently been broken at some point, app/controllers/pages_controller.rb#13 returns NilClass

Acceptance Criteria:
- [ ] Endpoint no longer broken
- [ ] Integration test covering future breaks

alexandria-gfsc:
Repair broken data migration / find-placecal endpoint broken (Fixes #874).
Update yarn lockfile.

### DIFF
--- a/db/migrate/20211203153422_correct_datamodel_by_importing_neighbourhoods_json.rb
+++ b/db/migrate/20211203153422_correct_datamodel_by_importing_neighbourhoods_json.rb
@@ -10,8 +10,9 @@ class CorrectDatamodelByImportingNeighbourhoodsJson < ActiveRecord::Migration[6.
     # We do not want to recreate wards that already exist, but we don't want to waste our
     # cycles finding unit levels that don't exist
     # Ergo, if the parent unit is a district that means that we are at ward level,
+
     u = if parent_unit == 'district'
-          Neighbourhood.create_or_find_by!(structure)
+          Neighbourhood.find_or_create_by!(structure)
         else
           u = Neighbourhood.create!(structure)
         end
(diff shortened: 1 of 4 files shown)


alexandria-gfsc:
I think a consequence of this is that data migrations will have to be re-run on staging, perhaps? I don't have a backup of staging before this migration took place and I'm not 100% sure of a way to ensure that it does not break as-such.

There are going to be duplicates of the original ids that we need to remove, so I think fixing staging is going to require one of us (me lol) to:
- Crack open staging
- Back up the staging database (In case of data cleansing error)
- Retrieve duplicate `unit_code_value` entries and remove the duplicates with a larger id number
- Re-run the database migration
- Verify that it worked by attempting to retrieve duplicate code values and double checking that the find-placecal endpoint is functional

Should take approximately 20 - 40 minutes after

alexandria-gfsc:
This run is listed as failing due to the partner policy test, I'll disable it in another PR given that it is still failing. However, Feburary it's probably worth drilling into why that policy test does not respect concurrency / idempotency

kimadactyl:
That's one way of doing it. The Rails way would be instead to just handle this with another migration (delete records where ancestry is not nil, then rerun the migration the right way). I know its a bit of a time to run but would kinda rather this way I think? If you remove the logging I think it'll run a lot faster.

kimadactyl:
It looks like the test is failing because the record ID numbers don't match - I this this is probably something real!

alexandria-gfsc:
> It looks like the test is failing because the record ID numbers don't match - I this this is probably something real!

It is, but it's alarmingly difficult to replicate, I think it's a concurrency issue between the partner policy tests given how it often succeeds and then sometimes... doesn't. This is the same issue that led to us disabling the partner policy test in the first place :)

PR #889, opened 2022-01-26
Partner ServiceAreas

ivankocienski:
Migration, model, a few crude tests and some view code for service areas.

Still to do: accepting attributes for ServiceAreas in Partner create/update, verifications that Partners need either (or both) a service area or an address, a UI element so commissioners can select preferred option.


test/factories/service_areas.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :service_area do
+    partner_id { 1 }
+    neighbourhood_id { 1 }
```
kimadactyl, line 4: think this only needs to be this to use the factory niceness (associations are implied)

```
partner
neighbourhood
```

ivankocienski: ah i have yet to get to sorting out the factories. this is not a full PR

app/views/admin/partners/_form.html.erb
```diff
@@ -22,6 +22,22 @@
   <br>
   <hr>
 
+  <h2>Service Area</h2>
+
+  <div class="partner__neighbourhoods">
+    <% @all_neighbourhoods.each do |neighbourhood| %>
+      <%= fields_for "partner[service_areas_attributes][#{neighbourhood.id}]" do |sna|%>
```
alexandria-gfsc, line 29: So I think? This is a place where simple_form's association dialogue would be appropriate

.ruby-gemset
```diff
@@ -0,0 +1 @@
+gfsc-place-cal
```
alexandria-gfsc, line 1: Is this likely to be applicable to other people's configurations? :)

test/models/service_area_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+describe ServiceArea do
+  # it "does a thing" do
+  #   value(1+1).must_equal 2
+  # end
+end
```
alexandria-gfsc, line 7: gotta remove this file :)

.ruby-gemset
kimadactyl: yeah let's not specify environment, we can add to `.gitignore` if needs be

app/views/admin/partners/_form.html.erb
kimadactyl: `|sna| %>` missing space here is killing me lmao

.ruby-gemset
ivankocienski: rbenv has a plugin allowing you a per-project set of gems (https://github.com/jf/rbenv-gemset). I find it keeps my dev environment lightweight and allows me to have multiple projects with different versions of gems at the same time. You can ignore and it won't interfere with anything, but it is super useful

kimadactyl: oh i hear you and this sounds good. just as project maintainers we don't really want to be in the business of telling people what their local development environment should be. dotfiles have a way of piling up you know? i think if you are really keen on putting this in the codebase then we should be adding stuff to the README etc about our suggested environment. maybe its fine for now tho as long as we have a note to do this later. altho pls remove space :) i.e. `gfsc-placecal`

ivankocienski: I see your point. dotfiles Is there a formal process for suggesting technologies? Like if we need to set up a background job system and needed to add redis to our dev/prod machines.

I checked the README and it doesn't mention anything about requirements at all. I could put something in right now to cover rbenv, rbenv-ruby ?

kimadactyl: yeah this is a great example of something that sounds great _one day_ but we truly do not need any time soon :)

you're welcome to file an issue and label it "code smell" or "tech debt" tho!

alexandria-gfsc:
LGTM :)

alexandria-gfsc:
idk if you would prefer to squash/merge or just merge, up to you :+1:

PR #894, opened 2022-02-01
Partners can add and remove service areas

ivankocienski:
Uses old UI with all the checkboxes for neighbourhoods but it works


README.md
```diff
@@ -6,6 +6,22 @@ PlaceCal is a large and very complicated app for collating organisation and even
...
 To get an idea of the project and what we're about, check out [the handbook](https://handbook.placecal.org/) which is also [in it's own respository](https://github.com/geeksforsocialchange/PlaceCal-Handbook), and especially the [Glossary](glossary.md).
 
+## Requirements
+
+To run Placecal locally you will need:
+
+- a mac or a linux machine
+- gcc
+- postgres 
+  - if you are running in a docker image you will still need the dev libraries for the gem to compile against
```
alexandria-gfsc, line 16: I think you need the postgres developer libraries regardless of whether you run in Docker or not? (It will use the same interface if you are running over TCP or via UNIX Domain Sockets) -- I would suggest also just making this a little bit more explicit that you mean:
- The postgres developer libraries (specifically libpq-dev (debian) / postgresql-libs (arch) / whatever)
- Postgres is the thing that is being dockerized in this example, not placecal :)

app/views/layouts/admin/application.html.erb
```diff
@@ -24,6 +24,7 @@
           </div>
         </nav>
         <main role="main" class="col-md-9 mt-3 ml-sm-auto col-lg-10 px-4">
+          <%= render_component "admin_flash" %>
```
alexandria-gfsc, line 27: Ah ok, this makes more sense than adding it individually to every page :+1:

app/components/admin_edit/_admin_edit.html.erb
```diff
@@ -1,5 +1,3 @@
-<%= render_component 'admin_flash' %>
```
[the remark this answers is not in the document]
kimadactyl: yeah i think we want to use this component still tbh, it just makes it a lot easier to edit them all as a one-of later

test/models/service_area_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+describe ServiceArea do
```
kimadactyl, line 3: no sneaking in RSpec ;) we are using minitest for everything else, let's keep it one flavour

alexandria-gfsc:
This looks good -- I think we should add cocoon to it rather than it being a separate PR, it should be a small enough change AFAIK given that you are not doing any of the weird stuff that the sites neighbourhoods required.

I think this also requires a db:migrate to ratify it with my PR on main, you miiiiiiight need to force push after that like I had to :sweat_smile:

test/models/service_area_test.rb
ivankocienski: good spot! i blame my muscle memory

ivankocienski: okay weird i just uncommented the demo test example in this file and it is valid minitest. checked i had removed rspec from my gemset and it still runs. i think it would be an idea to stick to one style of tests so everything is consistent. this policy should be documented (in the handbook) somehow?

ivankocienski: This has been removed

PR #900, opened 2022-02-02
Use cocoon selector on partners form

ivankocienski:
No more thousands of checkboxes when admins want to add or edit a
partners' service areas. Implements #898

### DIFF
--- a/app/helpers/partners_helper.rb
+++ b/app/helpers/partners_helper.rb
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 
 module PartnersHelper
-  def options_for_partners
-    policy_scope(Partner).all.order(:name).pluck(:name, :id)
-  end
 
-  def options_for_neighbourhoods
-    Neighbourhood.all.order(:name).pluck(:name, :id)
+  def options_for_service_area_neighbourhoods
+    # Remove the primary neighbourhood from the list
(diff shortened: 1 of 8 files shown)


alexandria-gfsc:
LGTM :+1:

app/helpers/partners_helper.rb
```diff
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 
 module PartnersHelper
-  def options_for_partners
```
alexandria-gfsc, line 4: Oh, can you just check that this isn't breaking the users form on the offchance? (Reason: only definition of options_for_partners is here, only reference is in `views/admin/users/_form.html.erb`)

kimadactyl:
Can't currently delete partners with ServiceAreas - fix this and add a test :)

ivankocienski:
> Can't currently delete partners with ServiceAreas - fix this and add a test :)

this seems to work on my machine. 

Steps taken
- on admin page
- click on partner XYZ
- scroll to bottom of page
- push 'Destroy Partner' button

Outcome:
XYZ is removed from the list of partners

Is that the same steps you are taking? Could it be that you are not an owner of the partner?

kimadactyl:
Yeah could be my local setup is a bit broke. Reminder to add tests for this either way.

Issue #897, opened 2022-02-01
Ensure admin frontend has appropriate flash and error messages


kimadactyl:
- [ ] Make sure all templates are showing flash and error templates
- [ ] Make sure all controller actions are passing flash messages on error where suitable

First one is simple, second one is fiddlier.

PR #910, opened 2022-02-08
Improve admin UI flash messages

ivankocienski:
This implements #897

### DIFF
--- a/app/controllers/admin/calendars_controller.rb
+++ b/app/controllers/admin/calendars_controller.rb
@@ -56,7 +56,7 @@ module Admin
       @calendar.destroy
       respond_to do |format|
         format.html do
-          flash[:success] = 'Calendar was successfully destroyed.' 
+          flash[:success] = 'Calendar was successfully deleted.' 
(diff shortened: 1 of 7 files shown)


app/controllers/admin/calendars_controller.rb
```diff
@@ -53,23 +55,27 @@ def destroy
       authorize @calendar
       @calendar.destroy
       respond_to do |format|
-        format.html { redirect_to admin_calendars_url, notice: 'Calendar was successfully destroyed.' }
+        format.html do
+          flash[:success] = 'Calendar was successfully destroyed.'
```
kimadactyl, line 59: change this to deleted like the others and otherwise looking great!

kimadactyl: we should probably also note if events on the calendar get deleted too - i don't remember what happens currently

ivankocienski: Is this handled by the Events controller?

PR #934, opened 2022-02-14
Add events to GraphQL endpoint

ivankocienski:
You can now query the API with things like this for event index:
```
query {
    allEvents {
        id
        summary
        description
    }
}
```

And this for show event:
```
query {
    event(id:123) {
        id
        summary
        description
        address {
            streetAddress
            city
            postcode
        }
    }
}
```

### DIFF
--- a/app/graphql/types/address_type.rb
+++ b/app/graphql/types/address_type.rb
@@ -1,7 +1,7 @@
 module Types
   class AddressType < Types::BaseObject
 
-    description 'An address representing a physical building'
+    description 'An address representing a physical location'
 
     # field :id, ID, null: false
     field :street_address, String
--- a/app/graphql/types/address_type.rb
+++ b/app/graphql/types/address_type.rb
@@ -1,7 +1,7 @@
 module Types
   class AddressType < Types::BaseObject
 
-    description 'A partner'
+    description 'An address representing a physical building'
 
     # field :id, ID, null: false
     field :street_address, String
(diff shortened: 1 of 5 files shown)


app/graphql/types/address_type.rb
```diff
@@ -0,0 +1,15 @@
+module Types
+  class AddressType < Types::BaseObject
+
+    description 'A partner'
```
kimadactyl, line 4: An address?

app/graphql/types/address_type.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    # field :id, ID, null: false
+    field :street_address, String
+    field :street_address2, String
+    field :street_address3, String
```
kimadactyl, line 9: could we also give people access to `address.full_street_address` and `address.all_address_lines` from the model just for convenience?

app/graphql/types/query_type.rb
```diff
@@ -23,6 +23,24 @@ def all_partners
...
 
+    ####
+
+    field :event, EventType, "Find Event by ID" do
+      argument :id, ID
+    end
+
+    field :all_events, [EventType]
+
+    def event(id:)
```
kimadactyl, line 34: is this trailing colon correct? just looks odd :)

ivankocienski: this i believe is a fancy ruby convenience way of doing `def event(opts={})` followed by `Event.find(opts[:id])`

PR #963, opened 2022-02-16
Add "article" model and related logic

merchygoedwig:


db/migrate/20220216162803_create_articles.rb
```diff
@@ -0,0 +1,12 @@
+class CreateArticles < ActiveRecord::Migration[6.1]
+  def change
+    create_table :articles do |t|
+      t.text :title
+      t.text :description
```
kimadactyl, line 5: I think now we've called this "article" maybe "body" makes more sense for this field?

db/schema.rb
```diff
@@ -28,6 +28,15 @@
     t.index ["neighbourhood_id"], name: "index_addresses_on_neighbourhood_id"
   end
 
+  create_table "articles", force: :cascade do |t|
+    t.text "title"
+    t.text "description"
+    t.date "published"
```
kimadactyl, line 34: should prob be "published_at" for consistency. maybe lets make this a DateTime too just in case we need the extra accuracy later?

PR #968, opened 2022-02-17
Show service area information on location lozenges

ivankocienski:
On guest users partner index page when seeing partners if the partner has a service area this now appears as a lozenge like the neighbourhood does (if present).

Things not covered in this PR:
* styling
* business logic covering how partners must have at least one of service_area or address

### DIFF
--- a/app/assets/stylesheets/base/turf.sass
+++ b/app/assets/stylesheets/base/turf.sass
@@ -1,4 +1,4 @@
-.neighbourhood
+.neighbourhood, .service-area
   // padding-top: 0.5rem
   span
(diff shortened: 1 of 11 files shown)


kimadactyl (changes requested):
- [x] I am getting both ward **and** service lozenges for partners that have both (see below)
- [x] Let's just make the tag "various" not "various areas", looks kinda janky
- [x] This is probably a good time to add a service areas details to the partner show page? Maybe just "We operate in X, Y and Z" above the address for now?

Screenshot 2022-02-17 at 13 18 03

PR #974, opened 2022-02-17
Add TagUser model + tags to user's form. Clean up stuff

### LINKED ISSUE #971: Users can be made "tag admins"
- [ ] Add migration to create UserTags
- [ ] Add select2 selector to User Edit

alexandria-gfsc:
Fixes #971 

- Add TagsUser model
  - Pluralisation fixed courtesey of Kim
  - Edit User, Tag models to fix relation
- Add Tags to User's form
  - Clean up User form
    - Move roles and tags out of user's controller into helper
    - Properly policy scope tags
    - Add role_label Singleton
    - Switch to using two column layout
    - Fix options_for_neighbourhoods formatting
    - Fix strategy_label formatting
- Add TagsUsers to Tag model and form
  - Add edit_permission_label singleton to Tag
  - Add edit_permission radio buttons to Tags form

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -8,8 +8,10 @@ class Tag < ApplicationRecord
 
   self.table_name = 'tags' # Maybe we can remove this? Tag should automagically railsify to tags right?
 
-  has_and_belongs_to_many :users
   has_and_belongs_to_many :partners
+
+  has_many :tags_users, dependent: :destroy
+  has_many :users, through: :tags_users
 
   validates :name, :slug, presence: true
   validates :name, :slug, uniqueness: true
@@ -23,4 +25,15 @@ class Tag < ApplicationRecord
   enumerize :edit_permission,
             in: %i[root all],
             default: :root
+
+  def self.edit_permission_label(value)
+    case value.second
+    when 'root'
+      '<strong>Root</strong>: Only root-level users may assign this tag'.html_safe
+    when 'all'
+      '<strong>All</strong>: Any user may assign this tag'.html_safe
+    else
+      value
+    end
+  end
 end
--- a/app/models/tags_user.rb
+++ b/app/models/tags_user.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class TagsUser < ApplicationRecord
+  self.table_name = 'tags_users'
+  belongs_to :tag
+  belongs_to :user
+  validates :tag_id,
+            uniqueness: {
+              scope: :user_id,
+              message: 'User cannot be assigned more than once to a tag'
+            }
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -116,20 +116,6 @@ class User < ApplicationRecord
     facebook_app_id.present? && facebook_app_secret.present?
   end
 
-  def self.role_label(value)
-    case value.second
-    when 'root'
-      '<strong>Root</strong>: Can do everything'.html_safe
-    when 'editor'
-      '<strong>Editor</strong>: Can edit news articles'.html_safe
-    when 'citizen'
-      '<strong>Citizen</strong>: ' \
-      'Can only edit entities listed on this page'.html_safe
-    else
-      value
-    end
-  end
-
   def assigned_to_postcode?(postcode)
     return true unless neighbourhood_admin?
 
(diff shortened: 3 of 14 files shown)


alexandria-gfsc:
The only other test that breaks is:
```
Error:
Admin::TagsControllerTest#test_root:_can_destroy:
DRb::DRbRemoteError: PG::SyntaxError: ERROR:  zero-length delimited identifier at or near """"
LINE 1: DELETE FROM "tags_users" WHERE "tags_users"."" IS NULL
                                                    ^
 (ActiveRecord::StatementInvalid)
    app/controllers/admin/tags_controller.rb:69:in `destroy'
    test/controllers/admin/tags_controller_test.rb:88:in `block (2 levels) in <class:TagsControllerTest>'
    test/controllers/admin/tags_controller_test.rb:87:in `block in <class:TagsControllerTest>'
    test/test_helper.rb:44:in `instance_exec'
    test/test_helper.rb:44:in `block (4 levels) in <class:TestCase>'
```
which is seemingly a Schema issue. 

As can be seen:
```
irb(main):001:0> TagsUser.first.id
  TagsUser Load (0.3ms)  SELECT "tags_users".* FROM "tags_users" LIMIT $1  [["LIMIT", 1]]
=> nil
```
While the existing schema is near-identical to that of `SitesNeighbourhoods`, they are missing `id` properties. I'm not sure how to resolve this, either.

app/views/admin/users/_form.html.erb
```diff
@@ -3,19 +3,46 @@
 <div class="form-group">
   <%= render "contact", f: f %>
 
-  <%# f.association :tags, label: false, collection: @tags, as: :check_boxes %>
-  <%= f.association :partners,
-                    label: tooltip_label('Partners', 'Partners this user can edit, as well as their associated calendars'),
-                    collection: options_for_partners,
-                    selected: (@partner.id if @partner),
```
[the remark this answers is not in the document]
alexandria-gfsc: Created #983. Should be solved as of next push.

alexandria-gfsc:
Should be ready for merge barring any failing tests (There shouldn't be any, but who knows !)

app/models/tag.rb
```diff
@@ -23,4 +25,15 @@ class Tag < ApplicationRecord
   enumerize :edit_permission,
             in: %i[root all],
             default: :root
+
+  def self.edit_permission_label(value)
```
ivankocienski, line 29: Please keep view code out of the model and move it to a helper.

app/models/user.rb
```diff
@@ -117,6 +116,20 @@ def has_facebook_keys?
     facebook_app_id.present? && facebook_app_secret.present?
   end
 
+  def self.role_label(value)
```
ivankocienski, line 119: Please move this code to the users_helper

app/models/tag.rb
alexandria-gfsc: I disagree that this is view code per se. It's rendering an opaque property of a model (in this case an enum) to a string, for the sake of future editing we should be able to edit and alter this in one single place. In addition, there is precedence for this being within the model as per the Neighbourhoods `shortname`, `contextual_name` values (https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/models/neighbourhood.rb#L34), in addition to other examples of this singleton pattern across the codebase where Enumerize is used.

This belongs here because while it does interact with a value passed from the view, it is rendering a property of a model, that should by itself belong to the model.

app/models/tags_user.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class TagsUser < ApplicationRecord
+  self.table_name = 'tags_users'
```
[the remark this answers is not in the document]
alexandria-gfsc: Found it! Prodding u on discord

app/models/tag.rb
alexandria-gfsc: Solved in standup

app/models/user.rb
alexandria-gfsc: Done

app/models/tags_user.rb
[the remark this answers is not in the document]
alexandria-gfsc: Discussed in standup :)

PR #985, opened 2022-02-21
Validates Partner model for optional fields

ivankocienski:
Must have either/both of service area or address

### DIFF
--- a/test/models/partner_test.rb
+++ b/test/models/partner_test.rb
@@ -4,7 +4,7 @@ require 'test_helper'
 
 class PartnerTest < ActiveSupport::TestCase
   setup do
-    @new_partner = FactoryBot.build(:partner, address: FactoryBot.create(:address))
+    @new_partner = build(:partner, address: create(:address))
     @user = create(:user)
   end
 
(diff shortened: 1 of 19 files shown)


alexandria-gfsc:
Creating a new partner without an address is valid, however the site consequently breaks when trying to show that partner.

test/models/partner_test.rb
```diff
@@ -4,107 +4,99 @@
 
 class PartnerTest < ActiveSupport::TestCase
   setup do
-    @partner = create(:partner)
```
alexandria-gfsc, line 7: Tiny nitpick: Is there any reason to use `FactoryBot.{build,create}`, when you can just do `build()` or `create()`? Most tests do the latter

test/models/partner_test.rb
```diff
@@ -161,3 +153,53 @@ class PartnerServiceAreaTest < ActiveSupport::TestCase
   end
 
 end
+
+class PartnerAddressOrServiceAreaPresenceTest < ActiveSupport::TestCase
+
+  setup do
+    @user = create(:root)
```
alexandria-gfsc, line 160: Especially since down here you do `create()` :slightly_smiling_face:

alexandria-gfsc:
It looks like you're creating a single `@new_partner` for all the tests? I'm not 100% sure that `setup` is run independently for each test, so this potentially subtly breaks the isolating we get across partner tests at the moment

Issue #988, opened 2022-02-21
Users must either have admin rights for a neighbourhood containing the address, or their AdminNeighbourhood must completely contain (or be the same as) the service areas


ivankocienski:
As part of #969

PR #989, opened 2022-02-22
Partners now validate service areas for users

ivankocienski:
Implements #988 

Business logic should be complete and ready for optimisation

### DIFF
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -77,6 +77,7 @@ class Partner < ApplicationRecord
   validates_associated :address
 
   validate :check_ward_access
+  validate :check_service_area_access
 
   validate :must_have_address_or_service_area
 
@@ -193,6 +194,8 @@ class Partner < ApplicationRecord
   private
 
   def check_ward_access
+    return unless address.present?
+
     user = User.where(id: accessed_by_id).first
 
     return if user.blank?
@@ -200,6 +203,21 @@ class Partner < ApplicationRecord
     unless user.assigned_to_postcode?(address&.postcode)
       errors.add(:base, 'Partners cannot be created outside of your ward.')
     end
+  end
+
+  def check_service_area_access
+    return unless service_areas.any?
+
+    user = User.where(id: accessed_by_id).first
+    return if user.blank?
+
+    my_neighbourhoods = service_area_neighbourhoods.pluck(&:id)
+    user_neighbourhoods = user.owned_neighbourhood_ids
+
+    p_nh = Set.new(my_neighbourhoods)
+    u_nh = Set.new(user_neighbourhoods)
+
+    u_nh.superset?(p_nh)
   end
 
   def must_have_address_or_service_area
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -81,7 +81,7 @@ class Partner < ApplicationRecord
 
   validate :must_have_address_or_service_area
 
-  attr_accessor :accessed_by_id
+  attr_accessor :accessed_by_user
 
   mount_uploader :image, ImageUploader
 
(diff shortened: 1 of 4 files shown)


test/models/partner_test.rb
```diff
@@ -201,5 +202,61 @@ class PartnerAddressOrServiceAreaPresenceTest < ActiveSupport::TestCase
...
+    new_partner = build(
+      :partner, 
+      address: create(
+        :address,
+        neighbourhood: @user_neighbourhood
+      )
+    )
+
+    new_partner.accessed_by_id = @user.id
+    new_partner.save!
```
alexandria-gfsc, line 224: From the fact the one set of GH actions tests failed and the other set of GH actions tests did not, it looks like it may introduce a test concurrency race condition?

```

Error:
PartnerAddressOrServiceAreaPermissionsTest#test_verify:_address_in_user_ward:
ActiveRecord::RecordInvalid: Validation failed: Partners cannot be created outside of your ward.
    test/models/partner_test.rb:224:in `block in <class:PartnerAddressOrServiceAreaPermissionsTest>'
```

app/models/partner.rb
```diff
@@ -202,6 +205,21 @@ def check_ward_access
     end
   end
 
+  def check_service_area_access
+    return unless service_areas.any?
+
+    user = User.where(id: accessed_by_id).first
+    return if user.blank?
+
+    my_neighbourhoods = service_area_neighbourhoods.pluck(&:id)
```
alexandria-gfsc, line 214: pluck will return only the first service area neighbourhood, I think

alexandria-gfsc: At least I had that exact problem when I used this in `owned_neighbourhood_ids`. You probably want `collect` or `map`, and I don't see an access to subtree so you're not accounting for neighbourhoods that are of a lower region service area (Might not be necessary?)

test/models/partner_test.rb
```diff
@@ -201,5 +202,61 @@ class PartnerAddressOrServiceAreaPresenceTest < ActiveSupport::TestCase
...
+
+    assert new_partner.valid?
+  end
+
+  test "verify: with service area in user neighbourhoods" do
+    
+    new_partner = build(
```
alexandria-gfsc, line 231: Formatting nitpick hehe: can the build() stuff be on a single line :)

test/models/partner_test.rb
```diff
@@ -201,5 +202,61 @@ class PartnerAddressOrServiceAreaPresenceTest < ActiveSupport::TestCase
...
+end
 
+class PartnerAddressOrServiceAreaPermissionsTest < ActiveSupport::TestCase
+  setup do
+    @user = create(:user)
+    @user_neighbourhood = create(:neighbourhood, unit_code_value: "E05011368")
+    @user.neighbourhoods << @user_neighbourhood
+  end
+
+  test "verify: address in user ward" do
```
alexandria-gfsc, line 214: oh! maybe some clearer naming here too :)

like, "we can save if address is in user ward" or "valid if address in user ward" or something, just for down the road when we've forgotten what this test does :slightly_smiling_face:

app/models/partner.rb
alexandria-gfsc: - Coming back to this: I was wrong! :)  I dunno what weirdness happened with pluck last time I tried it to make it return only 1 item when map/collect returned all 600-odd, but it seems to work fine now. Chalking this up to eldrich horrors / rails trauma.

I think there might? be a difference between `&:id` and `:id` here, and you will want the latter. Just checked in console and the former gives some wacky output by comparison

test/models/partner_test.rb
```diff
@@ -6,6 +6,7 @@ class PartnerTest < ActiveSupport::TestCase
   setup do
     @new_partner = build(:partner, address: create(:address))
     @user = create(:user)
+    @new_partner.accessed_by_user = @user
   end
```
alexandria-gfsc, line 10: lol. tiniest nitpick and im already sorry for it. but its probably cleaner to swap the user creation first and include it into the build arguments directly?

test/models/partner_test.rb
```diff
@@ -103,11 +104,17 @@ class PartnerTest < ActiveSupport::TestCase
...
-    @partner = create(:partner)
-    @neighbourhood = create(:neighbourhood)
+    @neighbourhood = neighbourhoods(:one)
+    @user = create(:user)
+    @user.neighbourhoods << @neighbourhood
+    @partner = build(:partner, address: nil, accessed_by_user: @user)
   end
 
   test 'is valid when empty' do
+    # give partner an address the user admins
```
alexandria-gfsc, line 114: this comment accidentally a word and im not sure what it is supposed to be saying? hehe :sweat_smile:

alexandria-gfsc:
LGTM :)

Issue #993, opened 2022-02-24
If user.partner_admin? is true, then they should be able to edit service areas


ivankocienski:
When creating or updating a partner, only allow users who have that partner as one of their "user_partner" entries to modify the partners service areas.

(at the moment any user who has the same neighbourhoods as the partner has service areas is allowed to change a partners service areas)

PR #1000, opened 2022-02-28
User service area partner admin

ivankocienski:
Implements #993 - users can only modify partners that they have links to


alexandria-gfsc:
LGTM

PR #1025, opened 2022-03-04
Add updated_at column to index pages / datatables

### LINKED ISSUE #995: Add "last updated at" column to datatables
**Is your feature request related to a problem? Please describe.**

It's hard to see which partners have been edited recently which is making debugging tricky

**Describe the solution you'd like**

Add a sortable "last updated at" column to all datatables

alexandria-gfsc:
Fixes #995

I'm not sure if we want to postprocess the times to be more human readable than they would be otherwise (maybe UK format them at least and do some funky timezone accounting), and I'm not sure if we want to add a custom ordering thing or just disable that entirely. Hence: draft PR


kimadactyl:
ah you get to use the best helper in rails :) 

https://apidock.com/rails/ActionView/Helpers/DateHelper/distance_of_time_in_words

alexandria-gfsc:
> ah you get to use the best helper in rails :)
> 
> https://apidock.com/rails/ActionView/Helpers/DateHelper/distance_of_time_in_words
> 
> edit: altho saying that im not sure how sort would work, hmmm

Oooh! Yes, I see

Yeah, right now with this you can search for `2021` and it will just list all entries from 2021, so it's possible that by making it more user friendly we're somehow lowering the utility of it. I'm wondering if searching / sorting needs to have some different logic to displaying the thing. Maybe we can respond to requests with an ISO/UNIX time and then the front end rejigs it into words. Or maybe there is room for a function that reverses distance_of_time_in_words based on the supplied table? hehe

alexandria-gfsc:
Haven't looked too deeply into this but -- https://datatables.net/blog/2014-12-18#Completed-plug-in

kimadactyl:
yeahh tbh for now lets just do whatever is the simplest way to make it sortable at all, why does everything have so much to think aboutttt lol. i mean at the very least we can return the results in most recently updated first on the rails end to make life easier for admins

Issue #1040, opened 2022-03-09
Reformat partner response structure (also quickly change API route)


ivankocienski:
Partners should look something like
```
allPartners(with_tag: 12, inside_neighbourhood: 12314) {
	@type: Organization
	name (from model)
	summary
	description
	accessibilitySummary (accessibility_info)
	placecalID (link to placecal site)
	logo (from image they uploaded to placecal)
	address {
		@type: PostalAddress
		streetAddress (lines 1-3)
		addressLocality (ward from neighbourhood)
		addressRegion (district from neighbourhood)
		addressCountry (from country code)
		postalCode (from postcode)
		neighbourhood {
			@type: AdministrativeArea
			placecalID
			firstHalfOfPostcode (not done yet)
			shortName (our abbreviation for it)
			containedInPlace
			containsPlace
		}
	}
	url
	facebookPage
	twitterHandle
	areasServed (from service_areas) {
		[
      (these fields come from neighbourhood)
			name
			abbreviatedName
			unit
			unitName
			unitCodeKey
			unitCodeValue      
    ],
		[],
		[]
	}
	contact {
		@type: ContactPoint
		name (from partner)
		email (from partner)
		telephone (from partner.public_phone)
	}
	openingHours {
		@type: OpeningHours
	}

}
```
where `@type` denotes the schema.org type definition and things in brackets are sort of notes or implementation hints

PR #1045, opened 2022-03-10
Partner schema enhancement

ivankocienski:
Implements #1040 

Improves schema for partners and adds versioning to API endpoint

### DIFF
--- a/app/graphql/types/address_type.rb
+++ b/app/graphql/types/address_type.rb
@@ -5,13 +5,31 @@ module Types
 
     # field :id, ID, null: false
     field :street_address, String
-    field :street_address2, String
-    field :street_address3, String
-    field :city, String
-    field :postcode, String
-    field :country_code, String
-    field :full_street_address, String
-    field :all_address_lines, [String]
+    field :address_locality, String
+    field :address_region, String
+    field :address_country, String, method: :country_code
+    field :postal_code, String, method: :postcode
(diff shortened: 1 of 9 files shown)


app/graphql/types/address_type.rb
```diff
@@ -5,13 +5,31 @@ class AddressType < Types::BaseObject
...
+    # TODO: figure out parent and children accessors
+    # field :containedInPlace
+    # field :containedPlace
+
+    def street_address
+      [ 
+        object.street_address,
+        object.street_address2,
+        object.street_address3
+      ].reject { |line| (line || '').empty? }.join(', ')
```
kimadactyl, line 23: I think this is a dupe of a method in the address model?

app/graphql/types/partner_type.rb
```diff
@@ -3,18 +3,38 @@ class PartnerType < Types::BaseObject
...
-    field :description, String
     field :summary, String
+    field :description, String
+
+    field :accessibility_summary, String, method: :accessibility_info
+
+    field :logo, String, method: :image
+    field :address, AddressType
 
     field :url, String
```
kimadactyl, line 17: Just to confirm theres not better types to use than string for email, phone, url etc?

ivankocienski: graphql has very limited basic types: https://graphql.org/graphql-js/basic-types/

Issue #1049, opened 2022-03-14
Reformat event API structure


ivankocienski:
GraphQL response format should be
```
// Sorted from current date in date order
event {
    @type: schema.org/Event
    id
    name (alias for summary)
    summary (string)
    description (string)
    startDate (from dtstart, string, ISO format)
    endDate (from dtend, string, ISO format, can be null)
    
    location (from address, AddressType)
    organizer (from partner, PartnerType)
}
```

kimadactyl:
looks great two small changes. lets add this as a convenience

```
name: summary
summary: summary
```

and organi**z**er to be schema.org complient (grr us english)

PR #1050, opened 2022-03-14
New response format for Events via graph api

ivankocienski:
Implements #1049

### DIFF
--- a/app/graphql/types/event_type.rb
+++ b/app/graphql/types/event_type.rb
@@ -1,13 +1,15 @@
 module Types
   class EventType < Types::BaseObject
 
-    description 'An Event that is run by a Parter'
-
     field :id, ID, null: false
-    field :description, String
+    field :name, String, method: :summary
     field :summary, String
+    field :description, String
+    field :startDate, String, method: :dtstart, null: false
+    field :endDate, String, method: :dtend
 
     field :address, AddressType
+    field :organizer, PartnerType, method: :partner
 
   end
 end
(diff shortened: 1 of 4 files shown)


app/graphql/types/event_type.rb
```diff
@@ -1,13 +1,15 @@
 module Types
   class EventType < Types::BaseObject
 
-    description 'An Event that is run by a Parter'
-
     field :id, ID, null: false
-    field :description, String
+    field :name, String, method: :summary
```
kimadactyl, line 5: Maybe just add a comment here like

# Summary and name are aliases, this is left as a convenience for people used to iCal format

app/graphql/types/query_type.rb
```diff
@@ -33,7 +33,7 @@ def event(id:)
     end
 
     def event_connection(**args)
-      Event.all
+      Event.sort_by_time.all
```
kimadactyl, line 36: Does this filter from `Time.now`?

ivankocienski: This orders by `dtstart` descending.

kimadactyl: I think there's no point returning events that are in the past so maybe should update the filter? (assume you mean ascending too)

ivankocienski: as i understand- the events are returned in the same paginated way as partners, just ordered by dtstart so if you have an offset of 100 then all the events will be returned after 100 ordered by dtstart. there is currently no way to specify a 'from date' field in the query

kimadactyl (approved):

Issue #1066, opened 2022-03-21
Create article API endpoint for Articles


ivankocienski:
Articles schema type [here](https://schema.org/Article)

```
Article {
    id: ID
    name: string, alias for headline
    headline: string, from title
    text: string, from body
    articleBody: string, from body
    datePublished: string, published_on
    dateCreated: string, from created_at
    dateUpdated: string, from updated_at
    # creativeWorkStatus: string, from is_draft
    provider: partner
    # author: person, from user
}
```

Articles are a sub-field on `PartnerType` for now. Only allow access to non-draft articles

PR #1067, opened 2022-03-21
Articles can be accessed through partner API

ivankocienski:
Implements #1066

### DIFF
--- a/app/graphql/types/article_type.rb
+++ b/app/graphql/types/article_type.rb
@@ -2,7 +2,7 @@ module Types
   class ArticleType < Types::BaseObject
     
 
-    description 'A piece of news involving a partner'
+    description 'A piece of news article posted on PlaceCal'
 
     # field :id, ID, 
     #   null: false,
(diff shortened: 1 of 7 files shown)


app/graphql/types/article_type.rb
```diff
@@ -0,0 +1,50 @@
...
+    field :text, String, 
+      description: 'The contents of the article (alias for articleBody)',
+      method: :body
+    
+    field :article_body, String, 
+      description: 'The contents of the article (in markdown format)',
+      method: :body
+
+    field :date_published, String, 
+      description: 'Date that the article was published on',
```
kimadactyl, line 28: on PlaceCal

app/graphql/types/article_type.rb
```diff
@@ -0,0 +1,50 @@
...
+      method: :published_at
+
+    field :date_created, String, 
+      description: 'The time that the article was created on PlaceCal',
+      method:  :created_at
```
kimadactyl, line 33: make these the iso strings

PR #1095, opened 2022-03-29
Reading articles through the GQL API

ivankocienski:
Implements #1094.

New `allArticles` endpoint of type `[ArticleType]`. Just returns whatever is in the articles table.

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -46,7 +46,7 @@ module Types
     end
 
     def article_connection(**args)
-      Article.published.order(published_at: :desc).all
+      Article.global_newsfeed
     end
   end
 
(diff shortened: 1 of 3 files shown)


kimadactyl (changes requested):
Looks good let's just

- [x] Preemptively add pagination
- [x] Make them come as reverse sorted by date

app/graphql/types/query_type.rb
```diff
@@ -33,6 +33,18 @@ def partners_by_tag(tag_id:)
     end
   end
 
+  module ArticleQueries
+    def self.included(klass)
+      klass.field :all_articles, [ArticleType] do
+        description 'Return news articles from all sites for all partners'
```
kimadactyl, line 39: "Return all news articles across the whole site"?

ivankocienski: no longer applicable

kimadactyl (approved):

PR #1096, opened 2022-03-29
Fix papertrail causing timeouts and add rake import:nuke_papertrail

### LINKED ISSUE #1087: Optimise or remove PaperTrail
https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/models/calendar.rb#L55

This code is only used in one place and seems to show recently modified events (updated, deleted). PaperTrail is used specifically to catch deleted events without doing any specific (handrolled) work to archive those events.

However, it's causing a timeout on a few calendar pages in production and could be quite troublesome based on the SQL query it executes now we have more throughput from calendars. On most calendar pages it returns absolutely nothing, however on certain calendar pages it executes a huge SQL query that causes the page to timeout (Viewing the log I've captured from one of these entries in certain editors causes huge freezing and slowdown of the editor...).

I'm not sure how useful this is from a user perspective though -- if it is useful we should find a better way to store recently (within about 2 weeks) deleted and updated events for this purpose (maybe just have a flag on them, or better yet stick them in a separate table), so that the lookup can be cleaner and organized per-calendar.

TL;DR: PaperTrail is annoying and:

- Only used in one single function
- Returns a value that is used exclusively in view code
- For something that we could do without PaperTrail, seemingly, and a hell of a lot faster on the view code rendering I think (Would make event deletion slower however this _might_ be negligible).
- It's causing timeout on edit pages to the point that the site gives up

alexandria-gfsc:
FWIW before I did import:nuke_papertrail, there were around a million papertrail versions... lol

Fixes #1087


lib/tasks/events.rake
```diff
@@ -20,6 +20,10 @@ namespace :import do
 
     import_events_from_source(args[:calendar_id], from)
   end
+
+  task nuke_papertrail: :environment do
```
kimadactyl, line 24: maybe lets make this a bit more neutral lol. "purge_papertrail"?

alexandria-gfsc: I did contemplate purge, yeah lol

kimadactyl (approved):

PR #1097, opened 2022-03-29
Clean up postcodes with excess whitespace

### LINKED ISSUE #1079: address and areasServiced data not consistant
**Describe the bug**
The frontend relies on data for address & service being standardised - if we sometimes get empty strings or prepended spaces, we accidentally will display them as string values. 
image

**To Reproduce**
- Example of badly formatted `postalCode`: Yellowbird Age Friendly (id 17) & New Hope Fellowship Church (id 49)
- maybe related Query Partner id 17 and get "message": "no implicit conversion of nil into String",
- Example of empty string for `abbreviatedName` Partner (id 145)

**Expected behavior**
- We expect `address` to contain a `postalCode` field with a UK post code formatted as "AB1 2GT" with a space between the area & district, but sometimes there is a prepended space.

- We expect `areasServed` to contain a list of service areas that contain `name` and maybe an `abbreviatedName` sometimes the abbreviated name is recorded as an empty string rather than `null`

**Screenshots**
image
image

ivankocienski:
Fixes #1079

There is a migration that must run on data in db.

### DIFF
--- a/app/graphql/types/neighbourhood_type.rb
+++ b/app/graphql/types/neighbourhood_type.rb
@@ -9,7 +9,6 @@ module Types
       description: 'The common name for this region'
 
     field :abbreviated_name, String,
-      method: :name_abbr,
       description: 'Shorthand version of name in common use'
 
     field :unit, String,
--- a/app/graphql/types/neighbourhood_type.rb
+++ b/app/graphql/types/neighbourhood_type.rb
@@ -9,7 +9,7 @@ module Types
       description: 'The common name for this region'
 
     field :abbreviated_name, String,
-      description: 'Shorthand version of name in common use'
+      description: 'Abbreviated version of name or just the name if this is not set'
 
     field :unit, String,
       description: 'Size of neighbourhood: country -> region -> county -> district -> ward'
--- a/app/models/neighbourhood.rb
+++ b/app/models/neighbourhood.rb
@@ -49,10 +49,26 @@ class Neighbourhood < ApplicationRecord
     end
   end
 
+  # give us a name we can use for the abbrevaited name even if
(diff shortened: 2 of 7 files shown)


test/models/address_test.rb
```diff
@@ -3,7 +3,17 @@
...
 class AddressTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+  test 'normalizes postcode' do
+    Geocoder.stub :search, [] do
+
+      address = Address.create(
+        street_address: '123 street',
+        postcode: '   ha8 5ha ',
```
kimadactyl, line 11: maybe add a few more test cases here including postcodes like "W1W 5AA" and "NG12 4FZ" with various spacing and also spacing between the two parts of it

app/graphql/types/neighbourhood_type.rb
```diff
@@ -9,7 +9,6 @@ class NeighbourhoodType < Types::BaseObject
       description: 'The common name for this region'
 
     field :abbreviated_name, String,
-      method: :name_abbr,
       description: 'Shorthand version of name in common use'
```
kimadactyl, line 12: "Abbreviated verison of name, or just the name if this is not set"

app/models/neighbourhood.rb
```diff
@@ -49,6 +49,20 @@ def fullname
     end
   end
 
+  def abbreviated_name
+    abbr_name = self.name_abbr.to_s.strip
+
+    return abbr_name if abbr_name.present?
```
kimadactyl, line 55: abbr_name.present? ? abbr_name : name

PR #1098, opened 2022-03-30
Remove event duplicates where dtstart, summary, and calendar_id are the same

### LINKED ISSUE #1072: Investigate duplicate events
**Describe the bug**

There are a large number of duplicate events for some MMU stuff.

Not sure if this is because of

* This specific importer
* The importer in general
* Something else

alexandria-gfsc:
Kinda Fixes #1072 

Also has our first yarddoc comment :partying_face:

### DIFF
--- a/test/models/events_test.rb
+++ b/test/models/events_test.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class EventTest < ActiveSupport::TestCase
+  setup do
+    @calendar = create(:calendar)
+  end
+
+  test 'ensure validation fails for duplicate event model' do
+    assert_raises ActiveRecord::RecordInvalid do
+      create_list(:event, 2,
+                  summary: 'Mom and Pops Pet Store',
+                  dtstart: DateTime.now - 1.hour,
+                  calendar: @calendar)
+    end
+  end
+
+  test 'can create two events with the same start date and calendar but diff summary' do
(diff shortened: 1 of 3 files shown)


kimadactyl:
Haha sadly Ivan has beaten you to this.

This looks good but maybe lets take this opportunity to refactor these really long methods with some helper private ones? They are hard to read

test/models/events_test.rb
```diff
@@ -0,0 +1,42 @@
...
+                  dtstart: DateTime.now - 1.hour,
+                  calendar: @calendar)
+    end
+  end
+
+  test 'can create two events with the same start date and calendar but diff summary' do
+    event_hash = { dtstart: DateTime.now - 1.hour, calendar: @calendar }
+    a = create(:event, summary: 'cycling triathalon', **event_hash)
+    b = create(:event, summary: 'foonlys get together :,)', **event_hash)
+    assert a
```
ivankocienski, line 23: ohh uhhh hmmm. i am not sure i like that you are using factories to test model validation code. it works but it seems odd- like why can't you just `Event.new` put in a bunch of data and then check `assert event_a.valid?`. in this way intent is explicitly defined in code. `assert a` is a very C flavour

alexandria-gfsc: I mean I can, but that would require doing a bunch of stuff to either skip or satisfy location validation. The whole point of factories is allowing us to create models that satisfy X constraints and allow me to focus on testing, solely, 3 variations of ±the same object, without having to put all the work into constructing them.

ivankocienski (approved):

PR #1109, opened 2022-04-04
Safely read opening_times from partner

### LINKED ISSUE #1099: Querying a partner for OpeningHours throws error if data has a nil value
**Describe the bug**
Cannt get opening hours data for a Partner with `nil` in them.

**To Reproduce**
1. Using graphiql:
```
query Partner($id: ID!) { 
  partner(id: $id) {
    openingHours {
      dayOfWeek
      opens
      closes
    }
  }
}
```

```
{"id": 17
}
```

**See error**
`"message": "no implicit conversion of nil into String",`

**Expected behavior**
Response json for a partner's opening hours

ivankocienski:
Fixes #1099

### DIFF
--- a/test/integration/graphql/partners_integration_test.rb
+++ b/test/integration/graphql/partners_integration_test.rb
@@ -253,4 +253,29 @@ class GraphQLPartnerTest < ActionDispatch::IntegrationTest
     partner_data = data['partnersByTag']
     assert partner_data.length == 1, 'expecting to see a tag on this partner'
   end
+
+  test 'returns null properly if openning times are missing' do
+
+    partner = FactoryBot.create(:partner, opening_times: nil)
+
+    query_string = <<-GRAPHQL
+      query {
+        partner(id: #{partner.id}) {
+          openingHours {
+            opens
+            closes
+            dayOfWeek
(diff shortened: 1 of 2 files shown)


test/integration/graphql/partners_integration_test.rb
```diff
@@ -253,4 +253,29 @@ def check_basic_fields(data, partner)
...
+            opens
+            closes
+            dayOfWeek
+          }
+        }
+      }
+    GRAPHQL
+
+    result = PlaceCalSchema.execute(query_string)
+    assert result.key?('errors') == false, 'errors are present'
```
alexandria-gfsc, line 274: I think you can use `refute` here rather than `assert x == false`?

ivankocienski: do you think i should change it?

alexandria-gfsc: It's up to you, and how much energy you have IMO. Like I think changing it would mean changing the whole file, but on the other hand `assert x == false` is generally an antipattern I think

kimadactyl (approved):

PR #1111, opened 2022-04-05
Tidier tests for graphql partner endpoint

### LINKED ISSUE #1107: Fix some minor incongruence in the GraphQL tests
We have a function called `check_address` and `check_service_areas` that take two inputs, presumably to compare the `data` against the `address` or `service_area`, but it discards the `address`/`service_area` argument and simply checks that `data` has the properties we asked graphql for. `check_opening_hours` and `check_areas_served` has a similar thing where it just checks that if is an array or not but not the return value.

See: https://github.com/geeksforsocialchange/PlaceCal/blob/main/test/integration/graphql/partners_integration_test.rb

We have a few open FIXMEs dotted around the place such as on line 115 (although this is codebase wide)

ivankocienski:
Fixes #1107

### DIFF
--- a/test/integration/graphql/partners_integration_test.rb
+++ b/test/integration/graphql/partners_integration_test.rb
@@ -59,25 +59,33 @@ class GraphQLPartnerTest < ActionDispatch::IntegrationTest
   def verify_field_presence(obj, name, value: nil)
     assert obj.key?(name), "field #{name} is missing"
     if value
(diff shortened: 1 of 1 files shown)


alexandria-gfsc:
:partying_face: LGTM!

alexandria-gfsc (approved):
yooo

PR #1112, opened 2022-04-05
Articles now have Image Headers

### LINKED ISSUE #1108: Add an image field to news articles (16:9, 636 x 357 on screen)

alexandria-gfsc:
Fixes #1108 

- Add `ArticleHeaderUploader`, make sure it properly crops and centers
- Add `article_image` field to Article, policy code, controller, etc.
- Add `highres_image` to Article for GraphQL
- Guard `author_name` to return empty string for GraphQL in Articles
- Add `article_image` upload button and preview to view code 
- Add `image` field and correct `author` field type in GraphQL
  - Also align the arguments of a method call when they span more than one line
- Change `author_name` to return an empty string as the field is required
- Add deep testing of Article models returned from GraphQL
- Rename `verify_field_presence` -> `assert_field_equals` (checked with Ivan :slightly_smiling_face:)
- Pull out `assert_field_equals` into `test_helper.rb` to be used in other tests
- Add `assert_field` to differentiate between field existence and property tests
- Some minor changes in GraphQL partner test to use `%r{}` so the regex looks
  easier on the eyes, and to use `Regexp.last_match(1)` rather than the
  non-descriptive `$1`. Also remove now-needless escaping of regex \\:D/

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -10,7 +10,7 @@ class Article < ApplicationRecord
 
   belongs_to :author, class_name: 'User'
 
-  mount_uploader :article_header, ArticleHeaderUploader
+  mount_uploader :article_image, ArticleHeaderUploader
 
   scope :published, -> { where is_draft: false }
   scope :by_publish_date, -> { order(:published_at) }
@@ -21,7 +21,15 @@ class Article < ApplicationRecord
     self.published_at = self.is_draft ? nil : DateTime.now
   end
 
+  # This retrieves the author's name for use in the GQL output
+  # We return emptystring because that indicates to the user that this field is required
   def author_name
-    author.full_name
+    author&.full_name ? author.full_name : ''
+  end
+
+  # This retrieves the url of the highres header image for use in GQL output
+  # We let it return nil because articles are not guaranteed to have images
+  def highres_image
+    article_image&.highres&.url
   end
 end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -10,6 +10,8 @@ class Article < ApplicationRecord
 
   belongs_to :author, class_name: 'User'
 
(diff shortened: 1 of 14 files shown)


app/models/article.rb
```diff
@@ -10,6 +10,8 @@ class Article < ApplicationRecord
 
   belongs_to :author, class_name: 'User'
 
+  mount_uploader :article_image, ArticleHeaderUploader
```
kimadactyl, line 13: rename uploader to ArticleImage

alexandria-gfsc: lol oop

test/integration/graphql/article_query_type_test.rb
```diff
@@ -25,7 +27,22 @@ class ArticleIndexTest < ActionDispatch::IntegrationTest
...
+              articleBody
+
+              image
+
+              datePublished
+              dateCreated
+              dateUpdated
+
+              providers {
+                id
```
kimadactyl, line 44: i think this should link to Partners?

alexandria-gfsc: Providers are partners

test/integration/graphql/article_query_type_test.rb
```diff
@@ -25,7 +27,22 @@ class ArticleIndexTest < ActionDispatch::IntegrationTest
           edges {
             node {
               name
+              headline
+
+              author
+
               text
+              articleBody
```
kimadactyl, line 35: text and articleBody are the same i think, not sure if this matters

alexandria-gfsc: That's fine, it's a check that articleBody also represents what we want -- it could always differ as requirements change or something :woman_shrugging:

kimadactyl (approved):

PR #1125, opened 2022-04-06
Articles can have tags (part one!)

alexandria-gfsc:
- Add ArticleTags relation via a migration
- Add policy code and document existing policy funcs :)
- Add view code (split into a side by side with Partners -- text will
  need workshopping maybe)
- Add some improved coverage in the Controller tests for checking we can
  edit partner_ids and tag_ids for root + editor (WILLFIX,TODO on
  expanding that)
- Some formatting stuff


ivankocienski (approved):
Seems good to me

PR #1131, opened 2022-04-07
Filter articles by tagID graphql endpoint

ivankocienski:
Implements part of the articles by tag id filter

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -22,6 +22,12 @@ class Article < ApplicationRecord
   
   scope :with_tag, ->(tag_id) { joins(:article_tags).where(article_tags: { tag: tag_id }) }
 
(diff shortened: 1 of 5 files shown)


app/models/article.rb
```diff
@@ -19,6 +19,8 @@ class Article < ApplicationRecord
   scope :by_publish_date, -> { order(:published_at) }
 
   scope :global_newsfeed, -> { published.order(published_at: :desc) }
+  
+  scope :with_tag, ->(tag_id) { joins(:article_tags).where(article_tags: { tag: tag_id }) }
```
kimadactyl, line 23: Should also get articles by partners with the tag, but if this is too complex we can rethink this

kimadactyl (approved):

PR #1134, opened 2022-04-07
Make login and password reset processes nicer

kimadactyl:
Left to do

- [x] make *.placecal.org devise screens redirect to just placecal.org (this is because of how the default site works, if you go to https://admin.placecal.org/users/sign_in it will load the wrong background, there might be other ways to do this)
- [x] check password reset css looks right (i can't on my local)
- [x] make it redirect to admin subdomain after login
- [x] make flash message work in admin after signin

### DIFF
--- a/app/assets/stylesheets/home/_form.sass
+++ b/app/assets/stylesheets/home/_form.sass
@@ -20,6 +20,21 @@
   input
     background: transparent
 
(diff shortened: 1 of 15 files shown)


ivankocienski (approved):
Looks good to me

PR #1163, opened 2022-04-21
Fix Uglifier issue during deploy

alexandria-gfsc:
- Fix Uglifier issue
- Also compress js on production


ivankocienski (approved):
Looks good to me

Issue #1122, opened 2022-04-06
Epic: streamline calendar importer


kimadactyl:
A big list of all the user stories the Calendar admin needs to think about.

# Acceptance criteria

_Do not change these to tickets - we will use these to validate when the ticket is done manually_

I am going to convert anything in here that looks broken to a ticket. We can still validate from these points once done. - rosie

## As a Coordinator or Partner

- [ ]  I want to see what calendars I can use, and what features they support
- [ ]  I want to add a calendar feed
  - If it’s an allowlisted URL, I want confirmation that it is supported
  - If it’s a blocklisted URL
    - I want to be able to pick which of the importer options to use
    - If there is no suitable importer, I want to know why and what happens next
- [ ]  I want to see if the calendar feed is valid or not (does the URL resolve? can PlaceCal read it?)
  - If the URL doesn’t resolve, I want to be told this
  - If the calendar feed is invalid, i want to know what happens next
- [ ]  I want to see if the calendar feed is importing data (does it import?)
  - If it is, I want to preview the event data that is being imported
  - If it’s not, I want to clearly see the reason for this
  - I want to see any events which are not importing on the calendar
- [ ]  I want to set how to determine event location
  - match the event location field with Partners inside PlaceCal for each event on a calendar
  - default Partner for all events on a calendar, can be overridden as above
  - default Partner for all events on a calendar, cannot be overridden
  - default Partner for all events on a calendar, with a room number
- [ ]  I want to set how to determine if this calendar’s events are online or not
  - events are all online
  - events are online if a rule in the importer says they’re online
  - events are all offline
- [ ]  I want to set how to determine where the online event location is, if detected as online
  - set a default URL which can be overridden
  - set a default URL which cannot be overriden
  - a rule in the importer says how to determine the url
- [ ]  I want to be able to see the effects of my choice of the above options

## As a Coordinator...

- [ ]  I want to see the status of all Calendars in my Neighbourhoods
- [ ]  I want to be able to clearly see which Calendars have stopped importing properly and why
- [ ]  I want to see who to contact about the contents of a Calendar

## As a Partner...

- [ ] I want to see the status of all my Calendars
- [ ] I want to see all Events I am importing in one place
- [ ] I want to see when my Events were last updated
- [ ] I want to get a big notification if my events stop importing

## As a Developer...

- [ ]  I want to get regular confirmation my implementations are still working
- [ ]  I want to have good testing data to validate against
[...]

PR #1175, opened 2022-04-26
Refactor calendar importer

ivankocienski:
Works towards #1122


kimadactyl (changes requested):
One small tweak I think we should make now is moving the custom importers (i.e. zarts and manchester uni) into their own folder just to tidy the repo a bit?

Otherwise this looks ace and I'm excited to test it!

kimadactyl (approved):

PR #1178, opened 2022-04-26
Meetup.com events without an end time don't break the importer

### LINKED ISSUE #1177: Meetup.com teething issues
**Describe the bug**

Meetup erroring on prod

**To Reproduce**

https://rollbar.com/gfsc/placecal-production/items/123/

ivankocienski:
Meetup events with missing durations end at the same time they start
(as a default value)


kimadactyl (changes requested):
Lets assume its an hour long or i think weird things might happen

kimadactyl (approved):

PR #1180, opened 2022-04-27
Admin calendar url hints

### LINKED ISSUE #1169: Provide a list of all supported calendar feeds and the valid URL schema in the calendar creation interface

ivankocienski:
Fixes #1169

### DIFF
--- a/app/jobs/calendar_importer/parsers/eventbrite.rb
+++ b/app/jobs/calendar_importer/parsers/eventbrite.rb
@@ -6,6 +6,9 @@
 
 module CalendarImporter::Parsers
   class Eventbrite < Base
+    NAME = 'Event Brite'
+    DOMAINS = %w[www.eventbrite.com www.eventbrite.co.uk]
+
     def self.whitelist_pattern
       /^https:\/\/www.eventbrite\.(com|co.uk)\/o\/[A-Za-z0-9-]+/
     end
@@ -23,7 +26,7 @@ module CalendarImporter::Parsers
 
       loop do
         @events += results
-        results = results.next_page 
+        results = results.next_page
         break if results.blank?
       end
 
--- a/app/jobs/calendar_importer/parsers/eventbrite.rb
+++ b/app/jobs/calendar_importer/parsers/eventbrite.rb
@@ -6,7 +6,7 @@
 
 module CalendarImporter::Parsers
   class Eventbrite < Base
-    NAME = 'Event Brite'
+    NAME = 'Eventbrite'
     DOMAINS = %w[www.eventbrite.com www.eventbrite.co.uk]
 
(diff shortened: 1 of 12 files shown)


app/jobs/calendar_importer/parsers/zarts.rb
```diff
@@ -4,6 +4,8 @@
 
 module CalendarImporter::Parsers
   class Zarts < Xml
```
kimadactyl, line 6: Lets use this opportunity to convert this to a proper ticketsolve importer (i.e. adjust whitelist url, rename etc)

app/jobs/calendar_importer/parsers/eventbrite.rb
```diff
@@ -6,6 +6,9 @@
 
 module CalendarImporter::Parsers
   class Eventbrite < Base
+    NAME = 'Event Brite'
```
kimadactyl, line 9: Eventbrite one word

app/jobs/calendar_importer/parsers/ics.rb
```diff
@@ -6,6 +6,13 @@
 
 module CalendarImporter::Parsers
   class Ics < Base
+    NAME = 'ICS'
```
kimadactyl, line 9: "Generic iCal / .ics"?

kimadactyl (changes requested):
looks great just a few tweaks for you :)

app/jobs/calendar_importer/parsers/zarts.rb
ivankocienski: so rename Zarts -> TicketSolve (change filename), change the `NAME` and the `DOMAIN = %w[ticketsolve.com]` and the pattern should be something like `/^https?:\/\/([^.]*).ticketsolve.com\.*/` ?

kimadactyl: Yep exactly (no capitalised S tho by the look of it)

app/jobs/calendar_importer/parsers/manchester_uni.rb
```diff
@@ -4,6 +4,9 @@
 
 module CalendarImporter::Parsers
   class ManchesterUni < Xml
+    NAME = 'Manchester University'
```
kimadactyl, line 7: We should prob have this hidden from the list of generic importers as its a bit confusing. Maybe we could have a "show to public" bool for the importer that prevents it showing up? Or do the move to a subfolder thing I suggested before?

kimadactyl: I think my pref would prob be the latter just to make it ultra clear whats what in the codebase

kimadactyl (approved):

PR #1209, opened 2022-05-03
Allow reimport of calendars ('force_import'), make force_import default for the webui importer

alexandria-gfsc:


ivankocienski (approved):
This looks fine. At some point when we move over to activejob workers we need to think of a better way to log output as it is very useful when running on localhost but very unhelpful when running on a production machine.

PR #1218, opened 2022-05-04
Make Calendar code slightly clearer

alexandria-gfsc:
- Drastically simplified `place` strategy
- Removed some unnecessary nesting to (hopefully?) make it clearer to
  reason about
- Explicitly document race conditions in comments
- Make sure that events_from_source passes force_import (we can't
  actually let the user set this because rake argument passing isn't
  nice about this)


ivankocienski (approved):
Looks good to me

PR #1232, opened 2022-05-05
Sanitise and combine strings on iCal event description fields

### LINKED ISSUE #1208: [Bug]: Importer not properly displaying punctuation characters from certain sources
## Description

When viewing issues, certain characters like quote marks (`'`) or dashes (`-`) are replaced with `?` in the description text

## Steps to reproduce

1. View [here](https://trans-dimension.placecal.org/events/277027) and [here](https://trans-dimension.placecal-staging.org/events/277343)
2. I see weird question marks in the description in the first one
3. I see lots of `\n` characters in the second one

Screenshot from 2022-05-03 13-06-29

I should be able to see punctuation instead of question marks, and newlines instead of `\n`

Screenshot from 2022-05-03 13-06-14

Screenshot 2022-05-03 at 17 36 55

ivankocienski:
Fixes #1208

### DIFF
--- a/app/jobs/calendar_importer/calendar_importer_task.rb
+++ b/app/jobs/calendar_importer/calendar_importer_task.rb
@@ -28,6 +28,11 @@ class CalendarImporter::CalendarImporterTask
     all_event_uids = Set.new(calendar.events.upcoming.pluck(:uid))
     active_event_uids = Set.new
 
+    # this can be useful if you want to force import on your local machine
+    # if @force_import
+    #   calendar.events.destroy_all
+    # end
+
     parsed_events.each do |parsed_event|
       # is source event valid?
       # now find all occurrences of event in calendar
--- a/app/jobs/calendar_importer/events/base.rb
+++ b/app/jobs/calendar_importer/events/base.rb
@@ -15,7 +15,7 @@ module CalendarImporter::Events
     end
 
     def sanitize_invalid_char(input)
-      input = I18n.transliterate(input)
+      # input = I18n.transliterate(input)
       input.encode('utf-8', :invalid => :replace, :undef => :replace, :replace => '')
     end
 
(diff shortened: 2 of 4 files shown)


app/jobs/calendar_importer/events/base.rb
```diff
@@ -15,7 +15,7 @@ def rrule
     end
 
     def sanitize_invalid_char(input)
-      input = I18n.transliterate(input)
+      # input = I18n.transliterate(input)
       input.encode('utf-8', :invalid => :replace, :undef => :replace, :replace => '')
```
kimadactyl, line 19: while we're here can we just kill this ancient hash syntax pls :) otherwise looks good!

app/jobs/calendar_importer/calendar_importer_task.rb
```diff
@@ -28,6 +28,11 @@ def run
     all_event_uids = Set.new(calendar.events.upcoming.pluck(:uid))
     active_event_uids = Set.new
 
+    # this can be useful if you want to force import on your local machine
```
kimadactyl, line 31: can we make this a yarddoc-compatible message so it comes up in the docs when you search the function? this feels like important info basically and we should make it easier to find :)

kimadactyl (changes requested):
couple tiny changes otherwise looks good!

app/jobs/calendar_importer/events/base.rb
```diff
@@ -15,7 +15,7 @@ def rrule
     end
 
     def sanitize_invalid_char(input)
-      input = I18n.transliterate(input)
+      # input = I18n.transliterate(input)
```
kimadactyl, line 18: why is this commented out and not deleted?

kimadactyl (approved):

Issue #1233, opened 2022-05-05
Configure ActiveJob to run a background worker on dev machines and production/staging


ivankocienski:

kimadactyl:
Is this the same as #831?

PR #1237, opened 2022-05-09
Migrate to delayed_job and add ActiveJob worker

ivankocienski:
Implements #1233

### DIFF
--- a/app/views/admin/jobs/index.html.erb
+++ b/app/views/admin/jobs/index.html.erb
@@ -1,0 +1,30 @@
+<h1>Jobs</h1>
+
+<p>Currently have <%= @job_count %> jobs</p>
+
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th>ID</th>
+      <th>Priority</th>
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -53,6 +53,8 @@
       end
     end
     get 'profile' => 'users#profile', :as => 'profile'
+    get 'jobs' => 'jobs#index', as: :jobs
```
kimadactyl, line 56: lets remove this old ruby syntax

app/views/admin/jobs/index.html.erb
```diff
@@ -0,0 +1,30 @@
+<h1>Jobs</h1>
```
kimadactyl, line 1: this should prob be a datatable with limits and stuff

config/routes.rb
ivankocienski: why? i could make it into a resource- but why? this is not a long term solution

kimadactyl (approved):

Issue #1243, opened 2022-05-10
Improve visibility of Calendar importing


ivankocienski:
## User story

As a Platform admin I want to see an overview of the Calendar importing process so I can tell if problems may be arising or calendars are having problems.

As a Partner/Site admin I want to be able to see if my calendar is being imported properly or if there is a problem. I would also like to be able to queue my calendar for import (also so end users can't queue up calendars that are already in progress).

## Acceptance criteria

- [ ] jobs status page has a readout of total working jobs, in progress jobs, error'd jobs
- [ ] end users can see that their calendar is being worked on

## Implementation notes & questions

Use of state machine on calendar model.

## Implementation plan

_To be written by the developer_

kimadactyl:
Just to note theoretically I think this is all you need to monitor the status in Stimulus

```js
export default class extends Controller {
  static values = { status: String }

  statusValueChanged() {
    fetch(this.urlValue).then(/* … */)
  }
}
```

PR #1245, opened 2022-05-11
Improve feedback on calendar importer process

ivankocienski:
Implements #1243 

Calendars have clearly defined state that is easy to query for UI and for control (calendars cannot be pushed to a queue more than once)


app/helpers/calendars_helper.rb
```diff
@@ -69,4 +69,12 @@ def calendar_import_sources
       yield parser::NAME, parser::DOMAINS
     end
   end
+
+  def calendar_last_imported(calendar)
+    if calendar && calendar.last_import_at
```
kimadactyl, line 74: think this can be `if calendar&.last_import_at`

kimadactyl:
OK this looks good to me. I left one code comment for you. Otherwise I'm nervous about taking out the "rerun job" on error thing unless thats litearlly the next ticket :) (generally removing functionality bad without a good reason)

PR #1255, opened 2022-05-19
Bug: Events should not set their own address

### LINKED ISSUE #1244: [Bug]: Event strategy always prioritises Place.address if one is set
## Description

This is a hypothetical bug rather than one I can reproduce.

Aster noticed that this line means that if a place is set, it will always take precedence over the event location, even if event or override strategy is selected.

https://github.com/geeksforsocialchange/PlaceCal/blob/bbb664137d51b4bda26fb21f6495ca9961dffaba/app/models/event.rb#L158

If this is just a misunderstanding with interpretation then we need to improve the comment to show this.

- [x] If event location is set and calendar strategy is 'event' or 'override', that location should be used instead of the partner's place
- [x] if event strategy is place, then it continues to work correctly
- [x] if event strategy is override and location is not set, it continues to work correctly

ivankocienski:
When an event is imported a factory is employed to give the event the
correct place/address. Therefore the event itself no longer has this
business logic in it

Fixes #1244

### DIFF
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -225,9 +225,9 @@ class CalendarImporter::EventResolver
 
     regexp = Regexp.new(regex_string, Regexp::IGNORECASE)
 
-    @event_location_components = data.location
-                                     .split(', ')
-                                     .map { |component| component.gsub(regexp, '').strip }
-                                     .reject(&:blank?)
+    @event_location_components = (data.location || '')
+      .split(', ')
+      .map { |component| component.gsub(regexp, '').strip }
+      .reject(&:blank?)
   end
 end
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -118,7 +118,7 @@ class CalendarImporter::EventResolver
       if place.present?
         # place = 'calendar.place'
         # address = 'calendar.place.address'
-        place = calendar.place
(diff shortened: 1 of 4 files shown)


test/jobs/calendar_importer/event_resolver_strategy_test.rb
```diff
@@ -0,0 +1,275 @@
...
+    # theory of test
+    #   given
+    #     data location is missing
+    #     calendar strategy is 'event_override'
+    #
+    #   then
+    #     event place = calendar place
+    #     event address = calendar place address
+
+    data = FakeEvent.new(
```
alexandria-gfsc, line 164: If you're using the same `data` for each test and only at most changing one or two values: I would put it into the setup and change that value on the struct in the test. This should just make it a little bit clearer that it is either the same structure, or mostly the same :)

ivankocienski: fixed

app/jobs/calendar_importer/event_resolver.rb
```diff
@@ -118,7 +118,7 @@ def event_override_strategy(place, address: nil)
       if place.present?
         # place = 'calendar.place'
         # address = 'calendar.place.address'
-        place = calendar.place
+        # place = calendar.place
```
alexandria-gfsc, line 121: What's the reason for commenting out handling of `place` after checking whether it's present or not?

alexandria-gfsc (approved):
lgtm

PR #1263, opened 2022-05-26
Map not showing bug

### LINKED ISSUE #1259: [Bug]: Maps are not working in the PlaceCal frontend
## Description

When I look at maps for individual partners or the partners page, the map does not work

Update: they seem to not load until you refresh the page

## Steps to reproduce

1. Go to https://hulme.placecal.org/partners and scroll to the bottom
1. Go to https://hulme.placecal.org/partners/zion-community-resource-centre
2. Go to https://hulme.placecal.org/events/289561

A map should appear in the empty spaces

Screenshot 2022-05-24 at 13 03 38
Screenshot 2022-05-24 at 13 03 32

ivankocienski:
Fixes #1259 

Due to the use of Turbo in rails where pages are long lived, the map component JS was getting confused and not showing properly. This PR introduces a new map management script and uses some tricky turbo page loading fu to move the map around the page without needing to reload


app/assets/stylesheets/globals.sass
```diff
@@ -15,6 +15,7 @@
 @import "base/pres"
 @import "base/typography"
 @import "base/icons"
+@import "base/map"
```
kimadactyl, line 18: alphebetise these, i think icons got chucked on the end too

app/helpers/map_helper.rb
```diff
@@ -0,0 +1,62 @@
+
+module MapHelper
+  API_TOKEN = 'pk.eyJ1IjoicGxhY2VjYWwiLCJhIjoiY2ptdzJqM3owMzN1bDNwbnhjbHIzb25layJ9.Kq2KjkWzSvLOpiHICuJiPA'
```
kimadactyl, line 3: We should move this to an environment variable tbh

config/database.yml
```diff
@@ -11,7 +11,7 @@ default: &default
 
 development:
   <<: *default
-  database: placecal_dev
```
kimadactyl, line 14: erm can we leave this as it is or is there a reason for the change?

app/assets/stylesheets/globals.sass
ivankocienski: probably not a good idea if they are order dependent?

kimadactyl: its for code organisation reasons not anything technical :) otherwise as the file grows it can be hard to find an include

kimadactyl (approved):
happy to merge when that one thing is fixed tho

Issue #1119, opened 2022-04-06
Add News to site frontend


kimadactyl:
**Is your feature request related to a problem? Please describe.**

We now have a lovely news feature that exports to an API for Trans Dimension!

We now need to make it show up on the regular *.placecal.org frontends.

**Describe the solution you'd like**

- [ ] Add 'News' menu option on subsites that shows up if there is at least one news article
- [ ] Show News from all partners within the site's scope
- [ ] We have a design somewhere in Google Drive

PR #1265, opened 2022-05-30
News UI frontend

ivankocienski:
Implements #1119 

- Routes
- Controllers
- Views
- DB scopes
- some tests


app/controllers/application_controller.rb
```diff
@@ -211,7 +211,8 @@ def set_navigation
                     [
                       ['Events', events_path],
                       # ['Places', places_path],
-                      ['Partners', partners_path]
+                      ['Partners', partners_path],
+                      ['News', news_index_path]
```
kimadactyl, line 215: should only show news link if there is at least one valid news article

app/controllers/news_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class NewsController < ApplicationController
+  before_action :set_article, only: %i[show]
+  before_action :set_site
+
+  def index
+    @articles = Article
+      .for_site(@site)
+      .published
+      .order(:published_at)
```
kimadactyl, line 11: we have a scope for this - `by_publish_date`

also needs pagination i think from now. so either make a new ticket or do it as part of this one, i think its essential

app/controllers/application_controller.rb
ivankocienski: Can do but that'll be a hit every time someone loads any page on the site

kimadactyl: maybe one for babys first caching experiment?

kimadactyl: oh or we could just add a bool to sites thats like "news_enabled?" maybe easier?

kimadactyl (approved):
too long didn't read

PR #1267, opened 2022-05-31
Fix bug where users were unable to edit a partner without a common ward

alexandria-gfsc:
- fix: users unable to edit partner without shared ward
- refactor: parts of controller test for ease of reading
- fix: event factories sequential summary generation


ivankocienski (approved):
Looks okay to me

PR #1270, opened 2022-06-06
Importer not logging importer used

### LINKED ISSUE #1261: [Bug]: importer_used is not set properly on calendars
## Description

For a non-ICal importer, we should be setting `importer_used`. 

## Steps to reproduce

1. Run the importer
2. Open the console and type `Calendar.where(importer_used: nil).count`
3. Observe that there are 53 calendars on Staging (And 63 locally?) that do not have `importer_used` set

The `importer_used` field should be set properly for all the calendars we have.

ivankocienski:
Fixes #1261 - calendar importer not changing state correctly if an import has no events

### DIFF
--- a/app/jobs/calendar_importer/calendar_importer_task.rb
+++ b/app/jobs/calendar_importer/calendar_importer_task.rb
@@ -11,79 +11,77 @@ class CalendarImporter::CalendarImporterTask
     @force_import = force_import
   end
 
-  # the main importing function
   def run
-    notices = []
-
-    parser = CalendarImporter::CalendarImporter.new(calendar).parser
-
-    calendar_source = parser.new(
-      @calendar,
-      from: from_date,
-      force_import: force_import
-    ).calendar_to_events
-
-    return if !@force_import && calendar.last_checksum == calendar_source.checksum
-
-    parsed_events = calendar_source.events.map do |event_data|
-      CalendarImporter::EventResolver.new(event_data, calendar, notices, from_date)
-    end
+    parsed_events = event_data_from_parser
 
     if parsed_events.present?
-
-      all_event_uids = Set.new(calendar.events.upcoming.pluck(:uid))
-      active_event_uids = Set.new
-
       # this can be useful if you want to force import on your local machine
       # if @force_import
       #   calendar.events.destroy_all
       # end
 
       parsed_events.each do |parsed_event|
-        # is source event valid?
-        # now find all occurrences of event in calendar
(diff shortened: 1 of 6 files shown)


app/jobs/calendar_importer/calendar_importer_task.rb
```diff
@@ -29,53 +29,54 @@ def run
...
-      # next if parsed_event.is_address_missing?
-
-      active_event_uids << parsed_event.uid
-
-      parsed_event.save_all_occurences
-
-    rescue CalendarImporter::EventResolver::Problem => e
-      notices << e.message
-    end
+    if parsed_events.present?
```
kimadactyl, line 32: This is a very big method could it be broken up a bit to make it easier to read what's going on? Maybe the big if loop in a helper method?

app/jobs/calendar_importer_job.rb
```diff
@@ -9,6 +9,11 @@ class CalendarImporterJob < ApplicationJob
     report_error exception, "Calendar URL is not accessible"
   end
 
+  rescue_from ActiveRecord::ActiveRecordError do |exception|
+    raise exception if Rails.env.development? && @silence_db_exceptions == false
```
kimadactyl, line 13: Assume you've thought about this but does this not want to activate for env == test too?

app/jobs/calendar_importer/calendar_importer_task.rb
ivankocienski: I agree it is a touch long but 30% of it is comment explaining the algorithm. Could be moved into helper methods but would that improve parsability? Much of the fiddly logic from the original has been moved into the EventResolver helper class already.

kimadactyl: i think the thing that increases parsability is simply having a method named say, "make_events_do_good_thing" so you can then just read the method name and not to get into the weeds. if that makes sense! almost kinda like writing in pseudocode except its not pseudocode?

kimadactyl: like yes it might increase the overall LoC but just having a simple human parsable top level method makes it a ton easier to remember whats going on i think

ivankocienski: i believe you are describing "self documenting code". i will have a stab at it.

yeah rubocop is way too strict out of the box but its better to be too strict and loosen up than have nothing and add rules (like- otherwise why bother?)

Issue #1028, opened 2022-03-07
Un-break static tags on placecal site homepages


kimadactyl:
The way the "internet access" block on say hulme.placecal.org works is a mess and the logic is split between the database and hard coded in the view. This is _not great_.

We might also want to add blocks to this section for e.g. unisex/accessible toilets, free wifi, etc. Hard coding in the tag id seems like its not the one.

In terms of user story I guess this is something like:

- As the PlaceCal team, we want certain important information that is useful everywhere like where has internet or wifi or accessible toilets prominantly placed on the Site homepages, in order that people can get access to support as easily as possible
- As a developer, I want to be able to use the existing tagging system to store this information, as adding a shitload of bools that are not user editable sounds like a fools errand when we have a perfectly functional tagging system
- As a developer, I don't want the pages this information is used to break if the database values change or something and so we need a proper registrty for reserved strings

The current names and slugs for these tags suck too. It should prob be "wifi available [wifi-available]" and "computers available [computer-access]" or something

<img width="326" alt="Screenshot 2022-06-07 at 13 43 00" src="https://user-images.githubusercontent.com/511734/172381682-e295c37a-6840-4f1b-9bd8-993b28cd043d.png">

## Ways this could be done

- **Add 'system' tags class or something like it that can't be changed and hard wire them (easy)** (selected solution)
- Add extra fields to tags like css class or whatever and 'can display on homepage' to allow us to sorta hack it in place (medium)
- Abandon tag system for this after all and add new properties to the partner model (medium)
- Add a new model for these home page blocks and allow them to be selected from a list in the site config (hard)

## Acceptance criteria

- [x] Tags can be prevented them from being edited with a tick box that locks them down
- [x] Split current internet access tag into "Public WiFi [wifi]" and "Computer Access [computers]" on staging and prod. Feel free to just unset all current assignments
- [x] #1276
- [x] Change frontend hard coded slugs to the new ones
- [x] #1279

PR #1274, opened 2022-06-08
Allow tags to be set as uneditable

ivankocienski:
Implements #1028 

Tags now have a system_tag flag. System tags can only be modified by root users.


kimadactyl (changes requested):
Looks good. Let's just disable access for everyone for slug and title if its a system tag

kimadactyl (approved):

Issue #1279, opened 2022-06-09
Reinstate 'Computer Access' info card


ivankocienski:
On Sites#index page- bring back the 'Computer Access' card. Also make the 'wifi' and 'computers' cards optionally render if they have content. Fix CSS to handle the four boxes in the layout

PR #1280, opened 2022-06-09
Site tag cards

ivankocienski:
Implements #1279

Cards covering computer access and public wifi now show up on sites index page (if partners are present with that tag)


kimadactyl:
Will do CSS on monday

PR #1291, opened 2022-06-13
Remove HTML from imported events

### LINKED ISSUE #1196: Remove HTML from imported event descriptions
**Describe the bug**

HTML from external sources is being directly added to PlaceCal which is confusing matters as we assume we're using markdown.

[Issue on TransDim](https://github.com/GenderedIntelligence/the-trans-dimension/issues/176#issuecomment-1096592250)

**To Reproduce**

1. Go to https://transdimension.netlify.app/events/232378

**Expected behavior**

Convert source HTML to markdown before saving it

**Screenshots**

image

ivankocienski:
Fixes #1196

### DIFF
--- a/app/jobs/calendar_importer/events/base.rb
+++ b/app/jobs/calendar_importer/events/base.rb
@@ -26,24 +26,24 @@ module CalendarImporter::Events
     # Strip out all shady tags
     # Convert all html to markdown
     def html_sanitize(input)
-      return if input.blank?
+      return '' if input.blank?
 
-      allowed_tags = %w[a strong b em i ul ol li blockquote h3 h4 h5 h6 br]
+      clean_input = sanitize_invalid_char(input)
 
-      str = Nokogiri::HTML.fragment(input)
-      str.css(*['h1', 'h2']).each { |header| header.name = 'h3' }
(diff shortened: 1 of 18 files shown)


app/jobs/calendar_importer/events/base.rb
```diff
@@ -26,24 +26,24 @@ def sanitize_invalid_char(input)
...
+        doc << '<br/><br/>'
+        doc << footer
       end
 
-      str = str.to_s
+      body_text = doc.serialize
 
-      str = ActionController::Base.helpers.sanitize(str, tags: allowed_tags)
-      str = sanitize_invalid_char(str)
+      allowed_tags = %w[p a strong b em i ul ol li blockquote h3 h4 h5 h6 br]
```
kimadactyl, line 43: This feels like something we should have stored somewhere thats not hidden in a function. Maybe just move to a const?

kimadactyl:
This looks really good but would like to see some more tests for different kinds of input. Maybe

- Something without proper closing tags
- Something with forbidden tags to check they get stripped
- Something with CSS classes to check they get stripped
- Markdown
- Anything else you can think of that people might fuck up somehow?

PR #1292, opened 2022-06-13
Admin logout redirect

### LINKED ISSUE #1286: [Bug]: Redirect to login or placecal.org when sign out of admin rather than attempt to reload page resuting in 401
## Description

Sign out of admin.placecal.org should not result in reload of the page I was on.

## Steps to reproduce

1. SIgn in to placecal.org
1. Press the sign out button
2. results in 401 error and no visible change

- Redirect me to https://placecal.org/users/sign_in
- or - 
- Redirect to https://placecal.org

ivankocienski:
Fixes #1286 

both log in and log out buttons

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -10,17 +10,22 @@ class Users::SessionsController < Devise::SessionsController
   protected
 
   # when we have authenticated the user take them to the admin site
-  # 
+  #
   # == Parameters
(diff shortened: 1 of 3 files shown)


kimadactyl:
I believe this works but it feels bad. Like do we really have to disable turbo for something so fundamental to a functioning app? Or are we missing something? I'm happy to merge this but it feels insane if this is the solution?

kimadactyl (approved):

PR #1305, opened 2022-06-15
Another fix for rake task that populates historical html

ivankocienski:


katjam (approved):
Let's hope this helps us get through the rest.

Issue #1307, opened 2022-06-15
[Bug]: Selecting a date to filter events by on a partner pages scolls to top


katjam:
## Description

On a partner page (e.g. https://moss-side.placecal.org/partners/52), if I use the filter controls for event listing, when the page reloads with my results, I land at the top and have to scroll to see events.

This is mildly confusing but not urgent.

![image](https://user-images.githubusercontent.com/861172/173832025-dc9636cd-7093-4b0e-b1a2-528747b7a7ec.png)

kimadactyl:
can do this as part of #1306 it should just be a simple html anchor right?

ivan-kocienski-gfsc:
Verified [here](https://moss-side.placecal.org/partners/37/events/2024/3/18?period=week&sort=#paginator) that this is fixed.

(Although it would be better to use hotwire to replace the contents of the filter results box)

PR #1312, opened 2022-06-15
Improve event display when there are limited events

### LINKED ISSUE #1306: Event view for partners with no or few events is confusing
## User story

1. As a resident I need to be able to see if partners have no calendar connected so that I don't just assume there are no events
1. As a resident I need to be able to see if partners have no upcoming events so that I know I can check back later and find out
1. As a resident I need to be able to see the next few upcoming events for partners who only put a few things on, so that I can see they do indeed have events without clicking around for ages

## Acceptance criteria

- [x] Partners with no calendars connected have no "events" section on their page and say something like "this partner does not list their events on placecal - try their website or social media"
- [x] Partners with a calendar connected and no upcoming events have no "events" section and say "no upcoming events"
- [x] Partners with less than 10 upcoming events are presented in a simple list rather than using the week picker
- [x] Partners with more than 10 upcoming events have the week picker but the next upcoming event is selected, OR we move to simple pagination

Can fix #1307 while we're here

## Implementation plan

1. Check how many future events the controller wants to load for the partner view page
2. Load a different component depending on the upcoming event count

kimadactyl:
Fixes #1306 and  #1307

### DIFF
--- a/app/controllers/partners_controller.rb
+++ b/app/controllers/partners_controller.rb
@@ -7,6 +7,8 @@ class PartnersController < ApplicationController
   before_action :set_site
   before_action :set_title, only: %i[index show]
 
+  PAGINATION_THRESHOLD = 20
+
   # GET /partners
   # GET /partners.json
   def index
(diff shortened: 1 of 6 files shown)


app/controllers/partners_controller.rb
```diff
@@ -29,21 +31,29 @@ def index
...
   # GET /partners/1.json
   def show
-    # Period to show
-    @period = params[:period] || 'week'
-    @events = filter_events(@period, partner_or_place: @partner)
-    @opening_times = @partner.human_readable_opening_times
-
-    # Map
-    if @events&.length.positive?
-      @map = get_map_markers_from_events(@events)
```
kimadactyl, line 39: Just a note i conciously cut this because idk if it even works and it seems overkill

PR #1320, opened 2022-06-16
Basic news summary on for article index page

### LINKED ISSUE #1300: [Bug]: Articles need summaries
## Description

When viewing a sites news articles, summaries are missing from the page.

## Steps to reproduce

1. Make article for site with body content
2. navigate to sites news page
3. only article titles are shown

On sites news index: should see summary text (a snippet of text from the body).

ivankocienski:
Fixes #1300 in a fast and simple way

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -42,14 +42,6 @@ class Article < ApplicationRecord
       .where(address: { neighbourhood_id: site_neighbourhood_ids } )
   }
 
-  def summary_text
-    body_str = body.to_s.strip
-
-    return body_str if body_str.length < 200
-
-    body_str[0...200]
-  end
-
   def update_published_at
     self.published_at = self.is_draft ? nil : DateTime.now
   end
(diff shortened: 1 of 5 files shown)


app/views/news/index.html.erb
```diff
@@ -16,7 +16,7 @@
 
         <div class='g articles__content'>
           <div class='gi articles__body'>
-            <% #= markdown article.body[0..200] %>
+            <p><%= article.summary_text %></p>
```
kimadactyl, line 19: i think actual truncate call should be here, and the other method just call like "plain_text" or something

app/models/article.rb
```diff
@@ -42,6 +42,14 @@ class Article < ApplicationRecord
       .where(address: { neighbourhood_id: site_neighbourhood_ids } )
   }
 
+  def summary_text
+    body_str = body.to_s.strip
```
kimadactyl, line 46: could we maybe also just remove all "#" charactors?

Issue #1330, opened 2022-06-20
Show News Article page styling


ivankocienski:
Make it look more in-line with the other show-resource page for public users

PR #1331, opened 2022-06-20
Show article page layout and styling

ivankocienski:
Is now more inline with the show partner/event

Implements #1330


katjam (approved):
I think this is good to go in.

PR #1348, opened 2022-06-22
Only root can edit users' neighbourhoods

### LINKED ISSUE #1104: Investigate neighbourhood setting breakage on a new user
**Current Behaviour**
- Make a new user using the `users/create` screen. Give them a neighbourhood region (like `london, england`)
- Log in as new user
- User is missing `neighbourhood_admin` permission and has no neighbourhood regions assigned to them  

**Intended Behaviour (Checklist for testing)**
- [ ] Create a new user, give them a neighbourhood region (like `london, england`)
- [ ] Log in as user
- [ ] User has `neighbourhood_admin` permission and has neighbourhood region assigned to them

**Approval Criteria**
- [ ] I can make a new user, assign one or more neighbourhoods to them, when they log in they have those neighbourhoods assigned and have the `neighbourhood_admin` permission set to true

ivankocienski:
Fixes #1104

### DIFF
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -19,11 +19,25 @@
     </div>
     <div class="col-md-6">
       <h3>Neighbourhoods</h3>
-      <p>Grants access to every partner and calendar in these neighbourhoods. Use with care!</p>
-      <%= f.association :neighbourhoods,
-                        label: false,
-                        collection: options_for_neighbourhoods,
-                        input_html: {class: 'form-control select2' } %>
+      <p>You have <%= pluralize @user.neighbourhoods.count, 'neighbourhood' %>.</p>
+
+      <% if f.object.root? %>
+        <p>Grants access to every partner and calendar in these neighbourhoods. Use with care!</p>
+
+        <%= f.association :neighbourhoods,
+          label: false,
+          collection: options_for_neighbourhoods,
+          input_html: {class: 'form-control select2' } %>
+
+      <% else %>
+        <ul class='neighbourhood-list'>
+        <% @user.neighbourhoods.order(:name).each do |hood| -%>
+          <li><%= link_to hood.name, admin_neighbourhood_path(hood) %></li>
+        <% end -%>
+        </ul>
+
+    <% end %>
+
     </div>
   </div>
   <div class="row">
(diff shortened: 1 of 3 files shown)


app/views/admin/users/_form.html.erb
```diff
@@ -19,11 +19,25 @@
...
     <div class="col-md-6">
       <h3>Neighbourhoods</h3>
-      <p>Grants access to every partner and calendar in these neighbourhoods. Use with care!</p>
-      <%= f.association :neighbourhoods,
-                        label: false,
-                        collection: options_for_neighbourhoods,
-                        input_html: {class: 'form-control select2' } %>
+      <p>You have <%= pluralize @user.neighbourhoods.count, 'neighbourhood' %>.</p>
+
+      <% if f.object.root? %>
```
kimadactyl, line 24: Is there a reason to use this method and not `if current_user.root?`?

kimadactyl: yeah i think this will allow changing this based on the user you are editing's permissions, not yours

app/views/admin/users/_form.html.erb
```diff
@@ -19,11 +19,25 @@
     </div>
     <div class="col-md-6">
       <h3>Neighbourhoods</h3>
-      <p>Grants access to every partner and calendar in these neighbourhoods. Use with care!</p>
-      <%= f.association :neighbourhoods,
-                        label: false,
-                        collection: options_for_neighbourhoods,
-                        input_html: {class: 'form-control select2' } %>
+      <p>You have <%= pluralize @user.neighbourhoods.count, 'neighbourhood' %>.</p>
```
kimadactyl, line 22: this doesn't really make sense because this is the form to edit the user, not the form to edit yourself

kimadactyl:
OK I think there's a bit of a mixup going on here

`/users/xx/edit` is the intended place for admins to configure other users and should be targetted at roots and neighbourhood admins

`/profile` is where people are intended to view and change their own profile and is aimed at everyone

I think citizens shouldn't be able to see `/admin/xx/edit` at all

kimadactyl:
Also lets fix this typo while we're here: "Partners this user can edit, as well as their **associatied** calendars"

app/views/admin/users/_form.html.erb
ivankocienski: are you negating your earlier comment? if so then just remove the earlier comment.

kimadactyl: no, the comments are in agreement, the first one is the method this should be using and the second one is why it's wrong

kimadactyl:
Scrapping this per our discussion

PR #1349, opened 2022-06-22
Change of text on admin edit partner page.

### LINKED ISSUE #1272: [Bug]: New partners should tell you if they won't show up on any sites
## Description

Georgie couldn't work out why a new partner wasn't showing events. It turned out it was because the appropriate tag hadn't been selected. This is a UX bug as it's not at all clear why or what to do about it.

## Steps to reproduce

1. Create a partner outside a current PlaceCal site, or without the right tag, etc
2. Partner will be created and give you a message that it shows up on no sites but not know what to do about it

Basically be a bit more helpful.

- Partners could warn you at creation that they're not going to show up
- The "this doesnt show up on any sites" text could be more helpful

## Screenshots

WhatsApp Image 2022-06-07 at 10 31 08 AM

ivankocienski:
Fixes #1272

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,5 @@ doc/**/*.html
 doc/**/*.css
 doc/**/*.js
 .yardoc
+tags
+Makefile
--- a/app/views/admin/partners/edit.html.erb
+++ b/app/views/admin/partners/edit.html.erb
@@ -19,7 +19,12 @@
     <% if partners_sites_links.present? %>
       <p>This partner appears on the following sites: <%= site_links %>.</p>
     <% else %>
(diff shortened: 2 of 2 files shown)


kimadactyl:
Hmm I think we could bit a bit more verbose. How about: 

```
This partner does not appear on any sites. This usually means one of two things:

* The address is outside any current PlaceCal instance's range due to where actual ward boundaries happen to fall. Please contact your PlaceCal organiser if this is the case.
* It is inside a PlaceCal range, but you forgot to add a tag. Resave this partner with the appropriate tag.
* If neither of these are the case, please contact support@placecal.org with the link to this page
```

kimadactyl (approved):

PR #1353, opened 2022-06-23
Admins updating neighbourhoods

### LINKED ISSUE #1352: Users with roles should only have access to certain admin pages based on those roles
## User story

As a root user I want to be able to modify any resource in the admin panel.

## Spec from Kim:
- partner admin and below cant see any user edit stuff and get redirected to profile if they try
- info about what neighbourhoods you have access to goes on profile (and edit neighbourhoods field is removed for everyone)
- only roots can edit neighbourhood access and they can only do it on the user edit pages
- neighbourhood admins can see user edit pages but not edit neighbourhoods (list them out instead)

ivankocienski:
Fixes #1352


app/views/admin/users/_form.html.erb
```diff
@@ -9,21 +9,35 @@
...
     </div>
+
     <div class="col-md-6">
       <h3>Neighbourhoods</h3>
-      <p>Grants access to every partner and calendar in these neighbourhoods. Use with care!</p>
-      <%= f.association :neighbourhoods,
-                        label: false,
-                        collection: options_for_neighbourhoods,
-                        input_html: {class: 'form-control select2' } %>
+      <p>This user has <%= pluralize @user.neighbourhoods.count, 'neighbourhood' %>.</p>
```
kimadactyl, line 23: I think just change this for a list of them. Maybe this should be the else block for the following? So if root show the picker, if not show the list of neighbourhoods.

kimadactyl (approved):
approving with the caveat @katjam is going to help test this before release to prod :)

PR #1368, opened 2022-07-04
fix: Root users incorrectly getting warning they have no rights

ivankocienski:
Fixes problem with original PR


app/helpers/users_helper.rb
```diff
@@ -15,6 +15,7 @@ def options_for_roles
   end
 
   def user_has_no_rights?(user)
+    return false if user.root?
```
kimadactyl, line 18: prob need to add editor in here too then this is good to go

kimadactyl (approved):

PR #1369, opened 2022-07-05
Article::for_site finds by tag and neighbourhood

### LINKED ISSUE #1332: [Bug]: News article lists not filtering correctly
## Description

Looking at news for The Trans Dimension

## Steps to reproduce

1. visit https://trans-dimension.placecal-staging.org/news
2. See the wrong articles: 
image

See these articles
image

Also on placecal.org/news even though we do not expect people to find it, this should list all news. At the moment only lists one item

ivankocienski:
Fixes #1332

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -27,20 +27,48 @@ class Article < ApplicationRecord
 
   scope :global_newsfeed, -> { published.order(published_at: :desc) }
 
-  scope :with_tag, ->(tag_id) { joins(:article_tags).where(article_tags: { tag: tag_id }) }
-
   scope :with_partner_tag, lambda { |tag_id|
     joins('left outer join article_partners on articles.id=article_partners.article_id')
     .joins('left outer join partner_tags on article_partners.partner_id = partner_tags.partner_id')
     .where('partner_tags.tag_id = ?', tag_id)
   }
 
-  scope :for_site, lambda { |site|
-    site_neighbourhood_ids = site.owned_neighbourhoods.pluck(:id)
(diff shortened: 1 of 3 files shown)


app/models/article.rb
```diff
@@ -27,20 +27,48 @@ class Article < ApplicationRecord
...
-      .distinct
+    site_neighbourhood_ids = site.owned_neighbourhoods.pluck(:id)
+    site_tag_ids = site.tags.pluck(:id)
+    return scope if site_neighbourhood_ids.empty? && site_tag_ids.empty?
+
+    where_fragments = []
+    where_params = []
+
+    # articles by neighbourhood
+    if site_neighbourhood_ids.any?
```
alexandria-gfsc, line 51: im wondering if its possible at all to wrap these? only because atm the cognitive load of reading this block with the long sql statements is very high

alexandria-gfsc (changes requested):
I feel like the current way you're doing the scoping SQL is kind of fragile -- it took me a while to figure out what exactly was going on with the `where_fragments`, etc., and I don't think we should be writing code that glues together bits of SQL, that then gets glued together and passed into a helper that then gets run. The pipeline there is too long and it makes it obtuse to debug and inspect. 

At best, we should have a comment explaining what we are doing (lowering the cognitive load) and why -- convince me that there's no other approach

kimadactyl (approved):

alexandria-gfsc (approved):
(zardoz voice) THE AR `OR` IS BAD, THE COMMENT IS GOOD

PR #1403, opened 2022-07-25
Make multi-day events show properly in interface (#1214)

### LINKED ISSUE #1214: Full support for multi-day events
## User story

As someone who hosts multi day events, I want to be able to upload things spanning multiple days, so that they show up properly on each day they are active

## Acceptance criteria

- [ ] Event index view shows events on every day they are active
- [ ] Event index view shows it the duration as the whole event length (i.e. "3 days" not "24 hours")
- [x] Events under 24 hours show up as an hour and minute count (same behaviour as before)
- [ ] Events that are 24 or more hours show up as a rounded up whole number of days (25 hours is 2 days, 48 hours is 2 days)
- [ ] Events like clubnights that cross a day boundary but are less than 24h show up as a number of hours properly (so they are in 2 diff days, but still have a normal amount of hours)
- [x] Event show view shows the day range and start and end time for the whole event

## Implementation plan

- The `find_by_day` scope is amended to include multi-day events
- I can think of two ways to amend the UI.
1. We remove the "1 day" bit and have the time listing be something like `12th July 12:00 - 14th July 13:00`. Then for an ordinary event where the end date is on the same day we have it just say `12th July 12:00 - 13:00`
2. We alter the "1 day" bit to show the correct number of days

alexandria-gfsc:
Fixes #1214

Wasn't sure who specifically to assign to review so - you get assigned, and you get assigned, and you ALL get assigned

### DIFF
--- a/app/components/event/event_component.rb
+++ b/app/components/event/event_component.rb
@@ -25,10 +25,11 @@ class EventComponent < MountainView::Presenter
 
   def duration
     return false unless event.dtend
+
     mins = ((event.dtend - event.dtstart) / 60).to_i
     hours = mins / 60 # Ruby presumes ints not floats, and rounds down
 
-    if hours < 24
+    if hours < 25
       mins_str = (mins % 60).positive? ? "#{mins % 60} mins" : ''
       hours_str = hours.positive? ? pluralize(hours, 'hour') : ''
       [hours_str, mins_str].reject(&:empty?).join(' ')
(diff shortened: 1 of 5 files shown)


alexandria-gfsc:
This is currently blocked pending feedback on the UI plan as listed in the issue ticket

ivankocienski (approved):
Looks okay to me

PR #1409, opened 2022-08-01
Refactor select2 javascript using Stimulus and add system tests for select2 inputs

### LINKED ISSUE #1387: transfer select2 functionality from various behaviors js files into it's own stimulus controller
## User story

As a developer adding select2 to my forms I want one place and method of working so that I don't have to worry about which loading events effect which form elements and can limit amount of code I need to read when dealing with bugs.

## Acceptance criteria

- [ ] All Select2 logic is removed from behaviors folder.
- [ ] A Stimulus controller is created that initializes Select2 elements.
- [ ] Clicking the back button no longer breaks Select2 inputs
- [ ] A system test is written to prove that it is working.

- I am not sure exactly what will happen when Cocoon inserts extra form elements, hopefully that is specified by the template and should not be an issue.
- Will need to decide how to assign `multiple`.

## Implementation plan

- At this stage I think it is likely that in the HTML `class="select2" will be removed in favor of `data-controller="select2"` which should remove the need for event listeners.
- I imagine that the controller code will be very similar to the internals of the forEach loops that exist to handle this already.

aaaaargZombies:
fixes #1387

- Removes select2 functionality from behaviors folder
- Updates views to use `data-controller="select2"` instead of `class="select2"`
- Adds a test for each select2 input


aaaaargZombies:
I've left stuff to do with "collections" alone because @kimadactyl said this had been moth balled.

kimadactyl (commented):
This is really great and works a treat. A significant hurdle in our codebase.

My only comments are about idiomatic rails stuff and organisation:

- [ ] All admin area tests need moving from `test/system` to `test/system/admin` to match the other folders, with test class names changed too (see other folders for inspo)
- [ ] `x_form_test` etc could be better named and relate to the task that is being done, e.g. `create_article_test`, `edit_article_test` etc. Or simply `article_test`. Basically `form` doesn't add much!

Apart from that this looks ace!

kimadactyl (approved):

kimadactyl:
Havn't been exhaustive but this seems working and much happier.

I can go to partner -> add new partner admin, press back, and not have all the select2s break.

Issue #1413, opened 2022-08-01
Write system tests for sites dashboard


lexiwitch:
## User story

As a developer, I want system tests for the sites dashboard, in order that if anything breaks I know about it before it hits prod

## Acceptance criteria

There are tests for https://admin.placecal.org/sites that validates there is...

- [ ] A table with the correct column headings
- [ ] A link to the article in the "name" field with the correct title and link destination
- [ ] Text for the name and the slug
- [ ] Last updated looks as it should (e.g. 2 months ago not raw datetime)
- [ ] Clicking "show" and changing the number updates the records shown (note: need to generate lots of records for this)
- [ ] Typing something in "search" correctly filters the list
- [ ] There is a paginator and clicking it changes the page

## Implementation notes & questions

This should solely validate the current functionality so we get notified if it changes

## Implementation plan

_To be written by the developer_

r-ferrier:
closing as wontfix

PR #1425, opened 2022-08-02
Remove unused behaviors.site.js code

### LINKED ISSUE #1423: Replace `behaviors.site.js`
## User story

As a developer I need move JS functionality to stimulus or onto the backend so that it is predictable and easier to reason about.

## Acceptance criteria

- [ ] `behaviors.site.js` is deleted
- [ ] The undesirable input is not accessible to the user.

- The only thing left in this file is a function that deletes some form fields
- It effects "Other neighbourhoods to include"
- [behaviors.site.js](https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/javascript/src/behaviors/behaviors.site.js)
- This looks like it should be handled in the template but I'm sure there was a reason it wasn't
- part of #1389

## Implementation plan

Either a stimulus component or never render the input if it is not needed.

aaaaargZombies:
fixes #1423
It turns out someone had already done in the view but left the JS in, so an easy fix for me :)


kimadactyl:
I'm good to cut stuff but let's leave a few tests where we go.

For e.g. #1413? Altho tbh this might be covered in datatables? I think just you know, any testing good here?

aaaaargZombies:
This only effects the edit/new site form
I can add a test to make sure we don't render an imput for "main neighborhood" on the edit page though?

And see why this test is breaking!

kimadactyl:
yeah fair i think test for cocoon working as thats what youre removing?

are you sure that code does nothing :)

aaaaargZombies:
Noop I was wrong about it not doing anything - and about what I thought it did initially. 

At least the tests work!

kimadactyl:
yeah tests ftw!!

PR #1426, opened 2022-08-02
Refactor conditional rendering for primary neighbourhood in site admin

### LINKED ISSUE #1423: Replace `behaviors.site.js`

aaaaargZombies:
fixes #1423

**Before**
- A class was conditionally added to form fields.
- If this class appeared on the front end the element would be deleted via JS.

**After**
- Only render needed form fields.
- Add test to ensure this is the case.


kimadactyl (approved):

kimadactyl:
Hard to validate a negative but all sites forms seem working on staging

Issue #1429, opened 2022-08-08
Delete behaviours folder and tidy up views


aaaaargZombies:
## User story

As a developer I need remove old code that we no longer use so that other work can continue unimpeded.

## Acceptance criteria

- [ ] Passing tests
- [ ] PR #1424 should already be merged
- [ ] views should not contain `<div class="behavior-trigger" data-behavior="user.form"></div>` etc
- [ ] behaviours js files and related imports removed 

## Implementation notes & questions

This would break select2 inputs on "collections" but this part of the site has been abandoned.

It will resolve the following individual issues: #1419, #1420,  #1422 but will avoid writing more repetitive tickets.

## Implementation plan

- Remove class and data bindings from views
- Remove imports from `admin.js`
- Delete behaviors folder.

Issue #1422, opened 2022-08-01
remove unused code from `behaviors.partner.js`


aaaaargZombies:
## User story

As a developer I need to not have to worry about unused code so that I can work on more important things.

## Acceptance criteria

- [ ] code is deleted

## Implementation notes & questions

- [behaviors.partner.js:L24](https://github.com/geeksforsocialchange/PlaceCal/blob/ecdc157e895df47435858e0d86507efaa77c78dd/app/javascript/src/behaviors/behaviors.partner.js#L24) and below
- Looks to be left over from `behaviors.place.js`
- part of #1389

## Implementation plan

complete #1421 and then delete file.

Issue #1420, opened 2022-08-01
Remove `behaviors.collection.js`


aaaaargZombies:
## User story

As a developer I need to not have to read old code we aren't using, so that I can reduce the amount of time fixing bugs or adding features.

## Acceptance criteria

- [ ] file is removed

## Implementation notes & questions

This only contains code which relates to select2, because this aspect of placecal is currently abandoned it was not refactored with the rest. Leaving it in creates confusion around select2s implementation.

This is part of #1389

## Implementation plan

Delete file, add a comment in the relevant template warning about this change and the new select2 implementation in case work starts again in this area.

Issue #1419, opened 2022-08-01
Remove `behaviors.place.js`


aaaaargZombies:
## User story

As a developer I need to not have to look at old code we aren't using so that I can reduce the amount of time fixing bugs or adding features.

## Acceptance criteria

- [ ] file is removed but no site functionality is lost.

## Implementation notes & questions

I believe this has been replaced by `behaviors.partners.js`

This is part of #1389

## Implementation plan

Delete file, run tests, manually check site.

kimadactyl:
Place and Partner used to be diff models then we merged them - this is prob just old crud

PR #1431, opened 2022-08-09
Remove unused behaviours JS and asociated DOM triggers

aaaaargZombies:
This should address the following issues which all relate to unused code in the project.
- #1429
- #1419 
- #1420 
- #1422


ivankocienski (approved):
As this is just removing obsolete code this seems fine to me.

kimadactyl:
Hard to confirm this hasn't removed something we don't know about but seems fine from a click around

Issue #1427, opened 2022-08-02
[Bug]: partners showing up multiple times on same page


aaaaargZombies:
## Description

I was visiting https://gmsc.placecal.org/partners and I saw that Flourish was listed 3 times

## Steps to reproduce

1. visit https://gmsc.placecal.org/partners

## What you expected to happen

No duplication of listings

## Implementation notes

This is proving to be a sticky issue, the following prob need doing

- [ ] Rewrite the AR stuff as straight SQL so it can be read easily 
- [ ] Write better tests for site partners page that do things like check tags work
- [ ] Write tests for the blocks on the site homepages like "where to get online"

kimadactyl:
Labelling urgent as I think this looks super janky and will cause Questions

kimadactyl:
Oops. put this in the wrong place. This is currenlty totally ignoring the tag filter - shit. My bad for approving this but this needs fixing asap

<img width="1407" alt="Screenshot 2022-08-22 at 13 00 19" src="https://user-images.githubusercontent.com/511734/185916431-a01a014a-104e-47df-b530-0f66466b8b2b.png">

PR #1432, opened 2022-08-15
Fix duplicate partners on partner index pages

ivan-kocienski-gfsc:
Implements #1427

### DIFF
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -114,19 +114,40 @@ class Partner < ApplicationRecord
   # @param site [Site] The site we want partners for.
   # @return [ActiveRecord::Relation<Partner>]
   scope :for_site, lambda { |site|
+    query = Partner
+    where_parts = []
+    where_args = []
+
     site_neighbourhood_ids = site.owned_neighbourhoods.map(&:id)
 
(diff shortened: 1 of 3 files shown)


aaaaargZombies:
Tests run on my machine, manual testing by visiting the effected page showed no sign of the issue. I will start having a read of the code with the docs open so I can get an idea of whats going on. But as far as I can tell LGTM

aaaaargZombies (approved):
:+1:

aaaaargZombies:
I'd be tempted to change the title from "Bug" to "Fix" but this is probably a matter of taste

kimadactyl:
For this repo the idea is that they're sentances that can go straight into release notes but I don't think this is written down anywhere yet!

kimadactyl:
Screenshot 2022-08-22 at 12 18 58

Confirmed working on staging - i reproduced the record and this seems to be because it had 3 service areas set it was showing up 3 times

kimadactyl:
Oh crap. It's not respecting the tag filter at all now. I assumed this was a staging situation but... it is not. My bad for breaking this but we need this fixed asap...

Screenshot 2022-08-22 at 13 00 19

PR #1434, opened 2022-08-16
Refactor opening times picker with stimulus controller

### LINKED ISSUE #1417: Rewrite opening times picker in Stimulus
## User story

As a developer, I don't want to have one component using Vue, in order that our code base is easier to understand and has less dependancies

## Acceptance criteria

- [x] Opening times picker just uses Stimulus instead
- [ ] Vue removed as dependancy
- [x] Systems tests exist for picker

## Implementation plan

aaaaargZombies:
Opening a draft with the suggestion of merging this part of the work in early as it's completed.

fixes  #1417
- [x] Opening times picker just uses Stimulus instead
- [x] Systems tests exist for picker

Currently doesn't fix
- [ ] Vue removed as dependency

The last part has a lot of sprawling bits that seem to reach beyond the opening times picker and might be better approached in it's own issue/PR


aaaaargZombies:
Also now would be a good time for UI feedback

kimadactyl:
Looking good! UI needs a little love obvi but great start. It currenlty looks like this for me - i think 'all day' checkbox last and get it all on one line but assume you're already on it.

The 'all day' thing does create some funky states but I think it's no big deal.

It should prob make it easier to add in the next day after selecting the current one too - maybe after entering a time it incremenets the day?

Screenshot 2022-08-17 at 10 36 50

app/views/admin/partners/_form.html.erb
```diff
@@ -111,33 +111,40 @@
...
-        <div id="opening-times" class="opening-times" data="<%= @partner.opening_times.presence || [] %>">
-
-          <div class="added-days">
-            <added-days :list="addedDays"></added-days>
+        <div data-controller="opening-times" data-opening-times-data-value='<%= @partner.opening_times %>'>
+          <ul class="list-group" data-opening-times-target="list">
+          </ul>
+
+          <div class="mt-3 input-group flex-row align-items-center justify-content-sm-between flex-wrap">
+            <label class="m-0" for="day">Day
```
ivan-kocienski-gfsc, line 119: i would rather see this using the rails helpers as that is more inline with the rest of the codebase. labels would be something like `f.label :day` and input would be `f.input :time` see [here](https://guides.rubyonrails.org/form_helpers.html#other-helpers-of-interest). and `f.submit`

app/javascript/controllers/opening_times_controller.js
```diff
@@ -0,0 +1,151 @@
...
+});
+
+const sortedOpeningHours = (openSpecArray) =>
+	[...openSpecArray]
+		.sort(
+			(a, b) =>
+				parseFloat(openingHoursObj(a).open.replace(":", ".")) -
+				parseFloat(openingHoursObj(b).open.replace(":", ".")),
+		)
+		.sort(
```
ivan-kocienski-gfsc, line 34: sorting the output twice is wasteful as only the last will have any output

app/javascript/controllers/opening_times_controller.js
```diff
@@ -0,0 +1,151 @@
...
+	day: openSpec.dayOfWeek.split("/").filter((str) => str.includes("day"))[0],
+	open: openSpec.opens.slice(0, 5),
+	close: openSpec.closes.slice(0, 5),
+});
+
+const sortedOpeningHours = (openSpecArray) =>
+	[...openSpecArray]
+		.sort(
+			(a, b) =>
+				parseFloat(openingHoursObj(a).open.replace(":", ".")) -
```
ivan-kocienski-gfsc, line 31: i would feel happier if all the `parseFloat(openingHoursObj(a).open.replace(":", "."))` was done once outside of the sort it you will have objects sent to the callback multiple times duplicating work

app/views/admin/partners/_form.html.erb
aaaaargZombies: OK this isn't a hill I want to die on so if it's important to you ill change it but my preference is to keep it as is. My main issue is that the format below doesn't look like the stimulus documentation and I find it easier to come back to things that are well documented.

```
<label class="m-0" for="open">Opening Time
    <input class="ml-1" id="open" type="time" name="" value="00:00" data-opening-times-target="open">
</label>
```
```
<%= f.label :open, "Opening Time", <%= f.time :open, class: "ml-1", value: "00:00", data: {opening_times_target: "open"} %> %>
```
I'll leave the final decision to you.

kimadactyl: yeah i kinda agree with alfie, rails helpers make sense if we are accessing rails models but we're not, this logic only exists for our lil json object, it implies something that's not really true

app/javascript/controllers/opening_times_controller.js
```diff
@@ -0,0 +1,151 @@
...
+const openingHoursEnglish = (openSpec) => {
+	const { day, open, close } = openingHoursObj(openSpec);
+	return open === "00:00" && close === "23:59"
+		? `${day} all day`
+		: `${day} from ${open} to ${close}`;
+};
+
+const removeTime = (openSpecArray, openSpec) =>
+	[...openSpecArray].filter(
+		(el) => JSON.stringify(el) !== JSON.stringify(openSpec),
```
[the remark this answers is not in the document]
aaaaargZombies: I've removed this function in favor of the lodash method. hopefully that resolves this.

app/javascript/controllers/opening_times_controller.js
aaaaargZombies: This now uses lodash `sortBy` though it still requires two passes. Let me know if this is acceptable.

app/javascript/controllers/opening_times_controller.js
aaaaargZombies: I'm not sure this is possible using the lodash method? would it involve mutating the array, sorting it, then mutating it back?

app/views/admin/partners/_form.html.erb
aaaaargZombies: OK so the markup has remained as is but has been moved into a partial with a comment that explains why it doesn't use the same syntax as the rest of the site. This was the agreed compromise at standup so hopefully that resolves this issue.

app/javascript/controllers/opening_times_controller.js
```diff
@@ -0,0 +1,144 @@
...
+});
+
+const sortedOpeningHours = (openSpecArray) =>
+	orderBy(openSpecArray, [
+		(el) => dayOrder.indexOf(openingHoursObj(el).day),
+		(el) => parseFloat(openingHoursObj(el).open.replace(":", ".")),
+	]);
+
+const nextDay = (day) => {
+	const index =
```
ivan-kocienski-gfsc, line 36: tip: `let nextIndex = (dayOrder.indexOf(day) + 1) % dayOrder.length;`

ivan-kocienski-gfsc (approved):
Looks okay to me

PR #1439, opened 2022-08-22
Fix site partner filter

ivan-kocienski-gfsc:
Fixes: 1427 (again)

If a site has tags then limit the search to partners that have either
addresses or service areas in that site AND they have at least one site
tag.


kimadactyl:
Will review tomorrow, gj. Given this is litearlly the core of our app maybe this should have proper docstrings and the like? Can we do a better job of making this code simple and easy to parse?

kimadactyl:
This is looking much better but I think this still needs a little work.

- [ ] There's a lot in the setup block that seems to not be reused like the list of tagged partners. If it's not used more than once it should prob just be in the one test for cognitive clarity what's going on and also to speed it up.
- [ ] I would like the tag tests to be more comprehensive and have the numbers like 5 be clearer what they are. Like we should be actively testing for records to get filtered out and from looking at the tests I can't quite read that this is what's happening
- [ ] "Computer access" could do with a test for this based on tag too

kimadactyl:
Optional 

- [ ] Move all validations to a concern/module
- [ ] Move all presenter logic to a concern/module (e.g. twitter link)

ivan-kocienski-gfsc:
I'd rather keep this PR tightly scoped on the core problem of the for_site scope and its tests as it is core PC functionality.

kimadactyl (approved):

PR #1449, opened 2022-09-06
Bundle foreman and run the bundled version

erbridge:


kimadactyl:
The docs say [not to bundle foreman](https://www.rubydoc.info/gems/foreman/0.87.2#installation), do you beg to differ? :)

erbridge:
I forget that foreman are like this. This change was brought to you by the magic attempt to detect foreman and install it if it's missing not working because I have it installed for different Ruby versions. I'll just do a manual installation.

kimadactyl:
I mean... we could just ignore them it is a bit of a pain and I admit I don't quite see the issue with putting it in the development group. But yeah. I'm sure there's Reasons.

erbridge:
Now you point it out I've been bitten by dependency conflicts in the past, so...

PR #1458, opened 2022-09-14
Chore: tests covering image upload problems

### LINKED ISSUE #1428: [Bug]: Site admin not accepting new header image
## Description

Trying to upload a new header image to the C2 PlaceCal site

## Steps to reproduce

1. Uploaded one image successfully (but it was the wrong one)
2. Followed the same process to try and upload a second image, but it would not replace it
3. Tried logging in and out again, tried with and without image credit, made no difference

Image to upload

For the image to actually upload

## Platform (if relevant)

macOS Big Sur, Firefox

ivan-kocienski-gfsc:
Closes #1428 

- verify the error box is present when problems exist
- verify the image field has feedback text

Pages covered:
- site: new, edit
- partner: new, edit
- article: new, edit
- user: new, edit, profile


app/views/admin/users/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render_component "error", object: current_user %>
+<%= render_component "error", object: f.object %>
```
erbridge, line 1: This tightly couples this partial to its caller. Can we pass this in as a local scope variable when we call it instead?

Also, I'm not sure what this change has to do with the rest of the PR. Why is this change needed?

app/controllers/admin/sites_controller.rb
```diff
@@ -76,7 +76,7 @@ def destroy
     private
 
     def set_site
-      @site = Site.friendly.find(params[:id])
+      @site ||= Site.friendly.find(params[:id])
```
erbridge, line 79: How does this change relate to the rest of this PR? Doesn't seem to be part of the changes here.

test/integration/admin/partner_integration_test.rb
```diff
@@ -115,6 +115,47 @@ class PartnerIntegrationTest < ActionDispatch::IntegrationTest
...
+    }
+
+    sign_in @admin
+    post admin_partners_path, params: { partner: new_partner_params }
+
+    assert_not response.redirect?
+
+    assert_select "h6", text: "1 error prohibited this Partner from being saved"
+    assert_select '#form-errors li', text: "Image You are not allowed to upload \"bmp\" files, allowed types: jpg, jpeg, gif, png"
+    assert_select 'form .partner_image .invalid-feedback', text: "Image You are not allowed to upload \"bmp\" files, allowed types: jpg, jpeg, gif, png"
```
erbridge, line 136: I haven't looked, but the inclusion of both `jpg` and `jpeg` here makes me wonder: are we just checking the file extension or are we actually looking at the contents of the file to work out what it is?

erbridge (commented):
There are a couple of changes in here that I'm not sure I understand why they're here (and therefore why they've changed). I suggest pulling them out into their own PRs instead.

The tests themselves look good to me.

Also. There is no such thing as a bad cat picture, how dare you?!

app/views/admin/users/_form.html.erb
ivan-kocienski-gfsc: Actually I think you may have it the other way around: using `current_user` in a partial gives the partial a dependency on state that is not passed to the partial by the caller. The problem was that (as a root user on an edit user page)  when you updated the values and pressed save the error box (this code) was pulling its errors from `current_user` and not `@user` (as is set up by the controller).

In this case the coupling is intentional: the `_form` partial renders the form and the `error` partial renders the errors at the top of a form page

app/controllers/admin/sites_controller.rb
ivan-kocienski-gfsc: The sites controller before_action logic is a little tricky. This line is from a method that is invoked before update and destroy action. It set the site (duh) but it also *reset* the site if it was called more than once.

In the update action if the save fails it renders the edit view with the feedback. The view has some values that need to be set up in `set_variables_for_sites_neighbourhoods_selection` (brilliant name) which in turn calls `set_site` and therefore any errors on the `@site` object will be lost.

The PR is about missing error feedback on admin pages and this line was responsible for resetting the site (as set by the user provided params).

test/integration/admin/partner_integration_test.rb
ivan-kocienski-gfsc: I don't know as this is all done with carrierwave before my time on the code. I think we need to go through all the uploaders for site, partner, article and user and consolidate the acceptable formats. (frankly i would only allow jpgs as they are smaller in size and pretty well known in none-tech circles but that is just me)

kimadactyl: pretty sure we are just looking at the extension yeah. and no we should allow pngs, gifs and wepbs at least :)

erbridge: If we care about size, we can post-process images into webp with jpg fallbacks and accept any format. But you know, later.

> pretty sure we are just looking at the extension

This error message could be better probably, but one for a future iteration.

app/controllers/admin/sites_controller.rb
erbridge: Ah. The commit message and PR description implied this was just about writing tests, so seeing non-test changes confused me. Thanks for explaining.

app/views/admin/users/_form.html.erb
erbridge: I guess I just don't like to see `f.object` without any declaration of `f`, but it's not a blocker, I guess. It now makes more sense to me why this is here now I understand that there are actually fixes in this PR, not just tests.

erbridge (approved):
Makes sense to me now.

app/views/admin/users/_form.html.erb
ivan-kocienski-gfsc: Ah.I see now. Yeah `f` (in this context) is defined [here](https://github.com/geeksforsocialchange/PlaceCal/blob/4af2a5447501e7f386ad8c3032cb73007efe103f/app/views/admin/users/new.html.erb#L9) and [here](https://github.com/geeksforsocialchange/PlaceCal/blob/4af2a5447501e7f386ad8c3032cb73007efe103f/app/views/admin/users/edit.html.erb#L11).

No it is not a good variable name (i think short letter names are okay in a small local scope like a loop but not as an argument). But the problem with PC is that the list of refactorable parts is literally endless and some of them may be dependent on this weird behaviour so i tend to take a light touch approach to fixing things as it can lead to scope creep

test/integration/admin/partner_integration_test.rb
ivan-kocienski-gfsc: I will make this into a chore ticket

app/controllers/admin/sites_controller.rb
ivan-kocienski-gfsc: I see your point. It was writing tests but when writing those tests I found bugs that needed to be fixed. literal tdd.

I will write a better merge description

PR #1459, opened 2022-09-14
Automate setup

erbridge:
- Clarify Postgresql via Docker is optional
- Use bin scripts for setup and updates

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,6 @@ GOOGLE_API_KEY=
 MAPBOX_TOKEN=
 
 # If you set up postgres via Docker as per the README, you should set these
-POSTGRES_HOST=localhost
-POSTGRES_USER=postgres
-PGPASSWORD=foobar
+# POSTGRES_HOST=localhost
+# POSTGRES_USER=postgres
+# PGPASSWORD=foobar
--- a/README.md
+++ b/README.md
@@ -38,11 +38,13 @@ To run PlaceCal locally you will need:
 
 With that said, here's what you need to get rolling.
 
-### Set up postgresql server (Via docker)
+### Set up Postgresql locally
 
-**Note: Skip this step if you're using a system-installed docker**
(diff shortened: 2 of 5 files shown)


kimadactyl:
Few quick notes/qs as I log off

- We don't use seed migration gem any more - our seeds need a total rethink how they work
- Does the js setup need to also set nvm version?
- Importing events can take a long time - wonder if this is too much to throw into the main setup script
- Will this update bundler and yarn every time it runs? Those messages do get annoying!
- How is ruby version set?
- We have a script for syncing the db from the main/staging server [private notion instructions here](https://www.notion.so/gfsc/PlaceCal-live-db-download-instructions-0cd4a7fb21314890a3e18d718a0b8f8b) - not sure where this should live

ivan-kocienski-gfsc (approved):
This looks okay to me.

erbridge:
@ivan-kocienski-gfsc I think there's value in tickets when there's extended work involved in something. I think for small tech debt improvements, creating a matching ticket is overkill and process for the sake of process. Like writing a ticket for this would have taken longer than the work to fix it.

Please do tell me if you think I'm wrong about that, though.

erbridge:
@kimadactyl All good questions, but remember: this is just about making the already documented stuff automatic instead of needing copy-paste. There are some assumptions built in here that you've done your environment set up correctly (using RVM or rbenv, have Postgresql running, etc), that making work in a platform agnostic way would be painful. I did consider printing a bunch of "do this first" stuff to surface those requirements, but decided against it at this point.

erbridge:
Re event importing, we should take a look at the best thing to do, but probably at another time. This PR is about automating the existing set up, not innovating on it. Also the event import was pretty quick for me.

erbridge:
@kimadactyl Could you write a ticket about the database sync stuff, please? I'm not sure I have enough context to at this point.

PR #1461, opened 2022-09-15
Fix: partner tag selector bug

### LINKED ISSUE #1460: [Bug]: Create Partner tag selection box
## Description

As an admin creating a new partner, I would like to use the fancy select2 tag selector.

## Steps to reproduce

1. Be logged in as admin
2. Navigate to 'Add new partner page'
3. Scroll to tag selection input

Be able to use fancy tag selector.

ivan-kocienski-gfsc:
Closes #1460 

On admin partner edit page, partners with missing opening times value would crash JS and cause tags selector to fail.

Quick sanitisation of value and a system test to cover that scenario

### DIFF
--- a/test/system/admin/partner_test.rb
+++ b/test/system/admin/partner_test.rb
@@ -91,4 +91,17 @@ class AdminPartnerTest < ApplicationSystemTestCase
       assert !data.include?(new_time)
     end
   end
+
+  test 'opening time picker on partner form survives missing value' do
+    @partner.update! opening_times: nil
+
+    click_sidebar 'partners'
+    await_datatables
+    click_link @partner.name
+    await_select2
+
+    # very specific bug here: if opening times has malformed data
+    #   it will crash the partner tags selector
+    assert_selector '.partner_tags ul.select2-selection__rendered'
+  end
 end
--- a/test/system/admin/partner_test.rb
+++ b/test/system/admin/partner_test.rb
@@ -100,8 +100,10 @@ class AdminPartnerTest < ApplicationSystemTestCase
     click_link @partner.name
     await_select2
 
-    # very specific bug here: if opening times has malformed data
-    #   it will crash the partner tags selector
+    # very specific bug in the view template here: if opening times has malformed data
+    #   it will cause problems for the javascript that runs the partner tags selector
+    #   (in the browser). so we verify the select2 code has worked by seeing if it has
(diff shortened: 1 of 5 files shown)


erbridge (approved):
Just to check I've understood this change:

Before: If a partner had no opening times, this would fall over because `data-opening-times-data-value='<%= opening_times_payload %>'` would evaluate to `data-opening-times-data-value='nil'` or similar.

After: If there are no opening times we default to an empty list, so things don't fall over.

This feels like something that should be the model's responsibility to do. Like the opening times should default to an empty list everywhere, instead of patching it in the view. Did you try this and hit yak shaving territory? If not, could you check if doing that still results in a green test suite instead of implementing a special case here? If it's not simple to do that, this is ok, though.

test/system/admin/partner_test.rb
```diff
@@ -91,4 +91,17 @@ class AdminPartnerTest < ApplicationSystemTestCase
...
+  test 'opening time picker on partner form survives missing value' do
+    @partner.update! opening_times: nil
+
+    click_sidebar 'partners'
+    await_datatables
+    click_link @partner.name
+    await_select2
+
+    # very specific bug here: if opening times has malformed data
+    #   it will crash the partner tags selector
```
erbridge, line 104: Is this referring to a bug that exists in the test, a bug we're testing for, a workaround, what?

ivan-kocienski-gfsc: i expanded the comment

ivan-kocienski-gfsc:
I have moved the json sanitisation into the model, which is a much better place for it

erbridge (approved):
Much nicer! You've disabled parallelization, which might have been an accident? Otherwise, looks good.

test/test_helper.rb
```diff
@@ -32,7 +32,7 @@ module ActiveSupport
   class TestCase
     include FactoryBot::Syntax::Methods
 
-    parallelize(workers: :number_of_processors)
+    # parallelize(workers: :number_of_processors)
```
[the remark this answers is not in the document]
ivan-kocienski-gfsc: i did because when i ran the systems test it spun up 4x servers and just killed my computer (i have a lot of stuff open to do all my projects). this then led the tests to have a 50% failure rate because they kept timing out (the socket).

PR #1464, opened 2022-09-20
Config: setting site domain on deploy

ivan-kocienski-gfsc:
When github action deploys to staging or production set the domain name environment variable so the application can generate proper URLs


erbridge (commented):
Does `dokku/github-action`'s environment get rolled in to the deployed application? Feels unlikely to me (it would be a massive security risk and reproducibility issue) and I couldn't see anything in the docs.

PR #1466, opened 2022-09-21
Fix of environment configurations

ivan-kocienski-gfsc:
(when configuring mailer in staging and production)


erbridge (approved):
Aha. Makes sense.

Issue #1462, opened 2022-09-19
[Bug]: Graphql query on Partner with logo field crashes


ivan-kocienski-gfsc:
## Description

try to look up a partner using the GraphQL API requesting the `logo` field and PlaceCal 500s

## Steps to reproduce

1. Wrote small script to read graphql endpoint (this could be a gist or an attachment)
2. Run query, print result

## What you expected to happen

See partner's logo field populated.

## Code used

```ruby
require 'graphql/client'
require 'graphql/client/http'

ENDPOINT = 'https://admin.placecal-staging.org/api/v1/graphql'
# ENDPOINT = 'http://admin.lvh.me:3030/api/v1/graphql'

module PlaceCal

  HTTP = GraphQL::Client::HTTP.new(ENDPOINT)

  Schema = GraphQL::Client.load_schema(HTTP)

  Client = GraphQL::Client.new(schema: Schema, execute: HTTP)

  PartnerQuery = Client.parse <<-GRAPHQL
    query($id: ID!) {
      partner(id: $id) {
        id
        name
        description
        summary
        contact { email, telephone }
        url
        address { streetAddress, postalCode, addressRegion, geo { latitude, longitude } }
        areasServed { name abbreviatedName }
        logo
      } }
  GRAPHQL
end

def lookup_partner(id)
  result = PlaceCal::Client.query(PlaceCal::PartnerQuery, variables: { id: id })
  puts JSON.pretty_generate(result.to_h)
end

lookup_partner 153
```

## What should happen

```json
{
  "partner": {
    "id": "153",
    "name": "Name",
    "description": "",
    "summary": "",
    "contact": {
      "email": "info@example.com",
      "telephone": "0808 123 4567"
    },
    "url": "",
    "address": null,
    "areasServed": [
      {
        "name": "England",
        "abbreviatedName": "England"
      }
    ],
    "logo": "http://localhost:3030/uploads/partner/image/153/reddit-grumpy-cat-2.jpg"
  }
}
```

## What happens

```json
{
  "errors": [
    {
      "message": "500 Internal Server Error"
    }
  ]
}

```

erbridge:
This still seems to give a 500 on staging for me. Needs further investigation.

ivan-kocienski-gfsc:
I am going to look into this for all image URLs  (and to verify that the dev, staging and prod servers are configured to generate full URLs properly

erbridge:
[...]

PR #1465, opened 2022-09-20
Fix: set up domain and build partner logo url

ivan-kocienski-gfsc:
Fixes #1462

PR #1469, opened 2022-09-27
Test graphql image urls

ivan-kocienski-gfsc:
Verifies #1462

For Trans Dimension we want to verify that the URLs we get back from PlaceCal for news and partners have the full protocol://domain:port part


app/graphql/types/article_type.rb
```diff
@@ -29,8 +29,7 @@ class ArticleType < Types::BaseObject
           method: :body
 
     field :image, String,
-          description: 'The image of the article (a URL)',
-          method: :highres_image
+          description: 'The image of the article (a URL)' #,
```
erbridge, line 32:
```suggestion
          description: 'The image of the article (a URL)'
```

erbridge (commented):
Do you think we could give PRs that fix things names and descriptions that reflect those fixes, please? I was expecting this PR to only contain some tests without any behaviour changes, and I'm not sure I follow why the behaviour has changed in the way it has to be able to review it. What was it doing before, why was that bad, and what is it doing now?

ivan-kocienski-gfsc:
@erbridge The situation is this: i am working on TD tickets that relied on functionality that was broken in PC. I implemented the functionality (for the specific issue of partner logos) in #1465 which itself involved setting up PC environments. This was done and the work pushed and merged.

Whilst it worked for the TD problem I was working on I couldn't figure out a way to test the functionality automatically (like as an external API user). After a weekend I thought of a way to test the GQL endpoint like an end-user would by (ab)using the system tests. This was not really the thing I was supposed to be working on (TD tickets) but I thought putting in tests was A Good Idea so I continued. 

The next TD ticket I got was putting images on news items. So I applied the same fix from earlier (on PC) and had gotten the system tests up to be useful. It's not a good idea at all to try to sneak extra work in but as this was the same functionality as in 1465 and this PR was already covering tests I felt that it wasn't an egregious creep of scope.

erbridge:
@ivan-kocienski-gfsc Failing test.

erbridge (approved):
Approving with the caveat that I'm not sure I understand why it wasn't working before, and that makes me nervous.

PR #1474, opened 2022-10-05
Use HTTPS for system tests

erbridge:
We run everything else assuming we're using HTTPS. This might be the cause of the flaky test failures.


erbridge:
This doesn't do what we wanted. Replaced by #1485.

PR #1475, opened 2022-10-05
Generate code coverage reports on test

erbridge:
We never want a reduction in coverage, only an increase.


kimadactyl:
The reason we took this out is that sometimes this was getting in the way as not everything you add to a code base always has coverage (e.g. factories, seeds, readmes etc). We were using code climate and it was just really overbearing. Happy to put back in just documenting why it's commented out!

erbridge:
I think some sort of intentional "I'm going to drop the coverage and that's ok" is the aim here. Rather than the accidental "oops, what are tests?" thing that happens.

ivan-kocienski-gfsc:
I am not sure that this should be merged too quickly as it will hinder our ability to push needed fixes to production.

But it is absolutely a good idea that I agree should be part of the workflow going forward.

kimadactyl:
Is there a way we can do this without having something as dramatic as having the CI just fail? I'm just very nervous about introducting something that I think will only increase stress based on an arbritrary measurement.

kimadactyl:
Closing for now to tidy up a bit, can remake later if we decide we need it

Issue #1463, opened 2022-09-19
Remove seed migrations


erbridge:
## User story

As a developer I need to not integrate with the unused seed migrations gem so that I can avoid distractions while developing.

## Acceptance criteria

- [ ] The seed migrations gem is no longer installed
- [ ] There are no remaining references to seed migrations in the code

## Implementation plan

_To be written by the developer_

kimadactyl:
Related to #1371

Issue #1371, opened 2022-07-05
[Bug]: README quickstart instructions incomplete - seeding the DB


aaaaargZombies:
## Description

I followed the "quick start" instructions to get the database up and running. It looked positive in the command line but when connecting to 
`lvh.me:3000` the page would not load.

@kim suggested running 

```sh
bundle exec rails db:setup db:migrate seed:migrate
bundle exec rails import:all_events
```
This resolved the issue and these commands are covered in the notion page "PlaceCal live db download instructions". But I am not clear which `db` commands where essential to getting up and running.

## Steps to reproduce

Follow quick start guide in readme

## What you expected to happen

To Follow quick start guide in readme and have a working DB

erbridge:
This is fixed by the introduction of setup scripts in #1459, I think. Can you confirm @aaaaargZombies?

erbridge:
Oh wait. I do too now. Odd. Thanks for testing.

kimadactyl:
I think you prob found this now (sorry looking at some old todos) but the live db sync instructions are on our [private notion](https://www.notion.so/gfsc/PlaceCal-live-db-download-instructions-0cd4a7fb21314890a3e18d718a0b8f8b). the prob is that we only ever develop with live data so the seeds are now out of date. this is covered in the new issues ofc.

PR #1480, opened 2022-10-06
Remove seed migrations gem and start new seeding system

erbridge:
Start of work against: 

- https://github.com/geeksforsocialchange/PlaceCal/issues/1463
- https://github.com/geeksforsocialchange/PlaceCal/issues/1371


kimadactyl:
This is maybe a bit in the weeds but it would be very nice to have a map of Normal Island you can refer to when writing tests - sounds silly but I think having a concrete visualisation will help people no end

erbridge:
This was mainly so I could actually run PlaceCal without having a production data dump, which I realised was why I was stuck in a redirection loop.

ivan-kocienski-gfsc:
Looks like [this](https://github.com/geeksforsocialchange/PlaceCal/blob/main/db/migrate/20211203153422_correct_datamodel_by_importing_neighbourhoods_json.rb) is where Neighbourhoods get loaded in

kimadactyl:
> Looks like [this](https://github.com/geeksforsocialchange/PlaceCal/blob/main/db/migrate/20211203153422_correct_datamodel_by_importing_neighbourhoods_json.rb) is where Neighbourhoods get loaded in

Ok but this ticket is explicitly **not** addressing neighbourhoods, it's just getting a basic seed system up and running

ivan-kocienski-gfsc:
Hmm perhaps I am misunderstanding terminology here: to me seeding a database is where you have nothing in the db and load all the sample data in that a dev will typically need day to day. 

So we assume the migration populates neighbourhood and then never touch that table again?  Okay.

Perhaps the original issue could describe what needs to be seeded like:

- Create Users, Site, Tags, Partners
- Link User to Site and different User to Partner
- Maybe make Online Partner, Address Partner and Service Area Partner (and Tag Partner?)
- Create Calendar
- Create some events linking Partner to Site. ( <-- could be tricky)

kimadactyl:
I don't think you're misunderstanding, you're just trying to fix everything in one go.

The goal here is to just improve it on how it is now - i.e. take out the seed migrations gem and make seeding a user actually work.

We don't have to eat the whole elephant in one bite! Each of those other steps is correct but is an additional ticket.

ivan-kocienski-gfsc:
Just so you know if you run `bin/setup` it will drop the database and bring it back up from schema (and then run the seeds). This means there is an empty database with no neighbourhoods.

kimadactyl:
Ahh that's a big gotcha and we should add a warning for that.

I don't think this needs any extra features tho! We can just merge it when the tests pass. We should be adding the seed files a bit more conciously IMO.

kimadactyl:
Closed but remade in other PR to avoid annoying git issues

Issue #867, opened 2021-12-21
Set up view_component


kimadactyl:
mountain_view is showing its age and there's now much better things to do component isolation with. We currently use a mix of mountain_view and vanilla Rails partials. This has worked OK to date but is proving a bit of an app bottleneck as its hard to isolate and test components, and preview them properly. Also, it has a lot of confusing issues with passing javascript back and forward.

View_component seems to be what we need!

## Acceptance criteria

- [x] Set up view_component gem
- [x] Migrate one simple component from mountain_view (or just make a "hello world" component)

kimadactyl:
Inspiration: https://components.publishing.service.gov.uk/component-guide

r-ferrier:
I think we should descope setting up lookbook completely. Here are my reasons!

We added a very similar system (storybook) to a react project I worked on for a similar sized app and setup to ours now. 

- You have to add extra code for each component to get it picked up by the preview, and organise how different variants & how many different variants are represented, which can significantly slow down development. 
- It adds an extra step to the CI pipeline, as you now have to approve design changes in a vacuum on storybook, but then also in context in staging
- As a developer I found that if I used it in a tdd kind of way, designing and previewing within the storybook ecosystem, I would frequently run into problems where I had overlooked different contexts in which the design needed to change or did not look quite right once viewed within the app as a whole - in order for it to be effective, you have to invest a lot of time into making sure every different possible situation is represented in both setups

I do think this kind of system is very useful if you're regularly communicating small design changes across multiple projects with clients or designers who you can't share your staging or local environments with. But I think in our case the client-dev relationship is much closer, and the separation of concerns is much less clear. 

TLDR: it's an extra system to maintain and extra steps to take every time you want to add a new component & make a PR. I don't think it would add any value to our processes at this point in time.

kimadactyl:
OK all good!

katjam:
Verified on staging:

![image](https://github.com/geeksforsocialchange/PlaceCal/assets/861172/b5491b31-d2bf-41af-bbb1-d35815386840)
![image](https://github.com/geeksforsocialchange/PlaceCal/assets/861172/54580eab-9d89-4a78-97a3-7050c3abe4ba)

PR #1481, opened 2022-10-07
WIP View component

erbridge:
Might be related to issue #867, removing Mountain View (?)

Looks like it is setting up [View Component](https://viewcomponent.org/).


kimadactyl:
This looks very nice and i think something @ivan-kocienski-gfsc was asking about regarding tests - like it would be nice to have each piece easily testable

erbridge:
A requirement, imo, so yes.

kimadactyl:
Adding to current milestone for @katjam to look at so this doesnt get forgotton

PR #1488, opened 2022-10-18
Use uk_postcode gem for postcode validation

### LINKED ISSUE #1483: [Bug]: LxZ events could not be added
## Description

When trying to add the following information, the form silently errors and will not save:

```
Name: LxZ Events
Summary: For women, GNC, trans & NB people who party like BALLERS
Street address: Beduin Smithfields
Street address 2: Charterhouse St
Street address 3: Smithfields
Postcode: EC1M 6HA
Website address: https://www.instagram.com/lxzevents/?hl=en
Facebook link: LxZeventsLDN/
```

On poking it it seems to be an issue with the postcode - if I change this then the entry saves.

## Steps to reproduce

1. Go to partners/new
1. Add information as above
1. Press 'save'
1. Form doesn't save

The event to save properly

## What would you like to happen instead?

- [ ] A validation error appears if the postcode is invalid
- [ ] In this case the postcode is not invalid so I expect the partner to save 
- [ ] This partner is added to PlaceCal when the bug is fixed

erbridge:
Rather than maintaining a giant regular expression, let's lean on prior art, maintained by someone else.

Fixes #1483

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,6 +88,7 @@ gem 'oj'
 gem 'paper_trail'
 gem 'rollbar'
 gem 'sendgrid-actionmailer'
+gem 'uk_postcode'
 gem 'virtus'
 gem 'whenever', require: false
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -504,6 +504,7 @@ GEM
     turbolinks-source (5.2.0)
     tzinfo (2.0.5)
       concurrent-ruby (~> 1.0)
+    uk_postcode (2.1.8)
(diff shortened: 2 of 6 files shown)


kimadactyl:
When this is deployed to prod we need to let Georgie know - its blocking TransDim

kimadactyl:
Can we add a test to this to make sure our problematic postcode passes along with some other ones from our database?

kimadactyl (approved):

kimadactyl:
Woo looks great!

PR #1494, opened 2022-10-20
Fix: Don't try to hash non string inputs

### LINKED ISSUE #1493: Bug: TypeError: no implicit conversion of HTTParty::Response into String
https://rollbar.com/gfsc/placecal-production/items/223/

ivan-kocienski-gfsc:
Fixes #1493 

There was a bug where non-string responses were being passed into string expecting methods (and crashing). Turns out the method in question was already defined on a base class in a more robust manner. So let's use that!


aaaaargZombies (approved):
:+1:

PR #1495, opened 2022-10-20
Fix: Address lookup failure recovery

### LINKED ISSUE #1484: [Bug]: Calendar ID:99 not being imported successfully
## Description

[GI London Youth Work calendar](https://admin.placecal.org/calendars/99/edit) is aborting on import with "Error: Internal database error for calendar GI London Youth Work calendar (id 99): Couldn't find Neighbourhood"

You could clone the URL and run it on your local machine first and if that doesn't work then move onto...

You may want to hop onto the server rails console and see which neighborhood is causing the problem and work back from their. Is it their data or our inter

ivan-kocienski-gfsc:
Closes #1484 

If an address is created that does not map to a neighbourhood in our database, just return nil to the callee.

In practice this is mainly used by the importer which does not deal well with ActiveRecord exceptions. This is by design as any AR problems should not be happening in the importer!

So now any failed lookups (by the importer) will cause it to mark the calendar as failed and move on.


ivan-kocienski-gfsc:
DJ MUMSNET / Georgie 

"Morning all. Getting this error for this calendar still - does anyone know how to fix this?"

kimadactyl (approved):

Issue #1407, opened 2022-07-26
[Bug]: #107 ActiveRecord::RecordNotFound: Couldn't find Event with 'id'=275583


kimadactyl:
## Description

https://rollbar.com/gfsc/placecal-production/items/107/

This is single handedly overloading our rollbar limit, labeled urgent but realistically we might be over limit before this gets tackled

## Steps to reproduce

1. What you did first
1. What you did next

## What you expected to happen

What would you like to happen instead?

## Platform (if relevant)

What device and browser were you using?

ivankocienski:
Might want to look at [this](https://riptutorial.com/ruby-on-rails/example/9621/rescuing-activerecord--recordnotfound-with-redirect-to) and check if PC has something along those lines.

Issue #1408, opened 2022-07-26
[Bug]: Missing assets causing rollbar pileup


kimadactyl:
## Description

These causing a huge rollbar pileup, marking urgent but might be too late by the time we get around to it

https://rollbar.com/gfsc/placecal-production/items/106/
https://rollbar.com/gfsc/placecal-production/items/105/

## Steps to reproduce

1. What you did first
1. What you did next

## What you expected to happen

What would you like to happen instead?

## Platform (if relevant)

What device and browser were you using?

kimadactyl:
Think this is resolved now right @ivan-kocienski-gfsc?

Issue #759, opened 2021-10-25
Add custom error pages


katjam:
## Acceptance

- [ ] As a user of the site, I get a branded error screen with helpful information if I encounter an HTTP error (e.g. 404. 50x)

## Notes

- Invalid subdomain

katjam:
This is also important for when subdomains get taken down e.g. https://mossley.placecal.org/

PR #1496, opened 2022-10-20
Feature: Not Found page

ivan-kocienski-gfsc:
Fixes: #1407 

When an ActiveRecord resource can't be located, show a user friendly page.


kimadactyl:
Will this also fix #1408?

Looks good to me tho! Can't believe this is was issue all along, lol.

ivan-kocienski-gfsc:
Unfortunately not. Assets are served outside of the app so only missing resources will get covered by this fix

aaaaargZombies (approved):
OK I'm approving because it delivers the not found page nicely when we don't have the resource for the matching ID. I assume this is good enough for roll bar but I can't see the actual errors it's producing.

kimadactyl:
Confirmed working but only for events - will open new tickets for other errors

Screenshot 2022-10-31 at 11 46 07
Screenshot 2022-10-31 at 11 46 01

kimadactyl:
Ah its here already: https://github.com/geeksforsocialchange/PlaceCal/issues/759

PR #1501, opened 2022-10-24
Fix: better calendar feedback when import fails

### LINKED ISSUE #1271: [Bug]: Calendar Importer doesn't deal well with API problems
## Description

When the calendar importer runs it assumes the API call was valid and successful. This may not be the case and the Parser may be raising all kinds of exception indicating a problem state. Some parsers catch these (see ICS and facebook Parser classes) and return an empty array to the importer.

## Steps to reproduce

1. Using a production DB snapshot on localhost (from 7th June, 2022)
2. Comment out the lines near the end of `app/jobs/calendar_importer/parsers/ics.rb` where there is a `rescue StandardError => e` to the end of the `parse_remote_calendars` method
3. Run `rails import:events_from_source[102]`
4. An exception should occur.

The following output is shown
```
Importing events for calendar The Outside Project (ID 102) for place The Outside Project (ID 137) is forced
>>>>>>
RuntimeError
Invalid iCalendar input line: */
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/icalendar-2.7.1/lib/icalendar/parser.rb:176:in `parse_fields'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/icalendar-2.7.1/lib/icalendar/parser.rb:159:in `next_fields'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/icalendar-2.7.1/lib/icalendar/parser.rb:42:in `parse'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/icalendar-2.7.1/lib/icalendar/component.rb:110:in `_parse'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/icalendar-2.7.1/lib/icalendar/component.rb:14:in `parse'
/home/ivan/work/clients/gfsc/PlaceCal/app/jobs/calendar_importer/parsers/ics.rb:43:in `import_events_from'
/home/ivan/work/clients/gfsc/PlaceCal/app/jobs/calendar_importer/parsers/base.rb:37:in `calendar_to_events'
/home/ivan/work/clients/gfsc/PlaceCal/app/jobs/calendar_importer/calendar_importer_task.rb:24:in `run'
/home/ivan/work/clients/gfsc/PlaceCal/app/jobs/calendar_importer_job.rb:40:in `perform'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/activejob-6.1.5.1/lib/active_job/execution.rb:48:in `block in perform_now'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/activesupport-6.1.5.1/lib/active_support/callbacks.rb:117:in `block in run_callbacks'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/i18n-1.10.0/lib/i18n.rb:323:in `with_locale'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/activejob-6.1.5.1/lib/active_job/translation.rb:9:in `block (2 levels) in <module:Translation>'
/home/ivan/.rbenv/versions/2.7.6/gemsets/gfsc-placecal/gems/activesupport-6.1.5.1/lib/active_support/callbacks.rb:126:in `instance_exec'
```

If a problem occurs in the parser (remote API fetch/parse) then the parser should flag an error state. When the importer task gets back its parser it should check if a problem occurred and save it into the calendar so the user can see it on their UI (and possibly fix it).

ivan-kocienski-gfsc:
Fixes #1271 

For ICS calendars, when an URL fetch returns invalid data log this problem on the calendars state machine so users can see the feedback (rather than be silent about it).


aaaaargZombies:
This will only show a warning after it has attempted to import, and then only on the calendar page. as a user I had to;

- create a calendar with the bad url
- add it to the import queue
- return to the calendar page to see the warning message

Is there anyway to make feedback quicker? Or should we include this validation step in the handbook? Perhaps this is out of scope?

Though this clearly an improvement for existing calendars that stop working.

kimadactyl:
I just went through this with Georgie actually and yes our "time to see result of your import" is far too long. I think it is deffo a new ticket tho!

aaaaargZombies:
yes that sounds sensible rather than overloading issues which is frustrating.

aaaaargZombies (approved):
works nicely

ivan-kocienski-gfsc:
The problem is that you can't tell if a calendar will import successfully without importing the calendar. The best you can hope for is to put together some heuristics that try to make a best guess. You may suggest that we could do the import in request immediately upon calendar save but they can take a very long time which is why they are background jobs.

So this is not an easy problem to solve.

PR #1506, opened 2022-10-26
Improve issue templates

erbridge:
This sets the "needs triage" label on all new issues to support that as a new workflow.

It also created a new technical debt template to reduce the amount of skipping outside of the templates being done.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,9 @@
 name: Bug
 about: Create a bug report
 title: "[Bug]: "
-labels: bug
+labels:
+  - bug
+  - needs triage
 ---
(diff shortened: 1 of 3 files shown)


kimadactyl:
Is `tech debt` and `chore` the same thing? Do we need both?

erbridge:
No, they're not the same thing imo, but I would happily burn "chore" as a label because nobody uses it to mean the same thing.

kimadactyl:
Yeah, `chore` is used by release drafter to mean "goes in the maintance section" - will approve this and leave it to you if you want to update the release drafter config to use `tech debt` instead as part of this PR

kimadactyl (approved):

kimadactyl:
Oh right it's on automerge, lol. Np either way :)

PR #1508, opened 2022-10-26
Improve the GitHub Actions

erbridge:
These are faster, we stop using whatever state the head of the action is, we use the same Node version we use in development, we have separate workflows for deployments (still with the guards for now to prevent accidental deployment to the wrong place, though branch selection should already avoid that).

The commit messages have more details.


kimadactyl (approved):
All looks very sensible to me if a bit out of my depth!

PR #1513, opened 2022-10-27
Fix deployment actions

erbridge:
I noticed that the new workflows weren't running. I think this is due to the test workflow being "called" `.github/workflow/test.yml` (or something) and not just `test`. So we give the workflow an explicit name here instead, so we can reference that.

While I was looking at that I realized that we needed to also only run on success, so I've implemented that, too.


kimadactyl (approved):
Are you sure they wern't running? Happy for this anyway

Issue #1522, opened 2022-11-01
[Bug]: Production isn't deploying


kimadactyl:
## Description

Merged main into prod and deploy is being skipped.

Logged as GFSC incident 4.

Assume this is similar issue to the other one but I can't see the diff between the configs! Weird.

PR #1523, opened 2022-11-02
Hopefully fix deployment workflows

erbridge:
Relates to #1522

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -11,10 +11,12 @@ on:
 
 jobs:
   deploy:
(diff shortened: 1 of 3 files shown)


ivan-kocienski-gfsc (approved):
Looks good to me

kimadactyl (approved):
A bit nervous about changing staging deploys now it's working but trusting the process!

erbridge:
If we change staging to match, we give ourselves more confidence in production. My theory here (which I should have written in the commits) is that the workflow runs "from" the default branch, so doesn't meet the condition to run from `production`, despite being triggered by a workflow run for the `production` branch.

PR #1524, opened 2022-11-02
Fetch the full Git history for deployments

erbridge:
dokku/github-action pushes a specific commit and needs to have a local copy of that commit to do so.

Relates to #1522


ivan-kocienski-gfsc (approved):
Looks good to me

PR #1526, opened 2022-11-02
Recombine deployment and test workflows

erbridge:
dokku/github-action makes some faulty assumptions about what to deploy and it's not worth the time to work out how to work around them.

Relates to #1522


ivan-kocienski-gfsc (approved):
Looks good to me

PR #1534, opened 2022-11-07
Fix: cleaning up HTML before parsing into markdown

### LINKED ISSUE #1530: [Bug]: Metadata description for events is Markdown when it should be plain text
https://trans-dimension.placecal.org/events/320734:

Screenshot 2022-11-02 at 18 48 09

I think that says it all?

ivan-kocienski-gfsc:
Fixes #1530 

Strip HTML tags of their attributes (but keep <a> tag hrefs)

### DIFF
--- a/app/jobs/calendar_importer/events/base.rb
+++ b/app/jobs/calendar_importer/events/base.rb
@@ -52,10 +52,13 @@ module CalendarImporter::Events
         # if we get HTML then remove all the attributes from all of
         # the tags so it doesn't interfere with the kramdown step
         doc.css('*').each do |tag|
+          # rubocop:disable Style/HashEachMethods
           tag.keys.each do |attribute_name|
             next if tag.name == 'a' && attribute_name == 'href'
+
             tag.remove_attribute attribute_name
           end
+          # rubocop:enable Style/HashEachMethods
         end
 
         if footer.present?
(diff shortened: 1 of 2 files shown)


app/jobs/calendar_importer/events/base.rb
```diff
@@ -52,10 +52,13 @@ def html_sanitize(input)
         # if we get HTML then remove all the attributes from all of
         # the tags so it doesn't interfere with the kramdown step
         doc.css('*').each do |tag|
+          # rubocop:disable Style/HashEachMethods
```
kimadactyl, line 55: why not just to what rubocop asks instead? :)

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashEachMethods

kimadactyl:
This looks good but prob as good a time as any to add at least a minimal test for this?

app/jobs/calendar_importer/events/base.rb
ivan-kocienski-gfsc: Because it is assuming `tag` is an enumerable hash, which it is not, so invoking `each_key` raises a `NoMethodError`. It's just guessing based on text parsing and heuristics, which in this case is wrong

kimadactyl: fair, will leave this to F to sign off

kimadactyl (approved):
Looking good!

PR #1537, opened 2022-11-08
Strip all HTML out of summary fields

### LINKED ISSUE #1533: [Bug]: Literal &amp;s are in the API output for event titles
Not sure if this is a problem PlaceCal or TransDim end but we have a HTML token here visible in the event title:

https://transdimension.uk/events/320735

Screenshot 2022-11-03 at 12 54 44

@katjam, what do you think?

ivan-kocienski-gfsc:
Closes #1533 

Assume summary field could contain risky content (like HTML) and scrub that out before saving in DB.

### DIFF
--- a/app/jobs/calendar_importer/events/base.rb
+++ b/app/jobs/calendar_importer/events/base.rb
@@ -27,7 +27,7 @@ module CalendarImporter::Events
     # Convert h1 and h2 to h3
     # Strip out all shady tags
     # Convert all html to markdown
-    def html_sanitize(input)
+    def html_sanitize(input, just_as_text: false)
       input = input.to_s.strip
       return '' if input.blank?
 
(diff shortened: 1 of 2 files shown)


app/jobs/calendar_importer/events/base.rb
```diff
@@ -27,7 +27,7 @@ def sanitize_invalid_char(input)
     # Convert h1 and h2 to h3
     # Strip out all shady tags
     # Convert all html to markdown
-    def html_sanitize(input)
+    def html_sanitize(input, just_as_text: false)
```
erbridge, line 30: Can we give this argument a more descriptive name, please? For example:

```suggestion
    def html_sanitize(input, convert_to_plaintext: false)
```

erbridge: Thought actually, thinking about it, why would we ever want the old behaviour? Can we just replace the old behaviour with the much simpler and more reliable new behaviour?

ivan-kocienski-gfsc: Some people may have HTML in their text fields (via a JSON or XML export API) and we need to preserve that document structure. The summary field should be basic plaintext but you can never rely on user provided data to be sane. This switch allows the old behaviour to do a full clean but also allows us to "just" get plaintext

ivan-kocienski-gfsc: fair enough

erbridge (approved):

PR #1542, opened 2022-11-09
Speeding up partner index page load time

### LINKED ISSUE #1532: [Bug]: TransDim partners view on PlaceCal end taking 10s+ to load
This page here is currently taking upwards of 10 seconds to load:

https://trans-dimension.placecal.org/partners

This isn't the biggest issue but it is making validation any discrepancies between PlaceCal and TransDim more of a chore than it should be

ivan-kocienski-gfsc:
Fixes #1532 

When pulling neighbourhoods out of the database only instantiate the fields that are required so ActiveRecord doesn't have to parse unused values


erbridge (approved):
Looks sensible to me.

PR #1573, opened 2022-11-14
Extend capybara timeout

### LINKED ISSUE #1538: Fix flaky test `CreateAdminUsersTest#test_visiting_the_index`
## Description

Fix the flaky test that sometimes causes CI to fail:

```
Error:
CreateAdminUsersTest#test_visiting_the_index:
Capybara::ElementNotFound: Unable to find field "First name" that is not disabled
    test/system/create_admin_users_test.rb:35:in `block in <class:CreateAdminUsersTest>'
```

ivan-kocienski-gfsc:
Fix for intermittent test failures: some pages take too long to load.


test/system_test_helper.rb
```diff
@@ -3,6 +3,7 @@
 require 'test_helper'
 
 Capybara.app_host = 'http://lvh.me'
+Capybara.default_max_wait_time = 60
```
erbridge, line 6: Is this 60 seconds? If our pages are taking 60 seconds to load, they're too slow by far and _should_ cause test failures, imo.

kimadactyl: I remember this being an issue when it was first set up. We have a "soon" ticket in the backlog to rip out all these datatables pages as is - potentially we just remove these tests for now and bump it up the backlog. I think it's related to the slight mess of JS we have that Alfie has now broken the back of but this might be the last straggling bit.

aaaaargZombies: I don't think the time it takes capybara to load stuff reflects real world use but it is annoying.

I created these helpers when I was using it which might be made redundant by the above fix.

https://github.com/geeksforsocialchange/PlaceCal/blob/fe3adf9bbde9d214a165c1a6d68b250e9016466f/test/test_helper.rb#L127-L133

Maybe there's some math about what capybara time to real world time is and we could set the value based on that? Though maybe we are also going to end up testing pageload speeds in each system test without being explicit about it?

ivan-kocienski-gfsc: As a maximum, I believe. It could be overkill but at least with this set we know the problem is somewhere on the javascript side

ivan-kocienski-gfsc: we can whittle it down later

erbridge: Agree with you Ivan. Can we not have this close that issue, though, until we've proven this resolves it?

aaaaargZombies (approved):
sounds like this is an improvement even if we don't love it.

PR #1575, opened 2022-11-14
Stop pinning gems

erbridge:
We have a lockfile to pin them for us, so we don't need to pin them.

This should allow Renovate to automerge updates to them if they pass.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,11 +11,11 @@ end
 # Core
 gem 'minitest-rails'
 gem 'pg'
-gem 'puma', '5.6.5'
-gem 'rails', '6.1.7'
+gem 'puma'
+gem 'rails'
 
 # Frontend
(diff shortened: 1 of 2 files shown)


kimadactyl (approved):
I don't really feel qualified to know if this is a good idea or not. My concern would be that most of these pins are from official install instructions so there was prob a reason for them. But I'll approve it as it's a sensible justification and it does seem somewhat arbritrary to have them pinned to whatever it said when we first installed it.

erbridge:
@kimadactyl My understanding is that those instructions are usually a workaround to make sure people initially install the version they intend regardless of what they already have installed. Once the initial installation is done, they lose their importance.

kimadactyl:
that makes loadssss of sense. ok with it!

Issue #1572, opened 2022-11-13
Tests should not depend on the real current datetime


erbridge:
## Description

```
Failure:
PartnerIntegrationTest#test_if_theres_a_lot_of_events_show_them_with_a_paginator [/path/to/PlaceCal/test/integration/partner_integration_test.rb:84]
Minitest::Assertion: Expected at least 5 elements matching "div.event", found 0..
Expected 0 to be >= 5.
```

Lots of instances of `PartnerIntegrationTest#test_if_theres_a_lot_of_events_show_them_with_a_paginator` have been failing this weekend on CI:

```
Failure:
PartnerIntegrationTest#test_if_theres_a_lot_of_events_show_them_with_a_paginator [/path/to/PlaceCal/test/integration/partner_integration_test.rb:84]
Minitest::Assertion: Expected at least 5 elements matching "div.event", found 0..
Expected 0 to be >= 5.
```

After a bunch of investigation, I worked out that the issue is to do with the way pagination works by grouping by week:

The factory generates events starting at `DateTime.now + 1.day` and ending at `DateTime.now + 1.day + 2.hours`. If `DateTime.now` is on Sunday, then that means created events start on Monday.

Pagination groups by week and defaults to the current week. But all of our events are next week, so we have no events to show this week. So the test fails.

I'm also getting the following error, which given the assertion that fails is looking for some events, I'd suspect the a similar issue (I haven't actually looked at the code):

```
Failure:
EventsControllerTest#test_should_get_index_with_configured_subdomain [/path/to/PlaceCal/test/controllers/events_controller_test.rb:52]
Minitest::Assertion: Expected exactly 2 elements matching "ol.events li", found 0..
Expected: 2
  Actual: 0
```

We should make our tests more deterministic and stop using real time in them anywhere, as I suspect this is the cause of other test flakiness. We could use something like [Timecop](https://github.com/travisjeffery/timecop) to do that.

## Acceptance criteria

- [x] Tests are configured to freeze time to a fixed, known date
- [ ] We test known time sensitive behaviour for boundary conditions and edge cases

kimadactyl:
Excellent catch!

kimadactyl:
Is this resolved now @aaaaargZombies?

aaaaargZombies:
@erbridge said to leave this open because of

> We test known time sensitive behaviour for boundary conditions and edge cases

[...]

PR #1576, opened 2022-11-14
tests that reference time use 2022/11/8 (transDim launch)

aaaaargZombies:
see #1572
> Tests are configured to freeze time to a fixed, known date

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -30,6 +30,14 @@ end
 
 module ActiveSupport
   class TestCase
+    setup do
+      Timecop.freeze(Time.zone.local(2022, 11, 8))
+    end
+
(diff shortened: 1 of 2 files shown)


test/test_helper.rb
```diff
@@ -30,6 +30,14 @@
 
 module ActiveSupport
   class TestCase
+    setup do
+      Timecop.freeze(Time.zone.local(2022, 11, 7))
```
erbridge, line 34: I'm sorry to be that person, but could we pick a different date? Any other date is fine, just this one has complicated significance for me.

aaaaargZombies: of course

erbridge (approved):

PR #1579, opened 2022-11-14
Better renovate

erbridge:


aaaaargZombies (approved):
This looks like an improvement

kimadactyl:
is there a way we can get it to skip sending updates to discord? assuming that's too fiddly

erbridge:
Not without building an extra layer. Also, I want the notifications in Discord, otherwise dependency updates would be basically invisible, and I want to know if something maybe updated that shouldn't have or if something breaks to know that might be why.

Issue #1536, opened 2022-11-08
Support automatic importing of any ld+json schema.org events


erbridge:
## User story

As an organiser, I need to import the events of partners, so that they appear on PlaceCal.

## Acceptance criteria

- [x] If an event listing page with ld+json doesn't fit any of the other criteria for auto-detection, we detect that it contains schema.org events and use a generic ld+json importer.
- [ ] We don't have special instructions or a special importer for Squarespace.
- [ ] We have tests covering all known generators of ld+json event lists.

## Implementation notes & questions

Prompted by #1444 seeming to also use ld+json, but not be a Squarespace site.

## Implementation plan

_To be written by the developer_

kimadactyl:
It would be nice if we can find a tool (browser plugin?) that detects if a page you have has a compatible ld+json feed - be v useful for placecal organisers

erbridge:
We could knock a browser extension together for that very easily, I think.

kimadactyl:
i was thinking of using an existing one but our own one would have lots of benefits to be sure - if it just spots and gives you the correct link that would save a shitload of time for admins im guessing. esp as the vendor list grows it would prob save a lot of time for people.

aaaaargZombies:
started looking into this https://crxjs.dev/vite-plugin

erbridge:
See also: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension

However, I think the browser extension is the least important part of this. Let's spawn a new issue for that.

erbridge:
#1546 for the browser extension (later).

ivan-kocienski-gfsc:
So we try fetching a URL and if the response is HTML then we attempt to scan it for ld+json content?

I'm a little concerned that although ld+json is well structured in specific records there's an infinite variety of nesting structures that will have to be dealt with. i.e. we have two importers (Dice and OutSavvy) that are both ld+json but are just different enough to warrant their own extraction logic.

erbridge:
Isn't this what the library we're using is for? To standardise the data structures by "expanding" or "compacting"? If they're being compliant with the standards, we should be able to implement a common parser that works for any standard-compliant implementation.

ivan-kocienski-gfsc:
Do we have a list of known ld-json sources we can use to test the third criteria? It seems like it may be a bit open ended as an AC

ivan-kocienski-gfsc:
[...]

PR #1586, opened 2022-11-22
Import generic ld+json event sources

ivan-kocienski-gfsc:
Implements #1536 

LD+JSON is now handled by a single code path. OutSavvy and Dice calendars use this as a basis.


app/jobs/calendar_importer/parsers/dice_fm.rb
```diff
@@ -5,53 +5,25 @@
...
 
 module CalendarImporter::Parsers
-  class DiceFm < Base
-    NAME = 'Dice FM'
+  class DiceFm < LdJson
+    NAME = 'Dice FM (LD+JSON)'
     KEY = 'dice-fm'
     DOMAINS = %w[dice.fm].freeze
 
     def self.whitelist_pattern
```
kimadactyl, line 13: tiny note but we should swap `whitelist` for `allowlist`, not sure if part of this pr or a new one

app/jobs/calendar_importer/events/linked_data_event.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    def initialize(data)
+      super data
+
+      @url = data['url']
+      return if @url.blank?
+
+      @description = data['description']
+      return if @description.blank?
```
kimadactyl, line 27: does this mean it won't import the event if it doesn't have a description?

kimadactyl:
Wow congrats on this!! Great work!

app/jobs/calendar_importer/events/linked_data_event.rb
ivan-kocienski-gfsc: That is correct

ivan-kocienski-gfsc:
Currently I require squarespace hosted URLs so I can check the ld-json parser can pick those up on its own.

But personally I think the SS importer we have works and should just use that

app/jobs/calendar_importer/events/linked_data_event.rb
kimadactyl: I think that's too restrictive - we should be importing events permissively and then deciding what to do with them afterwards. I vaguely remember discussing this before and not sure if we made notes?

app/jobs/calendar_importer/events/linked_data_event.rb
```diff
@@ -0,0 +1,65 @@
...
+      return if @start_time.blank?
+
+      @end_time = parse_timestamp(read_value_of(data, 'end_date'))
+      return if @end_time.blank?
+
+      @summary = data['name']
+      return if @summary.blank?
+
+      loc = data['location']
+      return if loc.blank?
```
kimadactyl, line 39: same here - i dont think a location should be mandatory

app/jobs/calendar_importer/parsers/dice_fm.rb
```diff
@@ -5,53 +5,25 @@
 # Parent parser classes should not be added.
 
 module CalendarImporter::Parsers
-  class DiceFm < Base
-    NAME = 'Dice FM'
+  class DiceFm < LdJson
```
kimadactyl, line 8: importers should have links back to the source event - maybe for another PR tho

app/jobs/calendar_importer/parsers/out_savvy.rb
```diff
@@ -5,46 +5,13 @@
 # Parent parser classes should not be added.
 
 module CalendarImporter::Parsers
-  class OutSavvy < Base
-    NAME = 'OutSavvy'
+  class OutSavvy < LdJson
```
kimadactyl, line 8: importers should have links back to the source event - maybe for another PR tho

kimadactyl:
This is prob a fresh ticket but I just realised we've not been writing footer methods for some of these new importers

Example:

https://github.com/geeksforsocialchange/PlaceCal/blob/fe3adf9bbde9d214a165c1a6d68b250e9016466f/app/jobs/calendar_importer/events/ticketsolve_event.rb#L19

It should be possible to link back to the original event to get tickets or more info

app/jobs/calendar_importer/events/linked_data_event.rb
ivan-kocienski-gfsc: it does sound familiar. its not validated in the model so i will fix that

app/jobs/calendar_importer/parsers/out_savvy.rb
ivan-kocienski-gfsc: if you mean events having a supplied URL then that already happens.

But i think you mean a new field on events that has a "this was downloaded from this page" url (?) which is defo a new feature.

app/jobs/calendar_importer/events/linked_data_event.rb
ivan-kocienski-gfsc: This has been fixed

app/jobs/calendar_importer/parsers/out_savvy.rb
ivan-kocienski-gfsc: the `publisher_url`?

kimadactyl: look at the comment further down this pr - there's a `footer` method

app/jobs/calendar_importer/calendar_importer.rb
```diff
@@ -16,19 +16,34 @@ class InaccessibleFeed < StandardError; end
...
-        PARSERS.find { |parser| parser.handles_url?(@calendar.source) }
-      else
-        PARSERS.find { |parser| parser::KEY == @calendar.importer_mode }
+    return @parser if @parser
+
+    if @calendar.importer_mode == 'auto'
+      @parser = DETECTABLE_PARSERS.find { |parser| parser.handles_url?(@calendar.source) }
+
+      if @parser.blank?
+        # TODO: now try to detect ld+json
```
erbridge, line 37: I think this `TODO` is done by the implementation. By attempting to use it we validate if it's valid.

app/jobs/calendar_importer/calendar_importer.rb
```diff
@@ -16,19 +16,34 @@ class InaccessibleFeed < StandardError; end
...
+        try_parser = CalendarImporter::Parsers::LdJson.new(@calendar)
+        nodes = try_parser.download_calendar
+        @parser = CalendarImporter::Parsers::LdJson if nodes.present?
       end
+
+      return @parser
+    end
+
+    @parser = MANUAL_PARSERS.find { |parser| parser::KEY == @calendar.importer_mode }
```
erbridge, line 46: I'm not sure we need all this? Can we just make `CalendarImporter::Parsers::LdJson.handles_url?` do:
```rb
try_parser = CalendarImporter::Parsers::LdJson.new(@calendar)
try_parser.download_calendar.present?
```
or similar and leave the rest of the detection logic as it was?

app/jobs/calendar_importer/event_resolver.rb
```diff
@@ -214,7 +214,7 @@ def save_all_occurences
 
       event_time[:are_spaces_available] = occurence.status if occurence.respond_to?(:status)
 
-      unless event.update(data.attributes.merge(event_time))
+      unless event.update!(data.attributes.merge(event_time))
```
erbridge, line 217: Doesn't this mean we never get errors reported to the user because the `!` makes it cause an exception instead? Why did we make this change?

app/jobs/calendar_importer/events/linked_data_event.rb
```diff
@@ -0,0 +1,70 @@
...
+      return if @end_time.blank?
+
+      @summary = data['name']
+      return if @summary.blank?
+
+      @location = extract_location(data)
+      # return if @location.blank?
+
+      @uid = @url
+      @is_valid = true
```
erbridge, line 51: Instead of exiting with partial data, wouldn't this be better to set `@is_valid` to check for presence of the required fields? So we always populate the fields we can find even if it's parked as invalid at the end?

app/jobs/calendar_importer/events/linked_data_event.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def attributes
+      valid? && in_future? && super
+    end
+
+    def valid?
+      @is_valid
```
erbridge, line 59: Could we just check the required attributes here instead of having a static `@is_valid` we have to keep up-to-date?

app/jobs/calendar_importer/parsers/dice_fm.rb
```diff
@@ -5,53 +5,25 @@
...
-
-      events = place_record['http://schema.org/event']
-
-      events
-        .map { |event_hash| CalendarImporter::Events::DiceFmEvent.new(event_hash) }
-        .keep_if(&:event_record?)
+    module EventConsumer
+      def consume_place(data)
+        Rails.logger.debug 'consume_place'
+        consume data['event']
```
erbridge, line 25: What's this doing that isn't done by the `LdJson` importer and why can't the `LdJson` importer do it instead? In an ideal world, we wouldn't have a special "Dice" importer at all, because the `LdJson` one does everything we need.

app/jobs/calendar_importer/parsers/out_savvy.rb
erbridge, line 8: I don't think we need an `OutSavvy` importer anymore. The `LdJson` one will just work for it instead, right? At least that's the point of this work to be so that we don't need to care about different providers.

kimadactyl: see [this ticket](https://github.com/geeksforsocialchange/PlaceCal/issues/1589) - i think at a bare min its nice to have a linkback to the source event and its hard to imagine we could autogenerate this. its prob fine to delete now and remake it when we need it tho.

i think its also nice to autodetect urls we know work? in a future iteration we can poss do some earlier autotedetection of the url, give some contextual advice to get to the right page, etc

erbridge: We're autodetecting already in these changes. We just do it by attempting to fetch events and seeing if it works.

erbridge: And to give backlinks, we have a `url` method we populate with the event's URL which we can use already.

app/jobs/calendar_importer/parsers/dice_fm.rb
ivan-kocienski-gfsc: i'll try and explain in more detail with examples when i have a fresher mind but if you're curious OutSavvy tucks their events under a `@graph` node and Dice hides their events in a `Place` record.

app/jobs/calendar_importer/parsers/out_savvy.rb
erbridge: Right, but this one doesn't actually override anything, does it? It's just plain old `LdJson` with a label.

ivan-kocienski-gfsc: But it does have `NAME`, `KEY` and `DOMAINS` that are used on the front end

ivan-kocienski-gfsc: i have done as you said. its now in only one parser.

app/jobs/calendar_importer/parsers/dice_fm.rb
ivan-kocienski-gfsc: This is now resolved into one unified ld-json parser

app/jobs/calendar_importer/calendar_importer.rb
ivan-kocienski-gfsc: Doing this now

app/jobs/calendar_importer/parsers/ld_json.rb
```diff
@@ -0,0 +1,225 @@
...
+          graph = data['@graph']
+          if graph.present?
+            graph.each do |graph_node|
+              consume graph_node
+            end
+            return
+          end
+
+          type = data['@type']
+          # puts type
```
aaaaargZombies, line 161: shall we pull these comments out?

aaaaargZombies (approved):
I made a small recommendation but I don't think it should block the merge

PR #1597, opened 2022-11-29
Scope events for calendars with UID

### LINKED ISSUE #1596: [Bug]: Duplicate Events from EventBrite calendar source
## Description

Calendar on staging/production is stuck `in_worker`: https://admin.placecal-staging.org/calendars/76/edit

The source URL is: https://www.eventbrite.com/o/research-in-arts-and-humanities-at-manchester-met-4168900447

Investigating seems to show a discrepency between how the [importer](https://github.com/geeksforsocialchange/PlaceCal/blob/6e98ddac7dfa7467635d74c5334e547c7762b33a/app/jobs/calendar_importer/event_resolver.rb#L206) checks for duplicate events and how the [event](https://github.com/geeksforsocialchange/PlaceCal/blob/6e98ddac7dfa7467635d74c5334e547c7762b33a/app/models/event.rb#L180) model validates duplicate events (involving the event UID).

This may be symptom of a deeper problem with how duplicate events are dealt with: in this context recurring events could be defined simply (for ICS sources where a single event can have multiple recurring instances) but many of the calendar events come from APIs that do not have such a feature.

ivan-kocienski-gfsc:
Fixes #1596 

This harmonizes the event duplication check in the calendar importer with that in the event model


ivan-kocienski-gfsc:
@erbridge This is the code fix that has me a little out-of-my-depth as it touches on business logic. 

TLDR: Events should be unique but how do we validate that uniqueness? In the event model it is scoped by `dtstart`, `summary` and `calendar_id` but in the importer it uses the same fields and a `UID` field. 

The original bug stems from an EventBrite calendar that had events with identical start, summary and calendar IDs but differ in UID. This is not picked up by the importer (as the UIDs are unique) but when the calendar was saved the invalid events would throw an exception causing the save to abort.

Essentially the importer can ingest repeating events from ical sources. It does this by looking at the repeat rules (for one upstream event) and creating a bunch of event records for each instance. To do this there is a UID field that ties all the events together.  This logic is only used for ical and ticketsolve calendar event creation (where UID is guarenteed to exist). 

So the question is: do we constrain event uniqueness by keying off of UID as well as the other fields or not? And if we do how does this impact other calendars?

app/models/event.rb
```diff
@@ -178,7 +178,8 @@ def require_location
   # Ensures that the event added is unique
   # Checks for a duplicate event with the properties dtstart, summary, and calendar_id
   def unique_event
-    return unless Event.where(dtstart: dtstart,
+    return unless Event.where(uid: uid,
```
aaaaargZombies, line 181: if the `uid` is unique do we still need to check `dstart`,`summary`, `calendar_id` ?

aaaaargZombies (changes requested):

app/models/event.rb
ivan-kocienski-gfsc: i felt making the scope tighter maintains backward compatibility with the rest of the event source importers whilst adding extra checks. 
in an ideal world UID would be unique per event (per calendar) and we could rely on it to determine uniqueness but determining if that is the case would be some effort.

aaaaargZombies: thanks for explaining, good to know that might not be the case.

PR #1626, opened 2022-12-21
delink partner addresses

### LINKED ISSUE #1615: [Bug]: partner address linked
## Description
quote from discord

so there's one contact who runs 2 groups - Queer Running Club & Pecs Education. When I change the address of one, it changes the address of the other. Any idea why this happens?
these are the correct addresses

Pecs Education is here:
Basis Ldn:
361-362 Warbuton street
E8 3RR

Queer Running Club is here;
Brunswick East
Ground Floor,
17 Amhurst Terrace,
Lower Clapton,
E8 2BT

only obvious link in the admin UI seems to be the shared admin
```
Partner admins
1 user can edit this partner
    [kole.fulmine@gmail.com](https://admin.placecal.org/users/111/edit)
```
## Steps to reproduce

1. What you did first
1. What you did next

## Platform (if relevant)

aaaaargZombies:
fixes #1615 

If a partner has an address already when a user goes to edit it, the form adds a hidden field containing the `id` of the address that the partner belongs to. When the form is submitted it updates the address at that `id` so all partners which reference it have their addresses updated. This is useful if lots of groups use the same community space and it moves or is renamed.  It's less good if one group moves or was added to the address by mistake.

To solve this I've added a simple toggle to remove/include this field. It's off by default so that groups won't have their addresses changed by accident. If the new address exists already it will be assigned to that `id` or a new one will be created if not. If the address fields are left the same when updating it will retain the same `id`.

### DIFF
--- a/app/javascript/controllers/address_toggle_controller.js
+++ b/app/javascript/controllers/address_toggle_controller.js
@@ -1,0 +1,29 @@
+import { Controller } from "@hotwired/stimulus";
+
+// Connects to data-controller="address-toggle"
+export default class extends Controller {
+	static values = { state: Boolean };
+	static targets = ["toggle"];
+
+	connect() {
+		this.addressId = this.element.querySelector(
+			"#partner_address_attributes_id"
+		);
+		this.toggleTarget.checked = this.stateValue;
+		this.stateValue ? this.linkAddress() : this.delinkAddress();
+	}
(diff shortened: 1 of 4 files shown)


aaaaargZombies:
we're changing the model to deal with this so these changes are not needed

PR #1675, opened 2023-02-04
Neighbourhood explorer

ivan-kocienski-gfsc:
As a free time project.

As an Admin of the site I would like a way to navigate Neighbourhoods so I can see the parents and descendants of a neighbourhood in one simple page. I would also like to see what Address Partners, Service Area Partners and Sites this Neighbourhood contains (with easy access to those listed).


kimadactyl:
This is cool - is it ready for review? Be good if @aaaaargZombies can do it I'm not sure the current state of my local PlaceCal!

app/views/admin/neighbourhoods/show.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  <li class="breadcrumb-item active"><%= safe_neighbourhood_name(@neighbourhood) %></li>
+</ol>
+
+<h1><%= safe_neighbourhood_name(@neighbourhood) %></h1>
+
+<p><%= link_to 'Edit', edit_admin_neighbourhood_path(@neighbourhood), class: 'btn btn-primary mb-3' %></p>
+
+<% if @neighbourhood.children.present? %>
+<% @neighbourhood.children.order(:name).each do |child| %>
+<p>
```
aaaaargZombies, line 14: I think things like this should be lists not paragraphs

aaaaargZombies (approved):
None blocking suggestion. It would be better for accessibility / semantics if the lists of sites/partners/etc could be in html lists.

Otherwise this is a nice way to see what exists where!

PR #1676, opened 2023-02-06
Validate partner service areas / addresses on save

### LINKED ISSUE #1663: [Bug]: Partner created without service area or address
## Description

A partner has been created without either piece of location info. I didn't create it so not sure how.

https://admin.placecal.org/partners/oshun-abi-dougill/edit

https://admin.placecal.org/partners/oshun-abi-dougill/edit

This should be impossible - PlaceCal should require either an address or a service area

Screenshot 2023-01-30 at 15 26 30

ivan-kocienski-gfsc:
Fixes #1663 

Not just on create

### DIFF
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -89,7 +89,7 @@ class Partner < ApplicationRecord
   validate :check_ward_access, on: :create
   validate :check_service_area_access, on: :create
 
-  validate :must_have_address_or_service_area, on: :create
+  validate :must_have_address_or_service_area
 
   attr_accessor :accessed_by_user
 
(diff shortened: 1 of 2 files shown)


kimadactyl (approved):
Cool - just a note to fix the partners in a broken state when we get this live

ivan-kocienski-gfsc:
Good point but how do we do that? Give them a service area or an address? Like a default address/sa? Something pulled from a calendar? Just checked on prod and there are only 7 partners with missing service areas and no addresses (20, 22, 23, 28, 33, 27, 223)

kimadactyl:
> Good point but how do we do that? Give them a service area or an address? Like a default address/sa? Something pulled from a calendar? Just checked on prod and there are only 7 partners with missing service areas and no addresses (20, 22, 23, 28, 33, 27, 223)

I'll ping them back to the PHT team to fix

PR #1677, opened 2023-02-07
Fix eventbrite calendar parser raising a "bad gateway" error

### LINKED ISSUE #1658: [Bug]: RestClient::BadGateway: 502 Bad Gateway
https://rollbar.com/gfsc/placecal-production/items/243/
https://rollbar.com/gfsc/placecal-production/items/205/
https://rollbar.com/gfsc/placecal-production/items/247/
https://rollbar.com/gfsc/placecal-production/items/253/
https://rollbar.com/gfsc/placecal-production/items/256/

Look like the same problem

```
File /app/app/jobs/calendar_importer/parsers/eventbrite.rb:26 in download_calendar
File /app/app/jobs/calendar_importer/parsers/base.rb:32 in calendar_to_events
File /app/app/jobs/calendar_importer/calendar_importer_task.rb:52 in calendar_source
File /app/app/jobs/calendar_importer/calendar_importer_task.rb:56 in event_data_from_parser
File /app/app/jobs/calendar_importer/calendar_importer_task.rb:15 in run
File /app/app/jobs/calendar_importer_job.rb:47 in perform
```

ivan-kocienski-gfsc:
Fixes #1658


kimadactyl (commented):
- Is there a way this can give more helpful feedback to end users? "The URL was found, but gave an error" or something?
- Will it continue to retry if this happens? I think it should

ivan-kocienski-gfsc:
Hmmm. I don't think there is a more English way of describing this fix without over simplifying the solution and making the title more confusing to people reading the git log.

- *Calendar Importer*: the part of the code that this bug is in
- *Eventbrite parser*: When dealing with Eventbrite sources the problem is in how we download the calendar from that API.
- *BadGateway*: the remote API request is giving us an HTTP failure.

ivan-kocienski-gfsc:
If a user got the feedback from the Calendar (importer) that there was a `BadGateway` response then how will they correct this problem (if it is there problem)?

We could track this problem in the calendar's notices field.

kimadactyl:
> If a user got the feedback from the Calendar (importer) that there was a BadGateway response then how will they correct this problem (if it is there problem)?

Not sure the most common cause of it but i think if im told the response is bad then it might be because im on the wrong url wheeras BadGateway is scary computer talk?

ivan-kocienski-gfsc:
As far as I can tell the bad response from Eventbrite is indicating that their servers are overwhelmed and they could not deal with our request at that time. The solution in this PR is to ignore it as their servers will become available later.

I guess I am wondering if we mark the calendar as having a `bad_source` and telling the user something went wrong the only thing the user can do is to retry the calendar (maybe can they do that?) or change the URL.

ivan-kocienski-gfsc:
We're not giving them any error. We are catching the problem and pretending the API gave back an empty response. As this is a upstream problem with Eventbrite it's their problem that their service is intermittent. If Eventbrite go completely offline then we tell the user that the URL no longer functions and put the calendar into a `bad_source` state.

kimadactyl:
So you get no user feedback at all if the calendar URL errors?

ivan-kocienski-gfsc:
No: all the error catching code is still in place and works as it has done before.

In this one case of `BadGateway` exception from the Eventbrite API, we ignore it. The calendar will be queued up again and it should work later.

But only for Eventbrite BadGateway problems as they are not really anything to do with us (or our users).

kimadactyl:
WhatsApp Image 2023-01-25 at 15 08 32

what about this?

ivan-kocienski-gfsc:
Not part of this ticket.

It may be an idea to put in an extra field that can catch better information like this for diagnostic purposes so users can have a simpler, friendlier message but we can capture more detailed engineering information.

kimadactyl (approved):
Ah OK I think we were talking cross purposes. Yeah OK sounds like a good ticket to write up as well, will approve this

Issue #1654, opened 2023-01-25
Allow multiple different kinds of "tags"


kimadactyl:
## User story

As a PlaceCal admin, I want to be able to have tags split into different types so they can be used in different ways, in order that we can improve partner metadata and UX

## Acceptance criteria

- [x] "tags" can be of type "facility", "category" or "partnership"
- [x] these are visible in the admin tags index interface both separately and together
- [x] tags are alphebatized
- [x] current tags are assigned the approprate type on prod and staging
- [ ] #1737

## Implementation notes & questions

I think this is a clear case for [Single Table Inheritance](https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html).

## Implementation plan

- [x] put type field on Tags table
- [x] #1687
- [x] #1690
- [x] #1692
- [x] #1695
- [x] #1702
- [x] #1708
- [ ] update controller/model/policy/tests to allow tag selection (as needed)

(guest user front end work will be done after this issue is closed)

ivan-kocienski-gfsc:
I have had a pass at this and so far I have identified 3 kinds of tags on PlaceCal:
- *Property Tags* - on Partners to indicate they have a particular service or feature. i.e. Accessible Toilette, Free Wifi, Public Computers
- *Category Tags* - on Partners and Events to indicate the (primary) subject of the event. i.e. cooking, biking, music
- *Site Tags* - that link Partners to specific Sites, like Trans Dimension

kimadactyl:
I like these but just vote to rename "properties" as "facilities"? i think the former is a bit programming-y?

kimadactyl:
The last step here is huge scope creep and covered by other tickets

ivan-kocienski-gfsc:
Could I get an expansion on what you meant by "these are visible in the admin interface both separately and together"?

Do you mean: Users can select all types of tags from the one tags box?
Or do you mean: There will be three separate (tag) boxes: Categories, Facilities and Partnerships?

kimadactyl:
I meant that

- You can see the list of all tags
- you can also see the list of just categories
- you can also see the list of just facilities
- you can also see the list of just partnerships

i think eventually we want to discontinue the first one entirely - but for now at least make it so we can browse each category individually?

its really seeming like replacing datatables is becoming a big blocker huh

kimadactyl:
possible inspiration: https://github.com/geeksforsocialchange/gofg-db/blob/master/app/views/people/index.html.erb

[...]

PR #1682, opened 2023-02-13
WIP Tag differentiation

ivan-kocienski-gfsc:
Partially Implements #1654

Start of Tag Specification.
    
- Tag table type migration
- All tags default to FacilityTag for now
- Tag subclasses for FacilityTag, PartnershipTag and CategoryTag
- Improved admin tags index page
- Fix of tests

### DIFF
--- a/lib/tasks/db_load.rake
+++ b/lib/tasks/db_load.rake
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+namespace :db do
+  namespace :load do
+    desc 'Loads the default category tags into the tags table'
+    task default_categories: :environment do
+      categories = [
+        'Activism',
+        'Arts & Crafts',
+        'Children & Family',
+        'Young Adults',
+        'Over 50s',
+        'Entertainment',
+        'Education',
+        'Food',
+        'Health & Wellbeing',
+        'Outdoors',
+        'Sports & Fitness',
+        'Places of worship',
+        'Community Hubs',
+        'Housing',
+        'Legal Advice',
+        'Immigration',
+        'LGBTQ+',
+        'Communities of Colour'
+      ]
+
+      categories.each do |category|
+        CategoryTag.find_or_create_by!(name: category)
+      end
+    end
+  end
+end
--- a/lib/tasks/db_load.rake
+++ b/lib/tasks/db_load.rake
@@ -16,7 +16,7 @@ namespace :db do
         'Health & Wellbeing',
         'Outdoors',
         'Sports & Fitness',
-        'Places of worship',
+        'Places of Worship',
         'Community Hubs',
         'Housing',
         'Legal Advice',
--- a/test/system/admin/tag_test.rb
+++ b/test/system/admin/tag_test.rb
@@ -28,11 +28,11 @@ class AdminTagTest < ApplicationSystemTestCase
     click_link 'Tags'
     click_link @tag.name
 
-    partners = select2_node 'facility_tag_partners'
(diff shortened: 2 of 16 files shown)


lib/tasks/db_load.rake
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+namespace :db do
```
kimadactyl, line 3: let's namespace this to `tags`?

test/system/admin/tag_test.rb
```diff
@@ -15,7 +15,7 @@ class AdminTagTest < ApplicationSystemTestCase
 
     @partner = @partner_admin.partners.first
     @partner_two = create :ashton_partner
-    @tag = create :tag
+    @tag = create(:tag, name: 'Hulme 1 Tag')
```
kimadactyl, line 18: i think this is an unhelpful name for a tag here - its unclear to me if this is a category or place we are testing?

lib/tasks/db_load.rake
ivan-kocienski-gfsc: I used this namespace as in the context of a rails script run from the console it is to do with the DB. It creates the `db:load:...` namespace that loads data into the database. It is used as a place to put one off scripts for... loading data into the database. I think it is clearer as it is. Are we going to have a `partners` namespace, a `partnerships`, a `calendars` ? this will clutter up the rails task output and make it harder to find (or guess at) that useful data loading script

lib/tasks/db_load.rake
```diff
@@ -0,0 +1,33 @@
...
+        'Children & Family',
+        'Young Adults',
+        'Over 50s',
+        'Entertainment',
+        'Education',
+        'Food',
+        'Health & Wellbeing',
+        'Outdoors',
+        'Sports & Fitness',
+        'Places of worship',
```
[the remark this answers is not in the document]
ivan-kocienski-gfsc: This is just sample data that has no functional bearing on the code at the moment but okay.

test/system/admin/tag_test.rb
ivan-kocienski-gfsc: The test breaks without the ' Tag' postfix. At the moment we aren't testing anything like that as they don't exist, as per your instructions

lib/tasks/db_load.rake
kimadactyl: But this isn't setting up a generic database which is what the `db` namespace is for - its our specific seed data for our specific instance of PlaceCal. It clashes with the existing Rails namespace. Other people using it would prob not use our categories, or at least we should decouple it

How about we start a namespace just for this? `pht` or something?

I do think they need to be seeded individually yes - we need granular control as once added we won't be deleting them. One big `load` task seems bad to me.

test/system/admin/tag_test.rb
kimadactyl: Sure - just saying there should never be a tag named after a place - so maybe call it `Arts and Crafts Tag` or something for now or it'll get confusing!

lib/tasks/db_load.rake
ivan-kocienski-gfsc: i would suggest then moving all of our scripts into the `placecal` namespace.

- move only this script (adhoc, need to move rest later, making more issue chaff)
- move all scripts (feature creep, would have to update all scripts, would have to update docs)

ivan-kocienski-gfsc:
This is overly complicated.

PR #1689, opened 2023-02-16
Admin tag type visibility

### LINKED ISSUE #1687: Make tag type visible to admin users
Acceptance Criteria:
- [x] Tag type can be seen on admin create tag form (but not on edit form)
- [x] Tag type can be seen on admin tags index page

ivan-kocienski-gfsc:
fixes #1687 

- tags can be created with a type
- editing a tag shows the type
- admin tag index lists the tags with their type
- updated tests

### DIFF
--- a/app/helpers/tags_helper.rb
+++ b/app/helpers/tags_helper.rb
@@ -21,14 +21,4 @@ module TagsHelper
       value
     end
   end
-
-  TAG_TO_NAME = {
-    'CategoryTag' => 'Category',
-    'PartnershipTag' => 'Partnership',
-    'FacilityTag' => 'Facility'
-  }.freeze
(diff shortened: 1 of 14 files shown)


app/helpers/tags_helper.rb
```diff
@@ -21,4 +21,14 @@ def edit_permission_label(value)
       value
     end
   end
+
+  TAG_TO_NAME = {
```
kimadactyl, line 25: should we rename the base tag names now or do you think it's ok?

kimadactyl (approved):
lgtm - only choice is if you want to change `FacilityTag` to just `Facility` etc before it gets too embedded in the codebase??

PR #1696, opened 2023-02-17
Tag filter for admin tags index page

### LINKED ISSUE #1695: Put in filter on admin tags index page
As an admin user who is looking at tags on the tags index page I would like some filters/pagination so I can better locate the tag I am interested in.

## Acceptance Criteria

- [x] Filter by type (Facility, Partnership or Category)
- [x] Basic name search
- [x] Pagination (page number and per-page limit)

ivan-kocienski-gfsc:
Closes #1695 

Filter by
- Type
- Name (very crudely)
- Pagination


test/controllers/concerns/tag_filter_test.rb
```diff
@@ -0,0 +1,219 @@
...
+      name: 'alpha',
+      per_page: '20',
+      page_num: '2'
+    }
+    filter = TagFilter.new(params)
+
+    output = filter.next_page_link(FakeView.new, Tag.all)
+    assert output.is_a?(String)
+
+    expected_html = '<a class="btn btn-link" href="/tags?name=alpha&page_num=3&per_page=20&type=facility">Next ...</a>'
```
aaaaargZombies, line 216: Can we make this a button and add a "previous" if we've already clicked "next?

aaaaargZombies (approved):
works nicely - I left a comment for a possible improvement that I'll leave up to you.

test/controllers/concerns/tag_filter_test.rb
ivan-kocienski-gfsc: Maybe? But there is a back button on their browser (hope that doesn't come off as too sarcastic).

When ever I see things like this one sites that do the whole pagination circus with `<< < 0 2 3 ... 45 > >>` type things i think about how i only really use the next page link.

aaaaargZombies: > Maybe? But there is a back button on their browser

happy with this

I do think next would benefit from looking like a button though but the PR is still good overall

PR #1700, opened 2023-02-20
Make rg search respect gitignore

aaaaargZombies:
I don't have tags set up on my system (maybe I should) but the current .gitignore breaks telescope which I use to search.

Does tags just live in a folder at root? would this meet everyones of our needs?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,6 @@ doc/**/*.html
 doc/**/*.css
 doc/**/*.js
 .yardoc
-./tags/
+./tags
 Makefile
 doc/rdoc
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,6 @@ doc/**/*.html
 doc/**/*.css
 doc/**/*.js
 .yardoc
-tags
+./tags/
 Makefile
 doc/rdoc


kimadactyl:
don't quite understand this sorry - where is this in the repo?

aaaaargZombies:
Sorry poor explanation. It's in the .gitignore

Ivan (maybe others) use [ctags](https://en.m.wikipedia.org/wiki/Ctags) with vim to help with code navigation. Because it generates a bunch of index cruft that's not part of the code we don't check it into git.

But a bunch of tooling will use gitignore to exclude files. Particularly the search tool I like most in my editor and command line.

I don't really understand why we are still able to track files in a `tag` folder as is. Maybe they were already tracked before this was added.

When reviewing the last PR this caused me some bother.

kimadactyl:
ah i see so completely unrelated to the current work on the tags system in placecal? got it :)

kimadactyl (approved):

aaaaargZombies:
@ivankocienski can you cast your eyes over this, I want to make sure I've covered the correct thing

ivan-kocienski-gfsc:
i will take a look now

PR #1706, opened 2023-02-22
Validate Address postcodes

### LINKED ISSUE #1705: [Bug]: Nil exception on production
## Description

https://rollbar.com/gfsc/placecal-production/items/277/

>  ActionView::Template::Error: undefined method `name' for nil:NilClass

From client site: consciouscollectivemcr.placecal.org

Source line: /app/app/views/admin/partners/edit.html.erb:8
https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/views/admin/partners/edit.html.erb#L8

ivan-kocienski-gfsc:
Fixes #1705 

When an address model is saved, validate the postcode can be mapped to a neighbourhood.

(with tests)


ivan-kocienski-gfsc:
The address resolution code is very confusing and could do with being streamlined. This PR just enforces the bare minimum of making sure postcodes actually exist

kimadactyl:
This is great work but just let me write up how it's _meant_ to work first. I'll try and do that on Monday. Then we can check that we are on the same page.

kimadactyl (approved):

PR #1716, opened 2023-03-01
Restrict creating, editing and deleting tags to root users

### LINKED ISSUE #1712: Only roots can edit Tags
## User story

As a PlaceCal root admin, I want to stop anyone else editing Tags, so that we get some stability in the system

## Acceptance criteria

- [ ] Only roots can create, edit and delete Tags

## Implementation plan

ivan-kocienski-gfsc:
Closes #1712 

Only root users can CRUD tags.

Temporary fix so we can review and define proper granular tag access controls

### DIFF
--- a/app/policies/tag_policy.rb
+++ b/app/policies/tag_policy.rb
@@ -14,26 +14,11 @@ class TagPolicy < ApplicationPolicy
   end
 
   def edit?
-    user.root? || user.partner_admin? || user.tag_admin?
+    user.root?
   end
 
   def update?
-    return true if user.root?
-
-    # system tags can only be edited by root
-    return false if @record.system_tag
-
-    # If the user is a tag admin and has been assigned this tag
-    return true if user.tag_admin? && user.tags.include?(@record)
-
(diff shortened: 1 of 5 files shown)


kimadactyl:
It should be that root users can create, edit, delete. And that everyone with an account can read? Does this remove everyone's read access?

kimadactyl:
Ugh - it also occurs to me that Partnership Admins can currently edit their partnership from the Tag Edit interface and if we do this then they won't be able to any more right? Maybe katja was right that we do need to do this properly

ivan-kocienski-gfsc:
> It should be that root users can create, edit, delete. And that everyone with an account can read? Does this remove everyone's read access?

We don't have a "read" page. It just links to the edit form. Would be weird to allow them to see the edit form and not allow them to change anything.

I would also recommend taking some time to think about this properly

kimadactyl:
Good point. Yeah ok.

katjam:
Closing as this was a quick fix that prevented use of tags. We're fixing it properly.

PR #1742, opened 2023-03-16
fix bad opening_times in partners

### LINKED ISSUE #1726: [Bug]: Tags selector incorrectly displaying as select box not select2
## Description

Looks like some JS not loading

https://admin.placecal.org/partners/flourish/edit

Screenshot 2023-03-06 at 17 51 42

aaaaargZombies:
fixes #1726

this solves our urgent issue but highlights problems we have with storing the opening times in the way we do.


aaaaargZombies:
I've captured the larger issue here #1743

ivan-kocienski-gfsc:
I am not sure what our policy on doing data migrations is for PC but this looks okay.

How does this stop the wrong value from being sent back from JS land and into the partners in the first place? Do we have any validations on that opening_times value? Because as it stands this only runs once?

Is there another issue/pr that adds checks/validations to that field?

aaaaargZombies:
discord comments

> Kim — Today at 4:08 PM
I think we found it's not good doing data stuff in migrations and better to put in rake tasks then run them whenever is my only comment (ugh still no email from wellcome my final thing I am waiting for)
IvanKocienski — Today at 4:08 PM
that is what i thought i remembered hearing
Kim — Today at 4:09 PM
Yes I don't remember why but we deffo had some problems with it
This feels sort of like a data health tuneup task to me

aaaaargZombies:
- validates that `opening_times` is either `nil` (it's initial state) or valid JSON (any time after it's been saved by the frontend)
- add rake task to cleanup any partners with the bad string data in found in original issue

ivan-kocienski-gfsc (approved):
This looks okay to me

PR #1747, opened 2023-03-21
Update the readme + add a makefile to simplify setup

### LINKED ISSUE #1748: Speed up the install process for new GFSC devs
## User story

As a GFSC dev I need to have more complete setup docs so that I can get placecal running locally.

## Acceptance criteria

- [x] The readme contains all the required steps & info to get a local version running with a working database
- [x] The readme does not contain any sensitive info

r-ferrier:
fixes #1748

This PR aims to improve the setup & running experience for any future & current GFSC devs, especially those using docker. Using all of the helpful notes I've been given, I've distilled the list of things needed to be done to get a working version of PlaceCal with a copy of some viable data running locally. 

To enable quick replication of these steps I have also:
- simplified the readme by making explicit a quick setup route for GFSC developers, and extracting some of the larger chunks of code into a makefile
- Added a makefile that contains all the commands needed to set up a GFSC environment (contains no sensitive information)
- Added other commands to the makefile required to run & tear down the environment, set up a docker db
- Added Ivan's makefile commands to run tests and use tags in linux. 

Totally fine if everyone disagrees this should be in the repo. In my mind, if it's a useful time saving script for one of us it's probably worth sharing with everyone else, but understand that this is at odds with reducing code bloat so if we hate the idea of adding this please let me know.

Also should add the most pain actually came from setting up ruby & postgres, once I had those in place the rest of this was sort of ok!

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,9 +1,9 @@
 #REQUIREMENTS BEFORE RUNNING MAKE ALL: ruby, yarn, docker, graphviz, imagemagick, nvm, postgres
-.PHONY: test
+.PHONY: test setup
 
-all: install_dependencies docker setup_env setup_db seed_db create_user run
+all: update test
 
-test_all: update test
+setup: install_dependencies docker setup_env setup_db seed_GFSC_prod_copy_db create_user run
 
 docker: setup_docker_network setup_docker_container setup_env
 
@@ -40,10 +40,10 @@ setup_env:
 	echo PGUSER=postgres >> .env
 
 setup_db:
-	bundle exec rails db:setup db:migrate db:seed
+	bundle exec rails db:setup db:migrate
 	bundle exec rails import:all_events
 
-seed_db:
+seed_GFSC_prod_copy_db:
 	rails db:dump_production_and_restore_other restore_on_local=1
 
 create_user:
--- a/makefile
+++ b/makefile
@@ -1,0 +1,59 @@
+#REQUIREMENTS BEFORE RUNNING MAKE ALL: ruby, yarn, docker, graphviz, imagemagick, nvm, postgres
+.PHONY: test
+
+all: install_dependencies docker setup_env setup_db seed_db create_user run
+
+test_all: update test
+
+docker: setup_docker_network setup_docker_container setup_env
+
+run:
+	docker start placecal-db || true
+	bin/dev
+
+down:
+	docker stop placecal-db
+
+# does not remove postgres image, just placecal container & network
+clean:
+	docker stop placecal-db || true
+	docker container rm placecal-db || true
+	docker network rm placecal-network
+
+install_dependencies:
+	yarn
+	bundle install
+
+setup_docker_network:
+	docker network create placecal-network
+
+setup_docker_container:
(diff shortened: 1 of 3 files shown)


aaaaargZombies (approved):
makefile looks good to me, it's how I have things set up so being able to run it rather than scratching around to find what I did in my notes is better.

docker is pretty system agnostic so hopefully that settles Katjas concerns with the other guide and anyone who doesn't want to use it is still free to setup postgres locally

:+1:

makefile
```diff
@@ -0,0 +1,59 @@
+#REQUIREMENTS BEFORE RUNNING MAKE ALL: ruby, yarn, docker, graphviz, imagemagick, nvm, postgres
+.PHONY: test
+
+all: install_dependencies docker setup_env setup_db seed_db create_user run
```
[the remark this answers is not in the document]
ivan-kocienski-gfsc: Also it assumes that the developer is running a dockerised postgres.

makefile
```diff
@@ -0,0 +1,59 @@
...
+	echo POSTGRES_USER=postgres >> .env
+	echo PGPASSWORD=foobar >> .env
+	echo PGHOST=localhost >> .env
+	echo PGUSER=postgres >> .env
+
+setup_db:
+	bundle exec rails db:setup db:migrate db:seed
+	bundle exec rails import:all_events
+
+seed_db:
```
ivan-kocienski-gfsc, line 46: This is cloning production database to your local env which is not really seeding with minimal data. You also call this step after calling `setup_db` which also runs the `rails db:seed` command.

makefile
```diff
@@ -0,0 +1,59 @@
...
+	rails db:dump_production_and_restore_other restore_on_local=1
+
+create_user:
```
ivan-kocienski-gfsc, line 49: This command duplicates behaviour from the `rails db:seed`  which also [creates an user](https://github.com/geeksforsocialchange/PlaceCal/blob/292638633b3604dd7ed5043e229165ae0009e05e/db/seeds/001_users.rb#L7).

r-ferrier: ah! Will remove `db:seed` from the db setup

makefile
```diff
@@ -0,0 +1,59 @@
...
+	bundle exec rails runner "User.create!(email: 'info@placecal.org', password: 'password', password_confirmation: 'password', role: :root)"
+
+update:
```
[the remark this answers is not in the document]
r-ferrier: I know what you mean. I namespaced it so that I could chain it with the test script you're running without changing that script, because the readme asks that `bin/update` be run before tests are run, and I know this is something I'll forget, but I didn't want to mess up your script! If it's not necessary to run it before tests or if you don't mind me adding it to the test & lint script, let me know.

makefile
r-ferrier: From what I understood from speaking with everyone, this is the minimal amount of data required right now to be able to properly work on any of the normal bug tickets that come up for us, but if there's a more minimal setup we can do let me know and I can change this! 

I could name this more explicitly to be something like `seed_GFSC_prod_db`?

makefile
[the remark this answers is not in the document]
ivan-kocienski-gfsc: that is good.

makefile
ivan-kocienski-gfsc: Sure, renaming it to `seed_GFSC_prod_db` is fine.

When we get seeds working (i.e. loading neighbourhood data) then the seed process will be complete. Cloning the production database can introduce problems if we become too dependent on the data in production for our code to work.

makefile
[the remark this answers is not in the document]
r-ferrier: ok cool, will take it out

PR #1752, opened 2023-03-22
use assert_match on two substrings instead of assert equal

### LINKED ISSUE #1751: [Bug]: Fix broken parser_base_test
## Description

One test fails on my machine: `read_http_source raises correct exception with invalid URL`
This fails because the error messages produced differ from one environment to another.

## User stories
- [x] As a Rosie I can run tests on my local machine and they all pass

## Implementation plan
- [x] Change test to look for presence of substrings `Failed to open TCP connection to dandilion.gfsc.studio:443` and `There was a socket error` instead of asserting a complete string match

r-ferrier:
Fixes #1751 

##Description

- Prevents test from failing in a different environment when the error message is subtly different by matching two substrings within the error message instead of looking for an exact match.
- Substrings looked for are `There was a socket error` and `Failed to open TCP connection to dandilion.gfsc.studio:443`
- These strings were chosen because they validate the principal of the test to check that the exception is correct and the URL is invalid

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ doc/**/*.js
 .yardoc
 ./tags
 doc/rdoc
+
+# Ignore IDE config files
+.vscode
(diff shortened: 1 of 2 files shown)


aaaaargZombies (approved):
LGTM

PR #1757, opened 2023-03-22
lockdown partnership tags

### LINKED ISSUE #1713: Lock down Partnership Tags
## User story

As a PlaceCal admin, I only want to be able to see my partnership tags and noone elses, in order that my interface doens't get hugely cluttered and we don't accidentally edit each other's stuff

## Acceptance criteria

- [x] Roots can see every Tag
- [x] Partnership admins can see their own Partnership Tag(s), plus all Categories and Facilities in all relevant dropdown menus
- [x] Everyone else can only see all Categories and Facilities in all relevant dropdown menus
- [x] Create another ticket for enforcing this in the backend - for now this can be a frontend-only change.

## Implementation plan

_initial plan_

- look into where tags can be accessed in the admin UI
- look to filter them depending on user
- add system tests to verify

aaaaargZombies:
fixes #1713

- Roots can see every Tag
- Partnership admins can see their own Partnership Tag(s), plus all Categories and Facilities in all relevant dropdown menus
- Everyone else can only see all Categories and Facilities in all relevant dropdown menus

I won't open the ticket to implement on the backed as this what we did here, it ended up being more efficient this way.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -30,8 +30,11 @@ class Tag < ApplicationRecord
   validate :check_editable_fields
 
   scope :users_tags, lambda { |user|
-    where id: user.tags_users.distinct.pluck(:tag_id)
-  }
+                       partnership_tags = Tag.where(type: 'Partnership', id: user.tags_users.distinct.pluck(:tag_id)).pluck(:id)
+                       other_tags = Tag.where("type != 'Partnership'").pluck(:id)
+
+                       Tag.where(id: partnership_tags + other_tags)
+                     }
 
   def name_with_type
     s_type = type || 'Tag'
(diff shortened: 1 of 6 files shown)


r-ferrier:
I have another question! What happens if I am a PlaceCal admin AND a root user? The user story says I should only be able to see my own tags if I'm a placeCal admin, but in this approach, won't my root user status supersede this and show me all the tags anyway? (I may be misunderstanding the admin roles here?)

r-ferrier (approved):
Note for future selves: we're splitting this into two issues now
- the quick fix for georgie ( not seeing all tags as a placecal admin) 
- the longer term fix to placecal admins (seeing all their own tags)

This PR deals with the second one!

PR #1761, opened 2023-03-22
Only show client root users the sites they're an admin for

### LINKED ISSUE #1746: Show less sites on dashboard for root admins
## Acceptance Criteria

~~Root user does not see all sites in their dashboard~~
~~There are  "Create a partner" "create a calendar" links where the site list used to be~~

- [x] IF a Root user is a site admin they do not see all sites in their dashboard - only those they admin for
- [x] IF a Root user is NOT a site admin they see all sites in their dashboard 
- [x] A Root user will always see some sites in their dashboard - either those they are admin for, or all sites

## Plan

- [x] Update the site scope to ensure only root admins with no site admin status see all sites in their dashboard
- [x] Update tests to assert working as expected

## Implementation Notes

I've rewritten the acceptance criteria, following clarification at standup that the problem we're addressing here is that site admins have been given root user permission as a temporary fix, but we do not want them to have access to all sites - just their own. Regular root users still need access to everything.

r-ferrier:
Fixes #1746 

### Description

- Modifies the site scope to allow access to only the sites a root user is an admin for if they are an admin for any
- If the root user is an admin for none, allows access to all sites
- Adds tests to assert working as expected

### Some qs!

- We still have an 'Add new site' button visible for the site admin roots - is this functionality we want them to have, or should this be removed too?
- The original issue ticket asked for `"Create a partner"  and "create a calendar" links where the site list used to be` . After discussing as a team it was agreed that we didn't want to remove the site list completely for any root user, which I thought made this redundant. Im now wondering though which type of root user was this for the benefit of, and do we actually want this in addition to showing the site list?

### DIFF
--- a/app/policies/site_policy.rb
+++ b/app/policies/site_policy.rb
@@ -41,10 +41,10 @@ class SitePolicy < ApplicationPolicy
 
   class Scope < Scope
     def resolve
-      if user.root?
-        scope.all
-      elsif user.site_admin?
+      if user.site_admin?
         scope.where(site_admin: user)
+      elsif user.root?
+        scope.all
       else
         scope.none
       end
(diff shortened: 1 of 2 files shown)


aaaaargZombies (approved):
works well :+1:

PR #1762, opened 2023-03-22
make type mandatory on Tag

### LINKED ISSUE #1737: tag type is mandatory

aaaaargZombies:
fixes #1737

- add a validation on the model
- fix up our test data as a lot of the tags were being generated without a type

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -20,7 +20,7 @@ class Tag < ApplicationRecord
   has_many :article_tags, dependent: :destroy
   has_many :articles, through: :article_tags
 
-  validates :name, :slug, presence: true
+  validates :name, :slug, :type, presence: true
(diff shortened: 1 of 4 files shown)


r-ferrier:
The code you've written looks good to me but I have a question! Is there any possibility we could have any tags in production currently without a type? ie do we need to write a migration too to find and assign one to these?

aaaaargZombies:
ok a quick check on prod and I think we're good

```sh
irb(main):002:0> Tag.where(type:nil).count
D, [2023-03-23T10:05:08.612164 #11] DEBUG -- :    (0.9ms)  SELECT COUNT(*) FROM "tags" WHERE "tags"."type" IS NULL
=> 0
```

r-ferrier (approved):
LGTM!

PR #1764, opened 2023-03-23
limit tags shown to root user if they are also a (Partnership)tag_admin

### LINKED ISSUE #1763: limit tags shown to root user if they are also a  (Partnership)tag_admin
## User story

building on  #1713 and #1757

As a PlaceCal admin, I only want to be able to see my partnership tags and noone elses, in order that my interface doens't get hugely cluttered and we don't accidentally edit each other's stuff

The origional describes what we want the admin permissions to be but we are currently giving admins root privilege so we may want to temporarily limit the amount of information certain root users see.

## Acceptance criteria

- [x] Root users who have the role Partnership admins can see their own Partnership Tag(s), plus all Categories and Facilities in all relevant dropdown menus

## Implementation plan

[modify this if statement to account for extra role](https://github.com/geeksforsocialchange/PlaceCal/blob/01c23833143ca985f23aed2ee73ec7ccbd34a950/app/models/tag.rb#L33)

aaaaargZombies:
fixes #1763 

- remove root user logic from policy scope
- exclude root user who is tag admin from receiving `Tag.all`
- add a tests to validate

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -30,7 +30,7 @@ class Tag < ApplicationRecord
   validate :check_editable_fields
 
   scope :users_tags, lambda { |user|
-                       return Tag.all if user.role == 'root'
+                       return Tag.all if user.role == 'root' && !user.tag_admin?
 
                        partnership_tags = Tag.where(type: 'Partnership', id: user.tags_users.distinct.pluck(:tag_id)).pluck(:id)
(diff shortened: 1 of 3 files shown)


r-ferrier (approved):
Works for me! Follow on tidy up task is now added here: https://github.com/geeksforsocialchange/PlaceCal/issues/1765

PR #1767, opened 2023-03-23
Allow partner admins to add & remove categories and facilities to partners from tags page

### LINKED ISSUE #1653: Partners can self assign themselves to categories and facilities (and not partnerships)
## User story

As a PlaceCal user managing a partner, I need to let people know what category of organisation the partner is in order that people can find the partner easily in a crowded listings service.

As a PlaceCal user managing a partner, I need to let people know what facilities the partner has in order that people can find the partner easily in a crowded listings service.

## Acceptance criteria

- [ ] #1766
- [x] I can unassign a service category from any partner I manage
- [x] Partners I manage show up in category results (in the tags page)
- [x] I can assign any number of facilities to any partner I manage
- [x] I can unassign any facilities from any partner I manage
- [x] Partners I manage show up in facility results (in the tags page)
- [x] If I edit a tag to assign it or unassign it to my partner, other partners I am not managing are not removed from this tag

## Implementation plan

- [ ] limit number of category tags to 3
- [x] make tagged partners visible to users with citizen roles in tags tab

r-ferrier:
Fixes #1653

### Description

- Updates the tags policy to make `:partner_id` a permitted attribute for a `:partner_admin`. This allows them to add or remove a partner that they manage from a category or facility tag.
- Adds tests to assert that the tags policy is working correctly for a `:tag_admin` or `:partner_admin` user
- Removed a redundant line of code from the tag policy - note on which below.

### Some Thoughts/ problems

- The two facility tags in our production copy db are currently also `system_tags` which means they cannot be edited from the tag dashboard, and so also cannot be assigned/removed from a partner from this interface. This feels like a subtly different bug - if I create a new `facility_tag` without a `system_tag` symbol then it works as expected. Do we want `partner_admins` to be able to attach/detach tags that are also `system_tags`? Or is this a weird error and do we want to remove this ability from them in the partner interface too? If whoever is reviewing this wants me to talk them through it let me know - it's confusing to read/write!
- The acceptance criteria for the user story wanted "up to 3" category tags to be able to be assigned but there was nothing in the user story mentioning this. I couldn't tell if this was a necessary limitation or not so have opened as a new issue.

### DIFF
--- a/app/controllers/admin/tags_controller.rb
+++ b/app/controllers/admin/tags_controller.rb
@@ -61,6 +61,12 @@ module Admin
     def update
       authorize @tag
       attributes = permitted_attributes(Tag.new)
+
+      if current_user.partner_admin?
+        attributes[:partner_ids] =
+          helpers.all_partners_for(@tag, attributes)
+      end
+
       if @tag.update(attributes)
         flash[:success] = 'Tag was saved successfully'
         redirect_to admin_tags_path
--- a/app/helpers/tags_helper.rb
+++ b/app/helpers/tags_helper.rb
@@ -13,4 +13,13 @@ module TagsHelper
   def show_assigned_user_field_for(form)
     [Tag, Partnership].include?(form.object.class)
   end
+
+  # prevent invisible partner ids from being overwritten when updating a tag
+  def all_partners_for(tag, attributes)
+    editable_partners = current_user.partners.pluck(:id)
+    uneditable_partners = tag.partner_ids - editable_partners
+    updated_partners = editable_partners & attributes[:partner_ids].reject(&:empty?).map(&:to_i)
+
+    uneditable_partners + updated_partners
+  end
 end
--- a/app/policies/tag_policy.rb
+++ b/app/policies/tag_policy.rb
@@ -23,9 +23,6 @@ class TagPolicy < ApplicationPolicy
     # system tags can only be edited by root
     return false if @record.system_tag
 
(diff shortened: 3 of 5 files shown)


ivan-kocienski-gfsc:
Question: if a partner admin is on an edit tag page and they can edit the partner the tag is assigned to does this mean they can change any partner? like if i was logged in i could remove other peoples partners from a tag?

r-ferrier:
> Question: if a partner admin is on an edit tag page and they can edit the partner the tag is assigned to does this mean they can change any partner? like if i was logged in i could remove other peoples partners from a tag?

No! You can only see/add/remove those partners which you are admin for.

r-ferrier:
> Hmmm interesting. How does the back-end support that? I thought when you are on a Tag page and POST to the update action it will apply the `partner_ids` field exclusively: any IDs in that set are added, any missing from that set are removed. So if a Partner admin can only see their partners for a given tag when that tag is saved it clears any IDs missing, ergo the Partner Admin just de-tagged any partners they don't admin.

ugh I was not aware this is how tags function! I will add some more acceptance criteria to this issue and have a think about how we're testing this.

r-ferrier:
I've fixed the problem Ivan found and added some more tests to make sure this works as it should! In doing so I have discovered tag_admins don't seem to be working in the way we would expect, but will make a separate issue for this.

aaaaargZombies (approved):
everything works  :confetti_ball:

PR #1784, opened 2023-03-31
Use same version of node locally as on prod & remove broken dependency

### LINKED ISSUE #1783: Production build environment differs from staging
## Description
We had an issue releasing to production because of a mismatch in node versions.

## Motivation

Our build environments need to match in order that we can be certain that the released code is identical, so that we can rely on testing in development and staging environments.

It appears that for the testing part of release process, it is using nodenv & the expected version but when it comes to the deploy / dokku release, it is not.

## Mitigated during 0.14.2

- Rolled back expected node version v16.13.1
- Rolled back lint-staged version
- Recreated yarn lock
- Pushed to staging
- Pushed to production

## Consider fixing
- Make sure all environments use the node version configs
- Turn off renovate
- Update INSTALL.md so it does not refer to stuff we no longer use (like travis)
- Think about why dev deps would be installed at all during production build process

r-ferrier:
Fixes #1783 

Prod is currently using node 16.13.1. We're unable to quickly work out why this is, so as a temporary fix, we're going to roll back node locally so that it matches prod, and roll back an update made by renovate that causes the build to break when using this node version.


katjam (approved):
Yes. Let's try this.

PR #1816, opened 2023-04-27
Partners cannot have more than 3 category tags, and have proper tag relationships

### LINKED ISSUE #1766: Limit the amount of categories possible to tag a partner with to 3
This has been generated by one of the acceptance criteria from #1653:

>  I can assign up to 3 categories to any partner I manage

aaaaargZombies:
fixes #1766

## Description

- add validation to partners model
- add tests for new validation


app/models/partner.rb
```diff
@@ -360,4 +362,11 @@ def opening_times_is_json_or_nil
 
     errors.add :base, 'Partner.opening_times must be valid json'
   end
+
+  def three_or_less_category_tags
+    number_of_category_tags = tags.count { |t| t.type == 'Category' }
+    return if number_of_category_tags < 4
+
+    errors.add :base, 'Partner.tags can contain a maximum of 3 Category tags'
```
katjam, line 370: Should we maybe also add some hint text in the UI so it's not a surprise? Or is that not precedent in this interface... like if it's not trivial to add or if it makes it look muddled, don't worry.

aaaaargZombies: It would probably be easy to add a little counter with stimulus to say something like "x/3 categories tags selected".

Is that the sort of thing you're thinking, or just like a paragraph next to the inputs warning that there's a limit?

test/models/partner_test.rb
```diff
@@ -187,4 +187,30 @@ class PartnerTest < ActiveSupport::TestCase
     problems = partner.errors[:base]
     assert_equal 'Partners must have at least one of service area or address', problems.first
   end
+
+  test 'partner can have up to 3 "Category" tags' do
+    partner = create(:partner)
+    assert_predicate partner, :valid?
+
+    partner.tags << create(:tag, type: 'Category', name: 'Category Tag 1')
```
kimadactyl, line 195: just an fyi this is a rare case you can use ruby `time` which you learn in the first day you learn ruby then never get to touch again lol https://www.dotnetperls.com/iterator-ruby

app/models/partner.rb
kimadactyl: feel like this is a good idea but do we want to do too much UX work until the tag input is split into partnerships, categories and facilities tags? what we have now feels like a temp holding solution. if its holding us back i think we should prob do that first?

app/models/partner.rb
```diff
@@ -360,4 +362,11 @@ def opening_times_is_json_or_nil
 
     errors.add :base, 'Partner.opening_times must be valid json'
   end
+
+  def three_or_less_category_tags
+    number_of_category_tags = tags.count { |t| t.type == 'Category' }
```
kimadactyl, line 367: this should just be `categories.count` if STI is set up properly!

aaaaargZombies: I'm sorry I don't understand this, searching STI has not helped either...

app/models/partner.rb
aaaaargZombies: Screenshot 2023-05-01 at 10-53-36 Editing African Caribbean Care Group PlaceCal Admin

Does this make it clear enough?

kimadactyl:
Just a note when you do your refactor that this can be simplified too now - im actually not sure what this is doing? 

https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/models/tag.rb#L35

app/models/partner.rb
katjam: Single Table Inheritance (in case anyone else is looking)

aaaaargZombies:
> Just a note when you do your refactor that this can be simplified too now - im actually not sure what this is doing?
>
> https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/models/tag.rb#L35

this is about permissions for access to partnership/site tags. feels a bit out of scope for this issue but I can continue on it tomorrow if it's important. I think  I'd rather make a new issue for it with some acceptance criteria because it feels like we're beginning to pull at a thread and I don't know how long it is!

kimadactyl:
> > Just a note when you do your refactor that this can be simplified too now - im actually not sure what this is doing?
> > https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/models/tag.rb#L35
> 
> this is about permissions for access to partnership/site tags. feels a bit out of scope for this issue but I can continue on it tomorrow if it's important. I think I'd rather make a new issue for it with some acceptance criteria because it feels like we're beginning to pull at a thread and I don't know how long it is!

It's not important no - to be clear all I meant is you can replace

```ruby
Tag.where(type: 'Partnership' ...)
```

With

```ruby
Partnership.where(...)
```

If you want :) nbd if not tho!

test/models/partner_test.rb
ivan-kocienski-gfsc: i used `x.times do` just today in some tests.

PR #1821, opened 2023-05-01
Tag filter in url

### LINKED ISSUE #1779: URL route contains partner category and options if selected
## User story

As someone wanting to link a set of filtered partners, I want the URL to be pastable to someone else in order that we both see the same thing when the link is shared

## Acceptance criteria

- [ ] URL bar has params in it when partner filter is used

## Implementation plan

Having reviewed the code I have the feeling this could do with a bit of a tidy up
- moved filter code into a concern (the same pattern as other filters on the site)
- moved form into partial
- changed form over to a GET
- moved filter logic from /partners/tag into filter
- removed unused action
- put in unit tests for filter class
- put in integration tests covering the UX

ivan-kocienski-gfsc:
Closes #1779 

Changed filter form from a POST to a GET. Tidied code into controller concern (with unit test). Some integration tests on the route/view level.


app/assets/stylesheets/partner.scss
```diff
@@ -1,3 +1,8 @@
 .map-marker {
 	height: 300px !important;
 }
+
+.breadcrumb .size-patch {
```
aaaaargZombies, line 5: should probably be located with the rest of the breadcrumb styles, I thought I'd fixed the uneven buttons sorry if not!

https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/components/breadcrumb/breadcrumb.sass

config/routes.rb
```diff
@@ -72,7 +72,7 @@
   get '/partners/:id/events/:year/:month/:day' => 'partners#show', constraints: ymd
   get '/places' => 'partners#index' # Removing separate Places view for now.
   get '/partners/:id/embed' => 'places#embed'
-  post '/partners/tag' => 'partners#tag'
+  # post '/partners/tag' => 'partners#tag'
```
aaaaargZombies, line 75: this could be deleted

aaaaargZombies: If we're not using the turbo-stream thing any more lets also delete this template

https://github.com/geeksforsocialchange/PlaceCal/blob/46b3db85c5f4ed399ffd194581d3a13fa6af816f/app/views/partners/tag.turbo_stream.erb

aaaaargZombies (changes requested):
I like this more than what I did, maybe I just didn't understand the turbo-stream thing enough but it seemed to break a lot of standard browser behavior. Just a few change requests but nothing major.

app/assets/stylesheets/partner.scss
ivan-kocienski-gfsc: That's a good idea but this is not a global fix for the entire site and is defined in a class that is only used in this one form.

Unless this filter pattern is going to be used site-wide?

aaaaargZombies: ok, you've convinced me, I think mark will redesign this at some point so we should just remember to remove it then.

ivan-kocienski-gfsc:
Ah i see it now. i only have a tiny screen on my laptop

aaaaargZombies:
yes that caught me out when I first implemented it just because I had the dev tools open when working

PR #1827, opened 2023-05-03
Fix Calendar Related Problems for Partners

### LINKED ISSUE #1790: [Bug]: Flourish calendar coming up as "bad source"
### Was
Flourish seems to be having problems: under the import status for their linked eventbrite calendar, PlaceCal back end says "bad_source", but everything looks right in terms of the calendar link? https://www.eventbrite.co.uk/o/flourish-cic-12424608687 (edited)

- [ ] Follow up with request in discord from 03/04/23

### LINKED ISSUE #1802: [Bug]: flourish events showing duplicates
## Description

What were you doing?

From Discord

> Hiya, Fourish CIC events which are pulled from Eventbrite are put 3 times onto place cal

## Steps to reproduce

1. Visit https://flourish.placecal.org/events

## Platform (if relevant)

r-ferrier:
fixes #1790 
fixes #1802 

#### #1802  
- Within event sorting, forces uniqueness within a list of events to remove duplicates

#### #1790
- Allows users to re-queue bad_source calendars for import as it looks like lots are being marked as bad source when an import fails once, and then a re-import is never attempted or allowed
- Provides some better information to the user as to why their calendar import may be failing

I think bad_source bug has been introduced with PR #1652 & from reading this would seem like only calendars that are specific types of failure should be marked as such, so it would be great to double check with you especially @ivan-kocienski-gfsc that I'm not going to introduce a new problem here by adding this ability to try re-importing them! 

I also thought about digging into why so many had been marked as bad_source when after re-running the workers only one returned as such, but thought we could save that can of worms for another time.


app/models/calendar.rb
```diff
@@ -44,10 +44,6 @@ class Calendar < ApplicationRecord
     default: :idle
   )
 
-  scope :where_busy, -> { where(calendar_state: %i[in_queue in_worker]) }
-  scope :where_idle, -> { where(calendar_state: :idle) }
-  scope :where_errored, -> { where(calendar_state: :error) }
-
```
r-ferrier, line 50: I also removed these scopes as they were unused anywhere

app/views/admin/calendars/_import.html.erb
```diff
@@ -3,7 +3,11 @@
 
 <%= render 'status' %>
 
-<% if @calendar.calendar_state.idle? %>
+<% if @calendar.calendar_state.idle? || @calendar.calendar_state.bad_source? %>
```
r-ferrier, line 6: This is the only bit I'm unsure about - we don't do this for errored calendars but we do allow them to be queued for import - I just wasn't sure what the associated risks were with letting someone re-run this. Especially if they've changed the URL, for example, it feels like we should let them rerun?

app/views/admin/calendars/_import.html.erb
```diff
@@ -3,7 +3,11 @@
...
+  <% if @calendar.calendar_state.bad_source? %>
+    <p class="bigger"> We're unable to read this calendar - the URL may be broken or there may be some invalid data in it.<br/>Please check the calendar is set up correctly and try again.<br/>
+    </p>
```
r-ferrier, line 9: Also this copy may be completely wrong - was trying to phrase this in a non-tech friendly way - but let me know if there's a better turn of phrase

app/views/admin/calendars/_import.html.erb
ivan-kocienski-gfsc: If they change the URL the calendar should be forced into the queue again (?) (or should be)

r-ferrier: But in this bug, the URL is fine & their calendar is listed as a bad_source despite being fine and they can't do anything about it

ivan-kocienski-gfsc:
If I remember the underlying problem is that some sources (eventbrite) had intermittent problems and would come back online after a bit. Originally the 'bad_source' was supposed to flag things that have gone forever.

You are right in that we should just let people reset the bad_source calendars so they can resolve this issue on their own.

app/views/admin/calendars/_import.html.erb
ivan-kocienski-gfsc: This seems okay. As a non-blocking suggestion it may be an idea to move this logic out of the view and into the model as a `can_be_requeued?` method so it can be unit-tested.

ivan-kocienski-gfsc (approved):
Seems okay to me

app/controllers/application_controller.rb
```diff
@@ -76,7 +76,7 @@ def sort_events(events, sort)
     if sort == 'summary'
       [[Time.now, events.sort_by_summary]]
     else
-      events.sort_by_time.group_by_day(&:dtstart)
+      events.distinct.sort_by_time.group_by_day(&:dtstart)
```
ivan-kocienski-gfsc, line 79: (non blocking) at some point we need to clean a lot of this type of thing out of the application controller

app/views/admin/calendars/_import.html.erb
r-ferrier: I've moved the logic into the model and refactored one other line to make it clearer what is being checked (whether a calendar is busy or not). I'm not going to write a unit test for this at this point because it would be super closely coupled to this specific method, but it's clearer from its name what it should be doing which is very helpful!

PR #1848, opened 2023-05-09
Database cleaner

### LINKED ISSUE #1847: Install database_cleaner gem
## Description

Install database cleaner gem

## Motivation

Reset database between tests

ivan-kocienski-gfsc:
Closes #1847

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,15 +118,15 @@ group :development do
 end
 
 group :test do
+  # gem 'simplecov', require: false
   gem 'capybara-select-2'
+  gem 'database_cleaner-active_record'
   gem 'graphql-client'
   gem 'json_matchers'
   gem 'minitest-rails-capybara'
   gem 'minitest-reporters'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -126,6 +126,7 @@ group :test do
   gem 'rails-controller-testing'
   gem 'selenium-webdriver'
   # gem 'simplecov', require: false
+  gem 'database_cleaner-active_record'
```
kimadactyl, line 129: non blocking - these are meant to be alphebatized is rubocop not working?

ivan-kocienski-gfsc:
This won't fix that problem. That problem is to do with the fixtures, that get rebuilt before every test is run. I just took out the line in the partner edit controller test that i just fixed and it's failing in exactly the way it would fail if there were two neighbourhoods with the same unit_codes.

r-ferrier (approved):
Niiiice!

PR #1863, opened 2023-05-12
contstrain text width on homepage

### LINKED ISSUE #1859: Apply max width constraints consistently to new components
## Issue

The new design system has a max-width constraint applied to the components which we've applied a little inconsistently. Once the latest batch of components have been merged it might be a simple win & would make Mark very happy if we go through them and make sure they all have their inside content constrained to the same max-width (I think it's 860px but don't quote me on that! All the information is written in the task)

https://xd.adobe.com/view/d737e480-ff3a-4e07-8d32-339f902ddf4b-8264/grid/  

## Acceptance criteria

- [ ] All of the text on the home page lines up at the edges (or does not go over the edges) of the same central column at each breakpoint

## Implementation plan

aaaaargZombies:
fixes #1859 

- created a wrapper component to we can use when building other components, this should help us keep consistent and allow us to make changes across the system if needed.
- I've applied it to the block quote even though it's not a new component, this makes the text line up better but makes the quote marks a little odd. I'm happy to remove at reviewers request!

### DIFF
--- a/app/components/full_width_action_component/full_width_action_component.html.erb
+++ b/app/components/full_width_action_component/full_width_action_component.html.erb
@@ -1,21 +1,21 @@
 <section class="full_width_action <%= @color_class[@color] %>">
-  <h3 class="full_width_action__title">
-    <%= @title %>
-  </h3>
-  <p class="full_width_action__content">
-    <%= content %>
-  </p>
-
-  <div class="full_width_action__btn">
-    <% if @color == "cream" %>
-      <%= render(LinkBtnLrgGreenComponent.new( link_url: @link_url)) do %>
-        <%= @link_text %>
+  <%= render(MaxWidthComponent.new()) do %>
+    <h3 class="full_width_action__title">
+      <%= @title %>
+    </h3>
+    <p class="full_width_action__content">
+      <%= content %>
+    </p>
+    <div class="full_width_action__btn">
+      <% if @color == "cream" %>
+        <%= render(LinkBtnLrgGreenComponent.new( link_url: @link_url)) do %>
+          <%= @link_text %>
+        <% end %>
+      <% else %>
+        <%= render(LinkBtnLrgLightComponent.new( link_url: @link_url)) do %>
+          <%= @link_text %>
+        <% end %>
       <% end %>
-    <% else %>
-      <%= render(LinkBtnLrgLightComponent.new( link_url: @link_url)) do %>
-        <%= @link_text %>
-      <% end %>
-    <% end %>
-  </div>
-
+    </div>
+  <% end %>
 </section>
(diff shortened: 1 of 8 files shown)


app/components/full_width_action_component/full_width_action_component.html.erb
```diff
@@ -1,21 +1,21 @@
...
-  </h3>
-  <p class="full_width_action__content">
-    <%= content %>
-  </p>
-
-  <div class="full_width_action__btn">
-    <% if @color == "cream" %>
-      <%= render(LinkBtnLrgGreenComponent.new( link_url: @link_url)) do %>
-        <%= @link_text %>
+  <%= render(MaxWidthComponent.new()) do %>
```
r-ferrier, line 2: I think having this as a component is really handy for instances like this but I'm wondering if there are also occasions where we might want to just use the class instead, for example when you want to apply the max-width to the parent container & don't want to have to add another div for that - could we maybe store the css in a shared file instead of at component level?

r-ferrier (approved):
Looks good! Just one question & one suggested change, although not a deal breaker, so approved.

I notice you haven't applied this to the neighbourhood cards - is this cos these sit outside this max width, and if so, do they need to have their own max width, or are they alright just filling the area aided by padding etc?

I wouldn't worry about the blockquote looking weird - I've refactored it into a component in my PR anyway so once this is in I can apply this styling to it as well.

app/components/full_width_action_component/full_width_action_component.html.erb
aaaaargZombies: I guess we're getting into the realm of global utility classes instead of BEM then? Maybe that's fine.

I think I need a more concrete example about the parent container thing your describing because I'm not picturing it. I was sort of thinking we'd have a go to thing that performs one way and if we needed to step out of that we'd use our components own styles.

aaaaargZombies:
> I notice you haven't applied this to the neighbourhood cards - is this cos these sit outside this max width, and if so, do they need to have their own max width, or are they alright just filling the area aided by padding etc?

I don't think it's the same width, it's not in XD so I just eyeballed it.  I also took the issue quite literally and just applied it to text.

app/components/full_width_action_component/full_width_action_component.html.erb
r-ferrier: I'm thinking about situations like the block quote I guess, where you might just want to apply the max-width to the <p> tag instead of needing to wrap it in an extra <div> with a new component. Is it bad design wise to make use of global utilities in addition to BEM? I feel like it would be neater here but I can also see why you'd be hesitant to do it

aaaaargZombies: I think I'm going to merge because you've not blocked it but I take your point that this component is essentially a very restrictive way of applying a global utility class.

I'm open to it being refactored out / modified in some way if it proves to not be useful in the future.

PR #1864, opened 2023-05-12
add [cCc] to filler content

### LINKED ISSUE #1858: All non-approved copy contains [cCc]
## Issue

We need to make sure no non-approved copy gets into the new release. Someone could do with doing a quick sweep of the home page and making sure it's very clear what is or isn't approved, maybe adding in [cCc] to all the developer/designer suggested copy or anything old and not yet reviewed?

## Acceptance criteria

- [ ] TBC

## Implementation plan

aaaaargZombies:
fixes #1858 

I'm a bit weary of this issue because I think it might make extra work if something existing can be approved we won't have to undo it. 

That being said - here it is. I hope it's useful.

### DIFF
--- a/app/components/home_strapline_component/home_strapline_component.html.erb
+++ b/app/components/home_strapline_component/home_strapline_component.html.erb
@@ -1,6 +1,6 @@
 <div class="home__strapline">
   <p class="home__strapline-quote">
-    PlaceCal is an online resource for community-led events and activities, owned, curated and operated by a real-world network of community organisers.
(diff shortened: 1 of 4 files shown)


r-ferrier (approved):
I think it's a good idea - would have had no idea that these were all designer/dev suggested otherwise. A find and replace in my software is super quick so very happy to sit down with someone more content minded at the end of the sprint and just run through any leftovers. In fact, I will make a ticket for just this so we remember to do the tidy up afterwards.

PR #1875, opened 2023-05-16
change how sites are dynamically pulled in

### LINKED ISSUE #1873: [Bug]:  Partnerships are hard coded - use is_published data instead
## Description

Use data in the same way as we do for neighbourhood instances

## Acceptance Criteria

- [x] The image is pulled from the logo
- [x] The summary is pulled from the tag line
- [x] The button says "[Partnership name] calendar" and goes to the events page

r-ferrier:
Fixes #1873

## Description

- Assigns published neighbourhood and partnership cards dynamically
- Tries to make styling a bit more uniform, but will need to be viewed with some better data to double check this is working - very confused about what exactly I'm looking at in local host as I thought I was were pulling this in from production but the published instances are not the ones we have listed - we definitely need to check this is all correct
- This was a bit of a mess and we're pushed for time so I've done what i can but am aware we could do more here! Think it just needs to be put to one side for a bit while we finalise all the content

### DIFF
--- a/app/components/featured_partnerships_component.rb
+++ b/app/components/featured_partnerships_component.rb
@@ -1,4 +1,8 @@
 # frozen_string_literal: true
 
 class FeaturedPartnershipsComponent < ViewComponent::Base
+  def initialize(sites:)
+    super
+    @sites = sites
+  end
 end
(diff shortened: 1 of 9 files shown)


ivan-kocienski-gfsc (approved):
I think this is okay to go.

PR #1885, opened 2023-05-17
Make emails the same

### LINKED ISSUE #1882: [Bug]: Email addresses not the same
## Description

Screenshot 2023-05-17 at 10 40 17

the email in join us is to `info@placecal.org` but in email us is `kim@placecal.org` - is that right?

r-ferrier:
Fixes #1882 

### Description

- Makes email addresses the same in the footer & get in touch form on the home site


katjam (approved):
As long as we know info@placecal exists.

PR #1886, opened 2023-05-17
Pin background image width

### LINKED ISSUE #1883: [Bug]: Background image different sizes on different pages
## Description

The background image size must be dependent on page length - needs the height fixing so it doesn't jump around from page to page

r-ferrier:
Fixes #1883 

### Description

- Sets fixed width to background image to stop it from changing size on pages of different lengths


katjam (approved):
Good spot

PR #1898, opened 2023-05-17
Only assign neighbourhoods to neighbourhoods row

### LINKED ISSUE #1897: [Bug]: Neighbourhoods cards also pulling in partnerships
## Description

The neighbourhoods cards are also pulling in partnerships from the site

r-ferrier:
Fixes #1897 

## Description

- Stops neighbourhoods pulling in all sites

### DIFF
--- a/app/components/case_study_component/case_study_component.html.erb
+++ b/app/components/case_study_component/case_study_component.html.erb
@@ -1,6 +1,7 @@
 
 <section class="case_study__section--outer">
   <section class="case_study__section--inner case_study__round-border--top">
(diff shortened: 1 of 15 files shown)


aaaaargZombies (approved):
Looks good, could this be applied to the homepage too. I saw on the other PR that you'd limited by 4 and that has the same affect but I think that might be by chance due to the low IDs of the published neighborhood sites

r-ferrier:
> 

Oh good point - I hadn't even thought about the fact this will also be happening there - will do

PR #1910, opened 2023-05-22
Refinement of filter category selection

### LINKED ISSUE #1909: If there are no partners with that tag, it is not shown as an option in the filter list
## Description

Regression that has reinstated the ability to filter on categories not used by the partner - suspected cause is https://github.com/geeksforsocialchange/PlaceCal/pull/1821/files#diff-94283ee9fa75fb5564c7731d2d787c3a4046255ffeca480bfd9ab3b47d6b4167L30

ivan-kocienski-gfsc:
Fixes #1909

On site partners page: the filter only shows categories that have been applied to partners who have an address or service area in the site's neighbourhood pool.

### DIFF
--- a/app/controllers/concerns/partner_category_filter.rb
+++ b/app/controllers/concerns/partner_category_filter.rb
@@ -42,6 +42,7 @@ class PartnerCategoryFilter
                     )
                     .group('tags.id')
                     .having('count(partner_tags.tag_id) > 0')
+                    .having('(count(addresses.id) > 0 OR count(service_areas.id) > 0)')
                     .order(:name)
   end
 
--- a/app/controllers/concerns/partner_category_filter.rb
+++ b/app/controllers/concerns/partner_category_filter.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class PartnerCategoryFilter
-  def initialize(params)
+  def initialize(site, params)
     @current_category_id = params[:category]
     @current_category = Category.where(id: @current_category_id).first if @current_category_id.present?
 
     @include_mode = params[:mode] # include/exclude
+    @site_neighbourhood_ids = site.neighbourhood_ids
   end
 
   def active?
@@ -27,10 +28,20 @@ class PartnerCategoryFilter
   end
 
   def categories
+    # only return categories that have partners
+    #   where those partners have addresses or service areas
+    #     in the neighbourhoods of the site
+
     @categories ||= Category
                     .joins(:partner_tags)
+                    .left_joins(:partner_tags)
+                    .left_joins(partners: %i[address service_areas])
+                    .where(
+                      '(addresses.neighbourhood_id in (:neighbourhood_ids) OR service_areas.neighbourhood_id in (:neighbourhood_ids))',
+                      neighbourhood_ids: @site_neighbourhood_ids
+                    )
                     .group('tags.id')
-                    .having('count(tag_id) > 0')
+                    .having('count(partner_tags.tag_id) > 0')
                     .order(:name)
   end
 
--- a/app/models/neighbourhood.rb
+++ b/app/models/neighbourhood.rb
@@ -20,6 +20,8 @@ class Neighbourhood < ApplicationRecord
            source: :partners,
            class_name: 'Partner'
 
+  # validates :unit_code_value, presence: true, uniqueness: true
+
   # validates :name, presence: true
   validates :unit_code_value,
             length: { is: 9 },
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
 		"lodash": "4.17.21",
 		"popper.js": "1.16.1",
 		"sass": "1.62.1",
-		"select2": "4.0.13"
+		"select2": "4.0.13",
+		"yarn": "^1.22.19"
 	},
 	"devDependencies": {
 		"husky": "8.0.3",
--- a/package.json
+++ b/package.json
@@ -16,13 +16,13 @@
 		"lodash": "4.17.21",
 		"popper.js": "1.16.1",
 		"sass": "1.62.1",
-		"select2": "4.0.13",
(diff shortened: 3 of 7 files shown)


ivan-kocienski-gfsc:
Well I was writing this but then realised i had made a mistake

> Essentially all this is doing is bringing the partner category filter in-line with how the `Partner.for_site` scope works by joining the category-count-partners query with a "only find partners that have an address or a service area in the pool of site neighbourhoods". Note we don't need to bother counting addresses or service areas as we are doing a `LEFT JOIN` and not a `LEFT OUTER JOIN` so partners must have either an address or a service area.

Problem is the query does a `LEFT JOIN` from partners to addresses/service_areas which is an AND so `PARTNER AND ADDRESS` and `PARTNER AND SERVICE AREA` which means a partner must have BOTH a address and a service area.

Oops

app/controllers/concerns/partner_category_filter.rb
```diff
@@ -27,10 +28,21 @@ def apply_to(query = Partner)
   end
 
   def categories
+    # only return categories that have partners
```
katjam, line 31: Is this a rails convention to progressively indent multi-line comments? I've not seen that before.

app/controllers/concerns/partner_category_filter.rb
```diff
@@ -27,10 +28,21 @@ def apply_to(query = Partner)
...
+                    .left_joins(:partner_tags)
+                    .left_joins(partners: %i[address service_areas])
+                    .where(
+                      '(addresses.neighbourhood_id in (:neighbourhood_ids) OR service_areas.neighbourhood_id in (:neighbourhood_ids))',
+                      neighbourhood_ids: @site_neighbourhood_ids
+                    )
                     .group('tags.id')
-                    .having('count(tag_id) > 0')
+                    .having('count(partner_tags.tag_id) > 0')
+                    .having('(count(addresses.id) > 0 OR count(service_areas.id) > 0)')
```
katjam, line 45: Is this necessary? Like how would they have been included if they didn't have those values? Does the in on line 40 include null matches?

app/models/neighbourhood.rb
```diff
@@ -20,6 +20,8 @@ class Neighbourhood < ApplicationRecord
            source: :partners,
            class_name: 'Partner'
 
+  # validates :unit_code_value, presence: true, uniqueness: true
```
katjam, line 23: should this have replaced the line below?

package.json
```diff
@@ -16,7 +16,8 @@
 		"lodash": "4.17.21",
 		"popper.js": "1.16.1",
 		"sass": "1.62.1",
-		"select2": "4.0.13"
+		"select2": "4.0.13",
+		"yarn": "^1.22.19"
```
katjam, line 20: I think if anything this is a dev dependency? If it wasn't here before it might just be a prerequisite?

test/integration/partner_category_filter_test.rb
```diff
@@ -2,8 +2,8 @@
 
 require 'test_helper'
 
-class PartnerCategoryFilterTest < ActionDispatch::IntegrationTest
-  setup do
+class PartnerCategoryFilterIntegrationTest < ActionDispatch::IntegrationTest
+  def setup
```
katjam, line 6: I'm not sure which is correct but in test/controllers/concerns/partner_category_filter_test.rb we use:

`setup do`

Would be good to stick with one pattern.

test/controllers/partners_controller_test.rb
```diff
@@ -16,6 +16,7 @@ class PartnersControllerTest < ActionDispatch::IntegrationTest
       a.save
       a
     end
+
```
katjam, line 19: Nice new empty line!

app/controllers/concerns/partner_category_filter.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class PartnerCategoryFilter
-  def initialize(params)
+  def initialize(site, params)
```
katjam, line 4: Not sure what the common pattern is but if we only need the neighbourhood ids, would we pass those rather than the whole site?

katjam (commented):
Nothing looks showstopping here but I've asked a few questions. In the interest of time, I'll see what @r-ferrier thinks in terms of any of my questions being an issue.

app/controllers/concerns/partner_category_filter.rb
ivan-kocienski-gfsc: nope. just a me thing i am afraid

app/controllers/concerns/partner_category_filter.rb
ivan-kocienski-gfsc: i think you are confusing `left joins` with `left outer joins`. `left joins` must have both but `left outer joins` can have one if present. we need to do it this way as partners can have EITHER a service area OR an address.

app/models/neighbourhood.rb
ivan-kocienski-gfsc: OH LORD NO! if you put this line in the test suite goes into FULL FIRE mode. and then we'd have to go through the entire test suite and make it so our factories produce valid models.

this would be one of those "our tests could be better" but "this would take a lot of work" examples.

but it SHOULD be a constraint.

r-ferrier:
If it's working then I'm happy to see it go in!

package.json
ivan-kocienski-gfsc: I can move it into the dev dependencies (?)

app/controllers/concerns/partner_category_filter.rb
ivan-kocienski-gfsc: doesn't passing in an array of integers sort of expose the internal workings of the class and break the abstraction?

katjam (approved):
Yep. Happy with this!

Issue #1478, opened 2022-10-05
Build a Normal Island fake neighbourhood for development


erbridge:
## User story

As an owner of PlaceCal I need to be able to roll PlaceCal out anywhere in the world so that I can make everyone's lives better.

As a developer of PlaceCal I need to not need to import the UK's geography and live data to work on the service so that I can work quickly within a smaller context.

## Acceptance criteria

- [ ] Local development happens on Normal Island, a fake geography with a fake geographic hierarchy
- [ ] A set of seed files I can understand by reading then in one sitting has everything I need to be able to have a blank slate geography for Normal Island to work with
- [ ] We seed at least one of every entity type, too
- [ ] We have tests to confirm the service behaves correctly for UK data
- [ ] The service still works with the UK's geography on staging and production
- [ ] There is an illustration of Normal Island we can refer to clearly for development
- [ ] The makefile and readme have been updated to use seed files and not prod db copy

## Follow up work

We then want to seed more interesting partners and all the rest to give us a "lived in" development environment.

## Implementation notes & questions

A side effect of this (and secretly the goal of it) is that we completely decouple the UK's specifics from PlaceCal, so PlaceCal works seamlessly with any geography.

## Implementation plan

(derived from #1947)

- [ ] clear out old seeding code
  - db/seeds and any migration that populates / manipulates data should be blanked
- [ ] generate actual seed data (may happen concurrently with step below)
- [ ] write seeding script
  - [ ] should be easy to use
  - [ ] should reset database completely
- [ ] update documentation

kimadactyl:
Linked to (superceeds?) #1083 and #876

r-ferrier:
Linked to #1748 (completed)

ivan-kocienski-gfsc:
the biggest hurdle for this is that we currently assume the dataset is UK postcodes because we use a UK postcode lookup service to map addresses to neighbourhoods. 

Kim has spoken of Zoning (my word) where we allow for different datasets for different regions). This would allow us to switch postcode resolution service but it would also be a lot of work.

kimadactyl:
yeah agreed - it mgiht be good to open a ticket specificallty for "allow choosing address to neighbourhood resolution mapping service" or something like that?

we are prob also heading towards needing some kind of global config options

ivan-kocienski-gfsc:
Notes and feedback

> Local development happens on Normal Island, a fake geography with a fake geographic hierarchy

[...]

Issue #1661, opened 2023-01-30
Add seeds for tags


kimadactyl:
## User story

As a developer, I need to get seeded with a set of tags that make sense, in order that I can develop well

## Acceptance criteria

- [ ] criteria one
- [ ] criteria two

## Implementation plan

_To be written by the developer_

aaaaargZombies:
currently there's a task

```
rails placecal:load_pht_category_tags
```

kimadactyl:
yep - ticket needs some love, we also need some base seed partnership and facility tags now

aaaaargZombies:
## verified

seeds are running, all PHT tags are present

```
irb(main):003:0> Tag.all.each{|t|puts t.name }
  Tag Load (1.3ms)  SELECT "tags".* FROM "tags"
Activism
Arts & Crafts
Children & Family
Young Adults
Over 50s
Entertainment
Education
Food
Health & Wellbeing
Outdoors
Sports & Fitness
Places of Worship
Community Hubs
Housing
Legal Advice
Immigration
LGBTQ+
Communities of Colour
C2 Connecting Communities
Computer Access
Conscious Collective Manchester
Flourish
Free WiFi
GM Systems Changers
The Trans Dimension
```

PR #1953, opened 2023-07-17
Database seeding script

ivan-kocienski-gfsc:
Works on #1478

- Cleans up migrations
- database seeding with `rails db:seed`
- fixes the bug with infinite redirects when the database is empty


r-ferrier (commented):
Unfortunately I just can't get this to run. Whatever I do, I just get back the following warning:
`you have data in your DB! (Neighbourhood=5)`

If I don't run `db:setup` first it complains that there is no db, so I run that, and then regardless of whether I running `db:migrate` or `db:schema:load` (or even nothing) I get that error. 

Is there something else I'm maybe doing wrong? I've double checked and I'm confident I'm completely removing the database each time I try it.

ivan-kocienski-gfsc:
Ah. Thanks for checking that. There are so many rails commands and they all differ in slightly different ways.

As I understand it `rails db:setup` does not run the migrations but instead loads the schema, which is defined in `db/schema.rb`. Now- do we want to run the migrations or load the schema? Not sure what the benefits or drawbacks are of each option (speed, maybe?).

The procedure I use is to run
```
rails db:drop # if it exists
rails db:create
rails db:migrate
rails db:seed
rails db:info # sanity check every thing is a-okay
```

The warning you are getting is because the seed script is verifying the database is empty (don't seed over pre-existing data).

If you have problems after running the migration the solution I used was to have this in another rails console:

```
User.destroy_all; Neighbourhood.destroy_all; Tag.destroy_all; Site.destroy_all; Partner.destroy_all; Calendar.destroy_all
```

ivan-kocienski-gfsc:
I also just checked `./bin/setup` on my local machine from not having a database at all and it works also. This is the script mentioned in the README.

r-ferrier (changes requested):
I think this works really well! I've opened a new issue to update the makefile so we can try defaulting to this setup now for local work. We can build on these seeds as we go to make ourselves a really useful little set. I have a couple of questions:

- would it be possible to add four place based sites and four interest based sites & have them marked as published, so that we can keep something occupying the homepage and find your placecal page?

- I'm still nervous about the migrations!  I have some I'll need to run after this PR. Is there any way you can test how a new migration will run against a copy of the production database once this PR has gone in?

ivan-kocienski-gfsc:
My thinking here is to get this merged in as a base and then build on top of it by having follow on tasks for defining
- tags (#1661)
- neighbourhoods (i think rosie is doing work to this effect)
- sites
- partnerships

@r-ferrier As for your second comment what you want to do is re-image a fresh copy of production on your local machine and run `rails db:migrate:status` will show you which migrations have run and which will run on next deploy

r-ferrier (approved):
Ok great, that sounds like a good plan. I tested I could still make a migration locally too on a copy of the production db and all was happy.

PR #1965, opened 2023-07-18
Fix calendar 140 crash

### LINKED ISSUE #1951: [Bug]: 501 on calendar 140
## Description

https://admin.placecal.org/calendars/140/edit

## The problem

Basically calendars have a field called `notices` which is a jsonb column type. Some of the importer (the event resolver) wants to dump hashes in there and other parts of the importer (the CalendarImporterTask) want to dump arrays of strings.

So the UI is crashing because it expects data in one shape but gets another.

ivan-kocienski-gfsc:
Fixes #1951

## Description

Turns out the importer was capturing calendar notices one way and the admin UI was expecting those notices in an entirely different way. 

This has been adjusted so they both use the same way. Tests have been written checking the correct shape of data is present.


test/jobs/calendar_importer/event_resolver_test.rb
```diff
@@ -10,7 +10,7 @@ class EventsResolverTest < ActiveSupport::TestCase
     :location,
     :rrule,
     :last_modified,
-    :ocurrences_between,
+    # :occurrences_between,
```
r-ferrier, line 13: can be deleted

r-ferrier (approved):
I can't see this failing in the wild anymore or work out how to trigger a crash locally to test this, but it looks sensible to me. I've made just a couple of requests to remove a commented out variable, but these are not blocking so am approving anyway!

PR #1980, opened 2023-07-24
Improve neighbourhood seeds

### LINKED ISSUE #1664: Neighbourhood seed data loading
## Description

As a developer who wants to rebuild a clean PlaceCal database I would like it if Neighbourhoods were populated with the correct records and relationships.

## Motivation

As it stands our `bin/setup` onboarding script is incomplete (and it doesn't reset the database). If I try to do anything in PlaceCal (like make a partner or a calendar) it won't work without the correct neighbourhood information.

r-ferrier:
Fixes #1664 

nb: This PR is a draft until the neighbourhood fix PR goes in and we make some decisions about how this runs.

## Description

- Adds neighbourhood seeds from the csvs we use on production, which should ensure our local data better reflects our production data
- Defaults to loading a reduced set of these, enough that we have one of each neighbourhood unit and that every unit loaded has an unbroken chain of parents/children
- Also can optionally load every neighbourhood as a seed if required, flagged using an environment variable
- Adds some console info for when this is happening as it takes a really long time


kimadactyl:
Do you think this should be a rake task instead of a seed?

I think we should prob have all the UK/PHT-specific stuff in its own seed file so we make sure we have separation of concerns here with our specific implementation vs the general PlaceCal system?

r-ferrier:
> Do you think this should be a rake task instead of a seed?
> 
> I think we should prob have all the UK/PHT-specific stuff in its own rake/lib file so we make sure we have separation of concerns here with our specific implementation vs the general PlaceCal system?

I don't know, but it's a good q! The reason it's started as seeds is that I found I couldn't easily drop and repopulate a neighbourhood table once it was set up because of the other relationships dependent on it within the database. 

But maybe seeding a simple subset with a totally fake hierarchy, but then having a rake task be able to `find_or_create` the full set that should be available to us in production would make more sense? I suppose then I could also do away with the migration of this data in my other PR if I implement it instead as a rake task?

kimadactyl:
Yeah - for me seeds are specifically for test data and we should aim to keep the main codebase as clean as possible of that or it gets messy fast. I know we have a few steps left to do in terms of making the geocoder UK-free but it seems a good step to me to have a PHT-specific namespaced rake class rather than having this in the main seeds.

kimadactyl:
There is also this which we had installed before but I think just confused things even further: https://github.com/pboling/seed_migration

ivan-kocienski-gfsc:
> Yeah - for me seeds are specifically for test data and we should aim to keep the main codebase as clean as possible of that or it gets messy fast. I know we have a few steps left to do in terms of making the geocoder UK-free but it seems a good step to me to have a PHT-specific namespaced rake class rather than having this in the main seeds.

I would disagree for neighbourhoods as the platform becomes essentially unusable if the correct neighbourhoods aren't populated. Partners won't be able to have Addresses, is basically what'll happen. And before you hop in and mention that we could switch the postcode resolution system to have some kind of dummy system so it won't hit postcodes.io lookup - yeah that is a nice idea but would be a fair bit of work

kimadactyl:
> the platform becomes essentially unusable if the correct neighbourhoods aren't populated.

I agree but don't see how this is related to my comment. The same data will end up in the database either way. Moving it to a rake task at least allows the separation of concerns to begin.

ivan-kocienski-gfsc:
I was responding to your comment about having a seed for neighbourhoods with only a small made-up geography. We currently have something like this in the seeds script but i doubt its utility

r-ferrier:
I've made an exec decision and changed the way my [other PR](https://github.com/geeksforsocialchange/PlaceCal/pull/1978) works to make the migration I had written into a rake task, removing the need for these to be seeds. 

The rake task is designed to only add neighbourhoods that do not already exist, which means it will be able to be reused on production next time we want to add new data, and in development any time we want to add all the available neighbourhoods. In addition I managed to speed it up by saving references to all the neighbourhoods first into an array, and then only creating or loading models when actually needed. Win win win!

PR #1986, opened 2023-07-24
custom banner text

### LINKED ISSUE #1983: banner text is customizable
image

aaaaargZombies:
fixes #1983 

- add `hero_text` to `Site` model
- limit length to 120 chars
- conditionally display custom text, the component already falls back to "PlaceCal is a community events calendar where you can find everything near you, all in one place."

### DIFF
--- a/app/views/sites/default.html.erb
+++ b/app/views/sites/default.html.erb
@@ -1,10 +1,21 @@
-<%= render(
-  HeroImageComponent.new(
-    image_path: @site.hero_image,
-    image_credit: @site.hero_image_credit
+<% if @site.hero_text.present? %>
+  <%= render(
+    HeroImageComponent.new(
+      image_path: @site.hero_image,
+      image_credit: @site.hero_image_credit,
+      title: @site.hero_text
+      )
     )
-  )
-%>
+  %>
+<% else %>
+  <%= render(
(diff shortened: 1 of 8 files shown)


app/views/sites/default.html.erb
```diff
@@ -1,10 +1,21 @@
...
+  %>
+<% else %>
+  <%= render(
+    HeroImageComponent.new(
+      image_path: @site.hero_image,
+      image_credit: @site.hero_image_credit
+      )
+    )
+  %>
+<% end %>
```
r-ferrier, line 18: Instead of using an if/else to render this, could you could just pass in @site.hero_text and handle the logic for whether to show it or not in the hero image component? 

Instead of setting the title as a default in the params there, you could just do something like
`@title = title.presence || 'default title'` in the init

r-ferrier (changes requested):
This looks really great, I just have one small request to remove a bit of repetition from the code.

app/views/sites/default.html.erb
aaaaargZombies: I tried to wrap just the  `title` argument bit in an if/else but it kept failing to run. This looks loads better, cheers.

r-ferrier (approved):
nice!

Issue #1335, opened 2022-06-20
[Bug]: Events per week in site dashboard is wrong


katjam:
## Description
Viewing event counts on sites dashboard.

## Potential fixes - cnvert Might and Consider to issues if we want to action them

- [ ] Must: Display correct count or remove altogether if not trivial
- [ ] Might: Add Partner count (as well or instead)
- [ ] Consider: Migrating sites list to a datatable

alexandria-gfsc:
I'm wondering if this is due to events not displaying on the page correctly. At least, I've hit on that being a bug during testing of the online events feature.

kimadactyl:
I think it only calculates based on region and doesn't account for tags as well

ivankocienski:
Could you give an example of this problem on staging or production?

kimadactyl:
I feel this is very low priority and will prob be superceded with something smarter soon. Moving to the 1.0 milestone.

katjam:
Maybe relevant to Place Health but clients don;t notice or need.

PR #1989, opened 2023-07-25
Speed up slow loading of admin dashboard

### LINKED ISSUE #1964: [Bug]: admin homepage takes a very long time to load
## Description

When first signing in as an admin it takes a very long time to respond and looks like it is not working.

After signing in it took placecal nearly 12 seconds to send the page back to the user.

Kim suggests the cause is

> its slow because its too busy incorrectly calculating the events total for each site lol

## Steps to reproduce

1. sign in at https://placecal.org/users/sign_in
2. use network tab in dev tools to record if you want more details

https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/78742a19-ae45-4572-928f-764d876cfae5

Dramatically reduced page load and possibly some indication that the site is still working and not crashed so I know to wait.

This article suggests 5 seconds as a max time limit which still seems very slow to me. https://infrequently.org/2022/12/performance-baseline-2023/#2023-content-targets

## Platform (if relevant)

ivan-kocienski-gfsc:
Fixes #1964 #1335 

The simplest solution is to remove the slow queries. Have added a comment to the originating issue covering why it is slow and a few notes on how it could be optimised further.

### DIFF
--- a/app/views/admin/pages/home.html.erb
+++ b/app/views/admin/pages/home.html.erb
@@ -5,10 +5,10 @@
   <p class='has-no-admin-rights-warning'>You have no admin rights. This probably means someone didn't set up your account correctly - please ask whoever invited you!</p>
 <% end -%>
 
-<% ['Partner', 'Calendar', 'User'].each do |model_name| %>
-  <%- if policy(model_name.constantize).create? %>
-    <%= link_to "Add New #{model_name}",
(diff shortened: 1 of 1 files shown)


ivan-kocienski-gfsc:
(note: commented out the code for future reference of how it should work)

kimadactyl:
This works. I think the count is wrong anyway lol

PR #1992, opened 2023-07-25
removed unused search bar from admin area

### LINKED ISSUE #1003: Remove the global search bar
At the moment there is a global search bar, see:

image

We should either make it do something (maybe, show all references to $term across all entities that we have (like, searching `sutton` would fuzzy search for `sutton` and return a neighbourhoods section, a sites section, etc. all with `sutton` somewhere in the resulting models)), or remove it

ivan-kocienski-gfsc:
Closes #1003


aaaaargZombies (approved):
LGTM,  about time this got done

PR #1993, opened 2023-07-27
Tweaked the README.

### LINKED ISSUE #1972: Review README
## Description

Look at the README and check it still accurately and clearly describes the provisioning process.

## Motivation

The steps described in the README should be clear for new developers to follow with little supervision or in-person assistance.

## Notes

Better deliniate between people who want to take the "system postgres" path and people who want to use the "docker postgres" path.

ivan-kocienski-gfsc:
Closes #1972 

- dependencies for docker and gcc
- added bit about lvh.me DNS
- removed the bit about non-docker setup
- changed the default login


ivan-kocienski-gfsc:
i still think we may be missing some dependencies but it is likely these are installed as dependencies of other things (i.e. when install ImageMagick).
- libffi
- libxml2

README.md
```diff
@@ -36,7 +38,9 @@ To run PlaceCal locally you will need:
 
 ## Quickstart with docker for GFSC devs
 
-Make sure all of the above dependencies are installed and ask someone to add your public ssh key to the server.
+First you need to append `lvh.me *.lvh.me` to your `/etc/hosts` file as a DNS alias for `127.0.0.1`.
```
r-ferrier, line 41: This isn't necessary - I haven't done this on my machine and everything works fine running from lvh.me, so I think this should come out.

The clarification below this is also not needed as this section is only for gfsc devs (I titled it 'quickstart with docker for gfsc devs') & without making some changes to the make file to add a non-staff command as well I don't think it will work for anyone else. If you think the title is confusing maybe you could just change that from 'devs' to 'staff' instead?

README.md
```diff
@@ -48,12 +52,6 @@ Local site is now running at `lvh.me:3000`
 
 Some other useful commands for developing can be found in the makefile.
 
-## Quickstart for everyone else
-
-### Set up Postgresql locally with docker
-
-If you don't already have Postgresql installed and running, you can set it up with docker, just run `make docker`. To tear down the docker setup, run `make clean`.
```
r-ferrier, line 55: I think you should leave this in? If you're not GFSC staff this is a simple way to get the DB up and running immediately

r-ferrier (changes requested):
I have some thoughts! Have left some comments.

README.md
ivan-kocienski-gfsc: i put these back how they were.

r-ferrier (approved):
Thanks, looks good

Issue #1828, opened 2023-05-03
Retire ward level sites


katjam:
- [x] Add in ward filter #1733 
- [ ] There are DNS redirects in place for sites that we take down (Hulme & Mosside & other manchester ward site go to manchester)

aaaaargZombies:
this can now begin as we have the ward / neighbourhood filter

katjam:
Won't fix until it comes up again. Ideally we want the word level sites to be active again!

PR #1999, opened 2023-08-03
Filter Partners by Neighbourhood

### LINKED ISSUE #1733: Allow guests to filter partner listings by ward
## User story

As someone browsing manchester.placecal.org, I want to be able to limit my results to just one ward, in order that I can find things just in my area

## Acceptance criteria

- [ ] manchester.placecal.org/partners has a list of wards i can filter to ([inspiration](https://loadstodo.co.uk/whats-on/))
- [ ] clicking a ward shows just partners in that ward
- [ ] clicking a diff ward changes from one to the other (not both)
- [ ] i can clear the ward selection
- [ ] Let people know this can now begin: https://github.com/geeksforsocialchange/PlaceCal/issues/1828

## Implementation plan

r-ferrier:
Fixes #1733 

## Description

- Adds a dropdown filter in the same style as categories to allow visitors to select a neighbourhood to filter on.
- Neighbourhoods shown in the dropdown match those shown by each partner, eg if a partner has 'manchester' as its neighbourhood, 'manchester' appears in the dropdown
- On selecting a neighbourhood from the dropdown, every partner is shown that:
  - has a neighbourhood that matches or is a child of that neighbourhood 
OR 
  - has that neighbourhood set as one of its service areas
 
- The filter can be used in conjunction with a category filter and selections persist when new choices are made in the other filter
- adds tests to assert this is working as expected

## Note

I got off to a slow start on this because I began by filtering Events by Neighbourhood instead. Then I thought we could have both BUT extracting neighbourhoods from events is very slow, and the number of events is very large, which meant that offering neighbourhoods for all the future events in a dropdown that persists across different page views is just too slow - I think this is something for a future version if we load & filter the events using JS instead of navigating to a new page each time.


ivan-kocienski-gfsc:
Some minor issues found

So if you apply a category filter and apply a neighbourhood filter you can get an empty page
Screenshot_2023-08-09_10-18-23

You can open the category filter and neighbourhood filter at the same time
Screenshot_2023-08-09_10-18-57

This neighbourhood has no partners but is still filterable
Screenshot_2023-08-09_10-20-54

You can select the Manchester filter which is redundant(?)
Screenshot_2023-08-09_10-24-18

Not sure how much of a show-stopper those are but as long as the behaviour is acceptable then I think this can be merged

app/controllers/concerns/neighbourhood_filter.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class NeighbourhoodFilter
```
ivan-kocienski-gfsc, line 3: Could be called `PartnerNeighbourhoodFilter` as the filter is specific to partners

app/controllers/concerns/neighbourhood_filter.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def active?
+    @current_neighbourhood.present?
+  end
+
+  def current_neighbourhood?(neighbourhood)
+    @current_neighbourhood.present? && (@current_neighbourhood.id == neighbourhood.id)
+  end
+
+  def apply_to_partner(query = Partner)
```
ivan-kocienski-gfsc, line 20: a little bit of a confusing name as it is not being applied to a partner but applying the filter to the query chain

app/controllers/partners_controller.rb
```diff
@@ -102,8 +97,13 @@ def set_title
         'All Partners'
       end
   end
-  # This controller doesn't allow CRUD
-  # def partner_params
-  #   params.fetch(:partner, {})
-  # end
+
+  def neighbourhoods_from(partners, badge_zoom_level)
```
ivan-kocienski-gfsc, line 101: This should be a class method in the Partner model as it is entirely a database operation

app/views/partners/_category_filter.html.erb
```diff
@@ -2,38 +2,41 @@
...
   </div>
   
   <%= form_tag partners_path, method: :get, class: 'filters__form', data: { turbo: false } do %>
-  <div class="filters__dropdown filters__dropdown--hidden js-filters-dropdown">
-    <div class="filters__group" >
-      <% filter.categories.each do |category| %>
-      <div class="filters__option">
-        <%= radio_button_tag 'category', category.id, filter.current_category?(category), class: "tag__button" %>
-        <%= label_tag 'category', category.name, class: "tag__label" %>
+    <% if params[:neighbourhood] %>
```
ivan-kocienski-gfsc, line 10: I feel that this is breaking the neighbourhood filter abstraction by reading params directly from the query. If the neighbourhood filter is changed this code won't be aware of that.

I don't think there would be anything bad with sending in both filters as they both need to be aware of each other? Then this logic could be moved into the neighbourhood filter where it will all be in one place.

app/views/partners/_neighbourhood_filter.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class="breadcrumb__filters filters">
+  <div class="js-neighbourhood-filters-toggle filters__toggle">
+    <a href="#" data-turbo="false"><span class="icon icon--arrow-down">↓</span> <span class="filters__link">Neighbourhood</span></a>
+  </div>
+  
+  <%= form_tag request.original_url, method: :get, class: 'filters__form', data: { turbo: false } do %>
+
+  <% if params[:category] %>
```
ivan-kocienski-gfsc, line 8: See my earlier comment in the category filter partial about abstractions

ivan-kocienski-gfsc (commented):
Some points found

kimadactyl:
> What sort of selection of neighbourhood names would a user expect to see in a dropdown?

They should be able to filter based on the visible neighbourhood lozenges- so this is a site level configuration

> Should different neighbourhood levels be visible in there?

I think this is the wrong way of thinking about it - filter option should match visible neighbourhood lozenges

> Should the names in the dropdown match the names shown on the partners - is it jarring to see a different list of words in the dropdown to the neighbourhood names on the page?

Yes, or at least the abbreviated one

> What about service areas - do we include all of these?

Yes - i think again its the neighbourhood lozenges that are the key data to filter on. I'm not sure on PlaceCal how it works but on TransDim it shows the address neighbourhood if it has one, then the service area one.

I think the memo people aren't really getting right now is to not set a service area unless it's explicitly something that does outreach. A community centre in Rusholme doesnt "serve" oldham, it serves its literal address. We need to make it clear people should be as specific as possible when selecting this, and things labelled "manchester" prob shouldnt be visible in the hulme results or theyre gonna get super polluted.

> What if they're outside of the site neighbourhood?

If there's more than one service area added you mean? Yeah this is a tricky edge case. I guess for now its fine to have a "various" option like we do for Trans Dim?

> What if service levels are set at a different level to the other areas?

Same as above I think

> If larger areas are visible in the dropdown, eg manchester, do you expect to see everything within manchester, or just partners that have set their service area to manchester specifically? Do you also expect to see partners with service areas within manchester?

I want to punish people who set their service area large if that's a way to think about it. We are about hyperlocal stuff so we want to disencentivse people from saying they are a Manchester service area when they're only active in one ward.

hope that helps a bit - can see why its non trivial!

r-ferrier:
Thanks @kimadactyl - super helpful! I think removing service areas from the equation is maybe the key to this I was missing before - then as you say, can just focus on the badges. Will maybe try to revisit on monday.

kimadactyl:
yeah - to reframe it in simple terms i think for the sake of this ticket it should be an irrelevance where the lozenge text comes from - but we should address that in another ticket to double check how its being generated

kimadactyl:
if theres any way of merging this before we move on that would be great as its a blocker for #1828 

np if not tho

Issue #1961, opened 2023-07-18
Users can select which year their neighbourhood boundaries are from


r-ferrier:
## Description

We need to make it possible for users to select which neighbourhood they want a partner to be associated with when more than one year exists. There should be a select ward form added to the partner admin which should only show when the list of possible wards is greater than 1.

## Acceptance criteria
- [ ] On the edit/add partner page in admin, there is a dropdown of neighbourhood options available for neighbourhoods with more than one available year
- [ ] When a neighbourhood is selected and changes are saved, the neighbourhood is updated on the partner

ivan-kocienski-gfsc:
Closing this as we have decided to not have two parallel sets of neighbourhoods.

Issue #1960, opened 2023-07-18
Make all possible wards available when editing partners


r-ferrier:
## Description

When editing a partner if the ward boundaries have changed over time, they should be able to select which year to base their neighbourhood on. 

## Plan

Within the `show` action for the partners_controller assign a list of possible wards for the partners' postcode

PR #2002, opened 2023-08-08
Neighbourhood updating.

### LINKED ISSUE #2000: Change how neighbourhood updater script operates
## Description

Change the neighbourhood updater script so that it loads newer neighbourhoods "on top of" the existing ward hierarchy and fix the neighbourhood resolution mechanisms to expect this.

## A/C

- [ ] script that updates neighbourhoods
  - [ ] keeps old neighbourhoods so they keep working as before
  - [ ] adds new neighbourhoods "on top"
- [ ] scripts to verify update
  - [ ] address -> neighbourhood resolution
  - [ ] sites -> neighbourhood display
  - [ ] sites -> partners/events display
- [ ] documentation describing the update/verification process

ivan-kocienski-gfsc:
Closes #2000 

This builds on #1978. We no longer keep separate neighbourhood sets by release data and only have the one set. Each neighbourhood record tracks the last version of the ONS data it was loaded from.

- adds new neighbourhood records, updates older ones, leaves neighbourhoods (that are not mentioned) alone
- reparents neighbourhoods as necassery
- simplifies how addresses find neighbourhoods: now only keyed off of the `unit_code_value` (ONS ID). These should always be unique and we only store the latest version of the record.
- optimises script for performance

There is a document (in /doc/updating-neighbourhoods.md) that covers the procedure. A number of support scripts have been written that allow us to validate that something has not gone wrong that will impact sites or end users.

The following issues are moot as we only have one set of data
- #1961
- #1960

## Likely issues

- Admins will see (and be able to select) all neighbourhoods from the UX, even "retired" entries. Locking off the scope so `where(release_date: LATEST_DATE)` would fix this
- Neighbourhoods are never deleted and will accumulate making the UI slower, including public guests to placecal sites.
- No way for partner admins to see if their address is obsolete

There are probably more problems here but they will be turned into issues and placed in the appropriate milestone


r-ferrier (changes requested):
Ok, I've run this and I think it works! My only niggle is with the duplicate neighbourhoods still sometimes showing up, but as you suggest, locking off scope to latest wherever these are shown would be a really easy way to fix this while preserving ex-neighbourhoods and their relations, so maybe you could make a new issue for this once this is merged and address this separately. 

I also have one change request: I'm not sure what the verifications you need to run before and after running the neighbourhood importer are really there for. and I couldn't really understand what they were supposed to be telling me. If they're not essential to the process and are just there for debugging, I don't think you should say that they need to be run in the updating neighbourhoods documentation. If they ARE essential to the process, then I think they need to run as part of the rake task, and not be separate steps you should have to run beforehand.

ivan-kocienski-gfsc:
Have simplified the documentation and moved the checking scripts out into the `placecal` namespace

r-ferrier (approved):
Nice! This makes much more sense to me.

PR #2018, opened 2023-08-16
Duplicate neighbourhoods

### LINKED ISSUE #2017: [Bug]: Duplicate neighbourhoods
## Description

Because the neighbourhood update script only adds new entries, some neighbourhoods may be "duplicated" (have the same name but differing release dates). 

Screenshot_2023-08-15_15-04-55

## Steps to reproduce

1. Visit any admin site edit page
2. Start typing a new neighbourhood to bring up the selector
3. Search for "Abbey Road"
4. Duplicate Abbey Road entries appear

## What you expected to happen / Acceptance Criteria

- [ ] For Partner Service Area Neighbourhoods, Site Neighbourhoods and User Neighbourhoods:
  - [ ] Show which neighbourhoods that entity is linked to (legacy or current).
  - [ ] Current neighbourhoods should be displayed as before.
  - [ ] Legacy neighbourhoods should be indicated in some way.
  - [ ] When saving an entity with legacy neighbourhoods, those legacy neighbourhoods should be retained.
  - Neighbourhood picker:
    - [ ] Shows all the latest neighbourhoods as before
    - [ ] Also includes only the legacy neighbourhoods that that entity is linked to
    - [ ] Cannot add "new" legacy neighbourhoods
    - [ ] Can remove legacy neighbourhoods

## Implementation plan / notes

Put a scope on service areas / neighbourhoods to only look at the newest neighbourhoods.

Pages that need to be fixed
- partners edit
    - service areas
- users edit
    - neighbourhoods
- sites edit
   - primary neighbourhood
   - other neighbourhoods
- neighbourhoods index

May need to check the neighbourhood filter on /partners for guest user site partners page.

ivan-kocienski-gfsc:
Closes #2017 

For such a simple issue this has struck a nerve at the core of our data models.

Tasks covered:
- latest neighbourhood release date is defined in Neighbourhood model
- a scope to only find `latest_release` in Neighbourhood
- update to UX so neighbourhood pickers only show latest neighbourhoods BUT will also show only the legacy neighbourhoods that have already been selected by that entity.
  - sites, partners (service areas) and users
- latest version is displayed on admin neighbourhood page
- Neighbourhood updating documentation: added text covering how to set the `latest_release` date in the Neighbourhood model

## Other work

There is a rake task `rails placecal:find_obsolete_neighbourhoods` that will look at every user, site and partner to see if they are linking to a legacy neighbourhood. Output is in markdown suitable for uploading to notion. This can be used by the client support team to possibly reach out to users to adjust their neighbourhoods.

## Example Neighbourhood picker logic

When a user has a legacy neighbourhood the picker looks like
Screenshot_2023-08-17_17-07-51

When a user has no legacy neighbourhoods only the latest set is used
Screenshot_2023-08-17_17-08-40

## Caveats

Problems found that would be out of scope for this PR:
- we need to prune legacy neighbourhoods that aren't linked to anything
- there are neighbourhoods that can be auto-replaced (in certain circumstances)
- sites' primary neighbourhoods that are linked to a legacy neighbourhood cannot currently be changed in the UX
- there is no way for users to tell if a partner's address is linked to a legacy neighbourhood.
- i have left the address postcode lookup routine (`find_from_postcodesio_response`) so it can find any neighbourhood (legacy or current) so that partners with obsolete addresses will keep working. because this method is only called with responses from a postcodes.io request i am assuming only the latest neighbourhoods will be found.
  -  (but obsolete partner address information is printed out in the `find_obsolete_neighbourhoods` script mentioned earlier)

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -29,8 +29,9 @@ module UsersHelper
     policy_scope(Partner).all.order(:name).pluck(:name, :id)
   end
 
-  def options_for_neighbourhoods
+  def options_for_user_neighbourhoods
     policy_scope(Neighbourhood)
+      .latest_release
       .where('name is not null and name != \'\'')
       .order(:name)
       .all
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -29,13 +29,21 @@ module UsersHelper
     policy_scope(Partner).all.order(:name).pluck(:name, :id)
   end
 
-  def options_for_user_neighbourhoods
-    policy_scope(Neighbourhood)
-      .latest_release
-      .where('name is not null and name != \'\'')
+  def options_for_user_neighbourhoods(for_user)
+    legacy_neighbourhoods = for_user.neighbourhoods.where.not(release_date: Neighbourhood::LATEST_RELEASE_DATE)
+
+    scope = Neighbourhood.find_latest_neighbourhoods_maybe_with_legacy_neighbourhoods(policy_scope(Neighbourhood), legacy_neighbourhoods)
+
+    scope
       .order(:name)
       .all
-      .collect { |ward| [ward.contextual_name, ward.id] }
+      .collect do |ward|
+        if ward.legacy_neighbourhood?
+          ["#{ward.contextual_name} - #{ward.release_date.year}/#{ward.release_date.month}", ward.id]
+        else
+          [ward.contextual_name, ward.id]
+        end
+      end
   end
 
   def options_for_tags
--- a/app/models/neighbourhood.rb
+++ b/app/models/neighbourhood.rb
@@ -35,6 +35,10 @@ class Neighbourhood < ApplicationRecord
 
   scope :latest_release, -> { where release_date: LATEST_RELEASE_DATE }
 
+  def legacy_neighbourhood?
+    release_date < Neighbourhood::LATEST_RELEASE_DATE
+  end
+
   def shortname
     if name_abbr.present?
       name_abbr
@@ -113,6 +117,15 @@ class Neighbourhood < ApplicationRecord
     def find_from_postcodesio_response(res)
       ons_id = res['codes']['admin_ward']
       Neighbourhood.where(unit_code_value: ons_id).first
+    end
+
+    def find_latest_neighbourhoods_maybe_with_legacy_neighbourhoods(scope, legacy_neighbourhoods)
+      scope = scope
+              .where('name is not null and name != \'\'')
+              .latest_release
+
+      scope = scope.or(where(id: legacy_neighbourhoods.pluck(:id))) if legacy_neighbourhoods.any?
+      scope
     end
   end
 
--- a/app/models/neighbourhood.rb
+++ b/app/models/neighbourhood.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Neighbourhood < ApplicationRecord
+  # WARNING: this must be updated for every new ONS dataset
+  #    see /lib/tasks/neighbourhoods.rake
+  LATEST_RELEASE_DATE = DateTime.new(2023, 5).freeze
+
   has_ancestry
   has_many :sites_neighbourhoods, dependent: :destroy
   has_many :sites, through: :sites_neighbourhoods
@@ -28,6 +32,8 @@ class Neighbourhood < ApplicationRecord
             allow_blank: true
 
   before_update :inject_parent_name_field
+
+  scope :latest_release, -> { where release_date: LATEST_RELEASE_DATE }
 
   def shortname
     if name_abbr.present?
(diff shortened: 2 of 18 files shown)


app/models/neighbourhood.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Neighbourhood < ApplicationRecord
+  # WARNING: this must be updated for every new ONS dataset
```
aaaaargZombies, line 4: this should probably be in the documentation md file

app/views/admin/neighbourhoods/index.html.erb
```diff
@@ -25,3 +26,4 @@ var columns = [
 <p><%= link_to safe_neighbourhood_name(root), admin_neighbourhood_path(root) %></p>
 <% end %>
 
+<h5>Latest Release Date: <%= Neighbourhood::LATEST_RELEASE_DATE %></h5>
```
aaaaargZombies, line 29: nit pick / maybe non-issue but maybe this could be explicit about it being ONS dataset from that date. just so there's no ambiguity about what has been released.

app/helpers/users_helper.rb
```diff
@@ -29,8 +29,9 @@ def options_for_partners
     policy_scope(Partner).all.order(:name).pluck(:name, :id)
   end
 
-  def options_for_neighbourhoods
+  def options_for_user_neighbourhoods
     policy_scope(Neighbourhood)
+      .latest_release
       .where('name is not null and name != \'\'')
```
aaaaargZombies, line 35: just a question for my own understanding. Is there a reason we wouldn't want this as part of the scope?

aaaaargZombies (dismissed):
I think this is OK to go, we need to make a note in the release email that partners are being migrated to the new ONS definitions and message David to decide how/who to contact to warn about opting in to a new service area if people have been effected by this in ways they are not happy with.

app/models/neighbourhood.rb
ivan-kocienski-gfsc: [like here](https://github.com/geeksforsocialchange/PlaceCal/blob/ik-2017-duplicate-neighbourhoods/doc/updating-neighbourhoods.md#adjust-the-neighbourhood-model)?

app/helpers/users_helper.rb
ivan-kocienski-gfsc: This constrains the front-end to only find the latest neighbourhoods for the neighbourhood pickers

app/views/admin/neighbourhoods/index.html.erb
ivan-kocienski-gfsc: have changed to "ONS Latest Release Date Loaded: `<%= Neighbourhood::LATEST_RELEASE_DATE %>`"

aaaaargZombies:
top of partner admin ui

image

mid page on site admin ui

image

two places where we need to decide if it's important to show that this is a legacy neighborhood...

OK just did some testing on my local machine with a fresh copy of prod data

- the mossley site has the primary neighborhood of  `mossley, tamesaide` (2019)
- add a service area for "burnt roti" to `mossley, tamesaide` (2023 - we don't give access to 2019)
- "burnt roti"  is not showing up on http://mossley.lvh.me:3000/partners

removing the address for "Emmaus Mossley (70)" and replacing it with my address gives back a '23 neighborhood. When I resave it with it's original neighborhood (Longlands Mill, OL5 9AH) this produces a '19 neighborhood

I  think showing which address is here might be important if partners aren't showing up on sites se we can see if we're comparing the correct neighborhoods

ivan-kocienski-gfsc:
I think i understand what you are saying:

- some partners have legacy neighbourhoods (as service areas or addresses)
- if a site has a legacy neighbourhood it will pick up partners with matching legacy neighbourhoods
- a site may want to "upgrade" their legacy neighbourhoods by adding new neighbourhoods (not removing legacy hood)
- then legacy partners will keep appearing on sites with legacy neighbourhoods

?

aaaaargZombies:
> I think i understand what you are saying:
> 
>     * some partners have legacy neighbourhoods (as service areas or addresses)
> 
>     * if a site has a legacy neighbourhood it will pick up partners with matching legacy neighbourhoods
> 
>     * a site may want to "upgrade" their legacy neighbourhoods by adding new neighbourhoods (not removing legacy hood)
> 
>     * then legacy partners will keep appearing on sites with legacy neighbourhoods
> 
> 
> ?

yes I think that's the gist of it, basically if we are encouraging partners to update to a new address we need to make sure they have somewhere to go. or at least know where to look if they disappear and someone asks for help.

aaaaargZombies (approved):
OK re-approved.

I have made a gotchas document here that should help if any Partners find they are not being picked up by a Site they expected to be on.

https://www.notion.so/gfsc/PlaceCal-Neighborhood-Gotchas-fc2970299bda4c7b8c0e148ec2386778

PR #2025, opened 2023-08-22
Update cron to run once an hour

### LINKED ISSUE #2024: [Bug]:  Calendar importer is running too often
## Description

Calendar importer runs every 10 minutes which feels like overkill and is (maybe) massively slowing down staging. Proposal to run just once an hour instead to see if this improves the situation!

r-ferrier:
Fixes #2024 

## Description

The calendar importer is currently running once every ten minutes and taking close to that amount of time to complete. We think this is making staging unworkably slow so want to try running it less frequently to see if this improves things. This could also make production much quicker for users.

### DIFF
--- a/app.json
+++ b/app.json
@@ -10,7 +10,7 @@
 	"cron": [
 		{
 			"command": "bundle exec rails import:scan_for_calendars_needing_import",
-			"schedule": "*/10 * * * *"
+			"schedule": "0 * * * *"
 		}
 	]
 }


ivan-kocienski-gfsc:
Will this affect production and do we want to drop production to once-an-hour scan times?

r-ferrier:
> Will this affect production and do we want to drop production to once-an-hour scan times?

That is my intent! I'm struggling to see the scenario in which importing calendar events more regularly than once an hour is necessary - and can't users re-queue idle calendars manually if they've made urgent changes to something? Or are those queued calendars then only picked up when the importer job runs?

ivan-kocienski-gfsc:
Has @kimadactyl signed off on this?

ivan-kocienski-gfsc (approved):
Lets do this

PR #2058, opened 2023-09-19
remove github diff link from admin UI

### LINKED ISSUE #2055: Make github release link in admin interface friendlier
## Description

This is not helpful for most users, remove for everyone except root? (Or maybe just remove for everyone)

## Acceptance criteria

- [ ]  build hash / link to latest live commit is not shown to citizen / editor users

## plan

I'm going to fully remove this. It takes you to a diff that is only intelligible to developers who will be able to find it on github themselves. It is also not a great diff because it is the merge from `main` to `prod` that combines all other PRs and strips the context.

It will still appear in the footer of the public site so is discoverable by people not familiar with the project. Perhaps this should be a link to current release of README though, but a question for another day.

aaaaargZombies:
fixes #2055 remove link to github diff that is confusing information for most users


ivan-kocienski-gfsc:
Hmm if you are removing it from the admin page because it is confusing and unused then why leave it on the guest-user page?

ivan-kocienski-gfsc:
So an idea that is more useful may be to show the last tag.

On my machine if I run `git describe --tags --abbrev=0` then I get "v0.16.1"

And the link to that tag would be `https://github.com/geeksforsocialchange/PlaceCal/tree/v0.16.1`

ivan-kocienski-gfsc:
Ah okay, we were wondering if anyone used this functionality.

aaaaargZombies:
> This is an active step back for me that makes my job harder and I would prefer not to merge it personally.
> 
 >Replacing it with the version number would be an improvement but just removing it entirely is unhelpful.
 
 @kimadactyl  the use case and is it just roots that should have it? or everyone. We are trying to reduce dead ends in the UI/general surface area for bugs/ irrelevant content and overwhelm for people who struggle with technology.

kimadactyl:
As an admin of any kind
When I access the PlaceCal backend
I want to see when the software has been updated

As someone waiting for the software to deploy an urgent fix
When I am waiting for it to deploy
I want to be able to spam refresh page until i can see its deployed

aaaaargZombies:
So link to the latest release is perhaps more useful than the diff? would be nice to get a date rather than a hash because that's computer speak

https://github.com/geeksforsocialchange/PlaceCal/releases/latest

ivan-kocienski-gfsc:
That is why I suggested the tag name rather than the diff (but poking around on prod i found it's not so simple to get the tag from the command line).

Looks like there may be something in the deploy script that captures this git info.

kimadactyl:
yeah either verison number or "last release was TIME ago" are both good

ivan-kocienski-gfsc:
You may be able to get the latest release date by adding something like this to an initializer
```ruby
module ReleaseDate
    extend self
    def read
        @release_date ||= File.stat(Rails.root.join('README.md')).ctime
    end
end
```

And then putting `<p>released on <%= ReleaseDate.read %></p>` in the footer

aaaaargZombies:
Ivan points out that in dokku we don't get full access to git we would want to set and environment variable ourselves on each new deploy using something like

```sh
dokku config:set your-app-name LATEST_GIT_TAG="$(curl  "https://api.github.com/repos/geeksforsocialchange/PlaceCal/tags" | jq -r '.[0].name')"
```
Not quite sure how to trigger that each release, the tag `curl | jq` command would work in the template but probably not wise to run it each time someone accesses the admin page.

ivan-kocienski-gfsc:
Not sure if what you put there will be set every deploy as i think once environment variables are set (in that way) then they just keep the same value?

app/views/layouts/admin/_admin_navigation.html.erb
```diff
@@ -30,8 +30,11 @@
 <p class="small text-info ml-3">
 <i class="fa fa-bug"></i> <%= link_to "Report a bug", "https://github.com/geeksforsocialchange/PlaceCal/issues/new?assignees=&labels=&template=bug_report.md" %>
 
+<% tag = `curl  "https://api.github.com/repos/geeksforsocialchange/PlaceCal/tags" | jq -r '.[0].name'` %>
```
ivan-kocienski-gfsc, line 33: oh golly.

this is a bit of a security/performance nightmare here. it is not good-practice to run shell scripts like this.
1. curling from a URL may fail
2. the URL can return anything, which we are then piping into `jq` which may have a bug and execute malicious code, possibly exfiltrating sensitive info from our production servers to anywhere
3. what happens if these commands fail? will it bring down production for everyone?
4. do we have jq installed on staging / production?
5. this can be done in ruby, once, and cached

ivan-kocienski-gfsc (changes requested):
This has some blocking problems, sorry dude

app/views/layouts/admin/_admin_navigation.html.erb
```diff
@@ -30,8 +30,11 @@
...
+<br><i class="fa fa-git"></i> Version: <tt><%= link_to tag, "https://github.com/geeksforsocialchange/PlaceCal/releases/latest" %></tt>
+</p>
 
-<% build = ENV['GIT_REV'] ? ENV['GIT_REV'][0,7] : 'development' %>
-<br><i class="fa fa-git"></i> Build: <tt><%= link_to build, "https://github.com/geeksforsocialchange/PlaceCal/commit/#{build}" %></tt>
 
+<p>
+<%= debug(ENV) %>
```
ivan-kocienski-gfsc, line 39: It's probably okay but should this be in here?

ivan-kocienski-gfsc (commented):
Small query here

kimadactyl:
I would say I don't think this is really needed for this release - for me I think this has been a lot of work for something of unclear benefit and we could just return to it at a later date?

ivan-kocienski-gfsc:
I am going to park this as it became a lot more complicated and raised questions about what it is we are trying to fix/implement here. Notes made on #2055

PR #2059, opened 2023-09-19
fix: prevent duplicate service_area crashing placecal

### LINKED ISSUE #2057: [Bug]: Adding duplicate service areas causes a db crash
## Description

If you try to save a partner and add two service areas that are the same service area at once, db borks

## AC

- [ ] You cannot submit a form with duplicate service areas

## plan

There is not an easy / obvious way to modify the `add service area` button to show a filtered list of options based on the previous time the form was updated.

So I will filter the params at the controller, to ensure they are unique. This is probably safer anyway because it also means the any API requests are dealt with, relying on server validation instead of client validation.

aaaaargZombies:
fixes #2057

- ensure new service areas are unique on `create` and `update` in the partner controller
- add system tests to replicate user attempting to perform this bad action

Desired behavior = Adding the same service areas twice has the same effect of adding it once.


ivan-kocienski-gfsc (approved):
Hmmm. to be honest this code looks a rough, but it works and is tested.

I think I'd prefer to see a solution that is in the model as this represents business logic.

ivan-kocienski-gfsc:
your solution is okay and maybe we should go with that. the way rails validates relations is complex and i am not sure if my solution work the same way.

Issue #2063, opened 2023-09-19
Audit and fix Partnership Admin permissions


kimadactyl:
## User story

As a partnership admin (PA), I should only have access to relevant data and actions related to my role.

## Acceptance criteria

- [x] A basic partnership admin (ie the Trans Dimension Organiser) has the permissions `neighbourhood_admin`: region covered by their activities, `partnership_admin`: tag they admin for. They might also be a site admin, partner admin, or editor.

In the admin interface, a PA can:

View everything a NA can, with restrictions:
- [x] All of the neighbourhoods
- [x] #2168
- [x] #2169
- [x] #2170
- [x] All Category Tags

Create everything a NA can, with restrictions:
- [x] #2171
- [x] #2172
- [x] #2173

Edit everything a NA can, with restrictions:
- [x] #2174
- [x] #2175
- [x] Everything about a calendar
- [x] #2176

Delete everything a NA can, with restrictions:
- [ ] #2177
- [x] Calendars within their scope

aaaaargZombies:
> - CRUD only partners in partnership The Trans Dimension
> - CRUD partners not in The Trans Dimension partnership (e.g. this means that it is impossible for this role to make a partner without a Trans Dimension partnership)

Will need to figure out how do existing partners get added to a new Partnership?

kimadactyl:
> Will need to figure out how existing partners get added to a new Partnership?

Agree totally this is an issue - we should open a ticket, but for now this isn't affecting anyone and if it came up GFSC staff could fix manually

kimadactyl:
> create users assigned to partners tagged The Trans Dimension 

afaik this already works and just needs tests adding

kimadactyl:
> CRUD news articles tagged The Trans Dimension

If we are removing this from scope are we just telling Georgie they can't edit the news articles any more?

ivan-kocienski-gfsc:
## Notes on: Acceptance criteria

- [ ] TDO has permissions (role) "editor"
- [ ] TDO has "neighbourhood_admin: england" (has a neighbourhood)
- [ ] TDO has "partnership_admin: the trans dimension" (has partnership tag)
- [ ] TDO is set as site_admin for The Trans Dimension (is owner of site)

In the admin interface, TDO can:

- [ ] CRUD only partners in partnership The Trans Dimension
  - extend permissions so a user with a partnership tag can see partner admin page
  - filter partners by partnership tag
  - cannot remove trans dimension partnership (tag).
[...]

Issue #2062, opened 2023-09-19
PlaceCal Norwich's Organiser should have the proper set of permissions


kimadactyl:
## User story

As PlaceCal Norwich's Organiser (PNO), I should only have access to relevant data and actions related to my role.

This should form the basis for a generic "new neighbourhood" type of user. It does not cover adding people to existing areas.

## Acceptance criteria

- [ ] PNO has permissions "citizen" and "neighbourhood_admin: norwich" and "site_admin for PlaceCal Norwich" and nothing else

In the admin interface, PNO can:

- [x] CRUD only partners in Norwich (including adding categories and facilities tags) see #2106
- [x] CRUD only calendars attached to partners in Norwich (see #2107)
- [x] #2109
- [x] #2113
- [x] #2108

PNO cannot:

- [x] #2106
- [x] #2107
- [x] #2104
- [x] #2105
- [x] Do anything with partnerships (just needs verifying)
- [x] Have anything to do with news articles
- [x] #2103
- [x] #2119

## Implementation plan

_To be written by the developer_

kimadactyl:
Prob this list needs a bit of a discussion

r-ferrier:
Things that a `site_admin` in charge of a place based site can edit:
- name
- place name
- tagline
- hero
- description
- published?
- theme
- logo, footer, other image things like credit
- badge zoom level

Things they should not have
- site admin assignation
- other neighbourhoods to include
- other neighbourhood info

kimadactyl:
Just noting theres more fields than that they shouldnt have - the should list looks correct tho :)

ivan-kocienski-gfsc:
meeting notes:

trying to not think about how non-root users can create other non root users as this is pretty complex and needs more thought.

ivan-kocienski-gfsc:
Note: users cannot remove (TD) partnership tags as this will push things out of their scope

kimadactyl:
> Create users assigned to partners in Norwich

I think this already works though? If not fine to remove but it's my understanding this is working and just needs confirming

> Note: users cannot remove (TD) partnership tags as this will push things out of their scope

Wrong ticket?

ivan-kocienski-gfsc:
## Acceptance criteria

- [ ] PNO has permissions "citizen"
  - given a user with a citizen role
- [ ] PNO is set as neighbourhood_admin: norwich
  - has one neighbourhood, norwich
- [ ] PNO is set as site_admin for PlaceCal Norwich
[...]

PR #2067, opened 2023-09-20
Removing "dead" links

### LINKED ISSUE #2056: [Bug]: dead links
## Description

Given I am a partner admin
That has a partner
When I log in
And visit the edit partner page
Then I should see the list of users for this partner
And that list should not be links to those user pages

Given I am a neighbourhood_admin
And a partner exists
That has a service area
When I visit the neighbourhood page
Then I should see a list of partners for that neighbourhood
But they should not be links to partners

- Partners with service areas in a neighbourhood are listed in the neighbourhood admins partner list, but cannot be visited.
- Partner admins can click add new partner admin from an individual partner page but it returns them to the dashboard

## AC

- [ ] Are visible but are not links

ivan-kocienski-gfsc:
Closes #2056 

- admin show neighbourhood page
  - will show you which partners are associated with this neighbourhood (via address or service area)
  - if you are a partner admin then you get links to those partners
  - otherwise just the name text of the partner
- admin edit partner
  - if you are root - you can add users to this partner - can see a list of links to users with this partner
  - if you are not
    - can only see a list of emails of users (not links)


app/views/admin/neighbourhoods/show.html.erb
```diff
@@ -31,7 +31,13 @@
 <% if @neighbourhood.address_partners.present? %>
 <ul class="list-group">
   <% @neighbourhood.address_partners.order(:name).each do |partner| %>
-  <li class="list-group-item"><%= link_to partner.name, edit_admin_partner_path(partner) %></li>
+  <li class="list-group-item">
+    <% if current_user.partner_admin? -%>
```
r-ferrier, line 35: Unfortunately I think this will only check if you're a partner admin of any partner, not this specific one. I think based on the user stories we generated yesterday with kim the logic here needs to be that you should only be able to edit this partner if you own it, it is in your neighbourhood, or if you are a root. This should probably be something we mandate at a user model level, through something like can_edit_partner(partner) or similar?

app/views/admin/neighbourhoods/show.html.erb
```diff
@@ -42,22 +48,34 @@
...
 
-<hr>
-<h2>Sites</h2>
-<p>Found <%= @neighbourhood.sites.count %>.</p>
-<% if @neighbourhood.sites.present? %>
-<ul class="list-group">
-  <% @neighbourhood.sites.order(:name).each do |site| %>
-  <li class="list-group-item"><%= link_to site.name, edit_admin_site_path(site) %></li>
+
+<% if current_user.site_admin? %>
```
r-ferrier, line 63: neighbourhood admins and root users also need to be able to see all the info about what sites are attached to a neighbourhood. Neighbourhood admins should not then be able to edit a site but I have addressed this in my neighbourhood pr!

app/views/admin/partners/edit.html.erb
```diff
@@ -59,14 +60,21 @@
...
+          <%= p_user.email %></li>
+        <% end %>
       <% end -%>
     </ul>
 
-    <p><%= link_to 'Add new partner admin',
-                    new_admin_user_path(partner_id: @partner.id),
-                    class: "btn btn-primary mb-3 mr-2 btn-sm" %>
-    </p>
+    <% if current_user.root? %>
```
r-ferrier, line 72: anyone who can create a partner needs to be able to add and remove admins, so neighbourhood admins whose neighbourhood this partner resides in also need to be able to do this - probably again, who can do this should be mandated at model level on the user when given a partners id

r-ferrier (changes requested):
sorry Ivan, a few things here that I don't think are quite working as they should. Give me a shout if you have any qs. I'm basing this logic on the user stories we generated yesterday as our new source of truth:
https://github.com/geeksforsocialchange/PlaceCal/issues/2062
https://github.com/geeksforsocialchange/PlaceCal/issues/2063

app/views/admin/partners/edit.html.erb
ivan-kocienski-gfsc: i don't think you can even see this page unless you are a partner admin. see this [policy](https://github.com/geeksforsocialchange/PlaceCal/blob/febc6292df99ad11339dbd1bc9204b870effc316/app/policies/partner_policy.rb#L20) logic.

might as well take the check out. unless there is a specific subset of those permissions that should only be able to "see" the users

app/views/admin/neighbourhoods/show.html.erb
ivan-kocienski-gfsc: I have updated this to use the same policy check that "update partner" action uses.

app/views/admin/neighbourhoods/show.html.erb
ivan-kocienski-gfsc: i have updated this to use the same policy checks for [sites](https://github.com/geeksforsocialchange/PlaceCal/blob/febc6292df99ad11339dbd1bc9204b870effc316/app/policies/site_policy.rb#L24) and [neighbourhoods](https://github.com/geeksforsocialchange/PlaceCal/blob/febc6292df99ad11339dbd1bc9204b870effc316/app/policies/neighbourhood_policy.rb#L21).

ivan-kocienski-gfsc:
If you have a specific logic in mind could you update the A/C in the issue to reflect that?

ivan-kocienski-gfsc:
@r-ferrier  how about: we only show links to users who have a policy covering the update model action? 

so in neighbourhood/show
- only show partners by address / service area if that user can update partner action
- only show sites if that user can update sites

r-ferrier (approved):
I like this approach, nice to handle the logic in the policy, and we can get granular over how it works in there once we've made those calls

PR #2068, opened 2023-09-20
Supporting generic iCal feeds

### LINKED ISSUE #1985: Calendar URL field is giving a false negative "not supported" message if using a generic ical feed
Currently the default is to auto-detect the importer mode that needs to be used when adding a calendar, however this doesn't work for some URLs, especially self-hosted .ics files. The text description should prompt the uploader to check the importer mode is appropriate, especially if their URL fails to import the first time. 

## Acceptance criteria

- [x] `https://www.birchcommunitycentre.co.uk/events.ics` can be added as a calendar url without getting told it cant be parsed
- [x] `https://membership.ambition.rip/public_calendar.ics` can be added as a calendar url without getting told it cant be parsed
- [ ] URLs we don't recognise can still be saved in general so the admin team can have a look at them

ivan-kocienski-gfsc:
Closes #1985 

If the URL starts with `https://` and ends in `.ics` then it is an ical feed.


kimadactyl:
also `webcal://`!

ivan-kocienski-gfsc:
> also `webcal://`!
Still supported by the earlier patterns.

> 
> also plain http?
fine.

test/jobs/calendar_importer/calendar_importer_task_test.rb
```diff
@@ -189,4 +189,26 @@ class CalendarImporterTaskTest < ActiveSupport::TestCase
       assert_equal "Source responded with invalid JSON (783: unexpected token at '{ \"key\": \"va')", error.message
     end
   end
+
+  test 'can import from generic iCal feed' do
+    VCR.use_cassette(:generic_ical_feed, allow_playback_repeats: true) do
+      calendar = create(
+        :calendar,
+        name: 'Generic iCal Calendar',
+        source: 'https://www.birchcommunitycentre.co.uk/events.ics',
```
aaaaargZombies, line 198: can we get tests for the `http` and `webcal` type links that got added (I'm assuming) after this test was written.

aaaaargZombies (changes requested):
small change request then good to go, I think Kim raises some good points about having very specific URLs for partners in there but they are outside the scope of this issue and shouldn't hold up this important feature.

test/jobs/calendar_importer/calendar_importer_task_test.rb
ivan-kocienski-gfsc: I put in a test for webcal. i can put in *a* test for `http` but to be thorough and check all the patterns i'd have to write an `http` variant for every type of ical feed we support.

ivan-kocienski-gfsc: if that is something that would be valuable then let me know

aaaaargZombies (approved):
cheers, looks good, wild we didn't have this sooner

kimadactyl:
> We also should more gracefully handle an unknown URL and make a suggestion to try one of the generic settings.

I don't think this was addressed in the PR

PR #2071, opened 2023-09-20
Admin jobs page

### LINKED ISSUE #2070: [Bug]: Calendar importer status is not working
## Description

The importer status returns zero for all its outputs:
Screenshot_2023-09-20_16-25-31

This is because the some keys are keywords and not strings, or something like that.

## A/C
- [ ] Job state count table has counts in it

ivan-kocienski-gfsc:
Closes #2070

Importer status table has actual counts in it

Screenshot_2023-09-20_16-36-44


r-ferrier (approved):
LGTM, let's get it on main and we can verify on staging

PR #2078, opened 2023-09-26
Allow partners to be filtered by neighbourhood

### LINKED ISSUE #1733: Allow guests to filter partner listings by ward

r-ferrier:
Fixes #1733 

### Description

This is a complete redo of https://github.com/geeksforsocialchange/PlaceCal/pull/1999 on a new branch. I did this in this way because the other PR was so far behind main at this point and the logic was so different that it seemed more sensible to just start again. 

Now the names you can select from in the filter exactly match the names shown on the partners index, and if you click on a name it will only show you matching names, and not things like children of that neighbourhood as well.

I have tried my absolute best to address all the points made by @ivan-kocienski-gfsc and match the logic specified by @kimadactyl in the previous PR but please let me know if you see any showstoppers, or anything I've missed.

I moved some of the precarious name logic into the partner model and removed an unused view that used to use it. Then if we want to change how the names are assigned in the future, it hopefully won't break this filter!


app/components/place_partner_preview/place_partner_preview_component.rb
```diff
@@ -3,22 +3,20 @@
...
   delegate :name, to: :previewee
 
   def link
     previewee
   end
 
-  def show_neighbourhood?
-    show_service_area? ? false : show_neighbourhoods
+  def show_service_area?
+    service_areas.count.positive?
```
kimadactyl, line 15: think u can do `service_areas.any?`

app/models/neighbourhood.rb
```diff
@@ -113,6 +113,12 @@ def country
     ancestors.where(unit: 'country').first
   end
 
+  def name_from_badge_zoom(badge_zoom_level)
+    return district&.shortname if badge_zoom_level == 'district'
```
kimadactyl, line 117: this seems a lil hard to read could it just be a normal if / else?

app/models/partner.rb
```diff
@@ -311,10 +311,43 @@ def owned_neighbourhood_ids
     neighbourhood_ids
   end
 
+  def neighbourhood_name_for_site(badge_zoom_level)
+    if service_areas.count.positive?
+      if service_areas.count > 1
+        'various'
```
kimadactyl, line 317:
```suggestion
        'Various'
```

app/models/partner.rb
```diff
@@ -311,10 +311,43 @@ def owned_neighbourhood_ids
     neighbourhood_ids
   end
 
+  def neighbourhood_name_for_site(badge_zoom_level)
+    if service_areas.count.positive?
```
kimadactyl, line 315:
```suggestion
    if service_areas.any?
```

app/components/place_partner_preview/place_partner_preview_component.rb
kimadactyl, line 15:
```suggestion
    service_areas.any?
```

app/models/partner.rb
kimadactyl: theres perhaps no need for a nested if here - could just do elsifs for >1, 1, and 0

r-ferrier: for some reason rubocop is losing its mind over an elsif here so I'm leaving the nested statement for now, but will change the string.

kimadactyl: ha, when i last used rubocop it was losing its mind over too many levels of indentation, sigh

ivan-kocienski-gfsc:
There seems to be a bug (or undefined behaviour) when combinations of category filter and neighbourhood filter are used:

1. apply neighbourhood filter
2. apply category filter
3. neighbourhood filter is removed
4. apply neighbourhood filter again
5. both filters operate (but the page is empty).

i also think there is a bug in the neighbourhood reset link where you want to remove the `neighbourhood_name=Moss+Side` part but you are only removing the arg value from the URL (so e.g. `http://manchester.lvh.me:3000/partners?Moss+Side`).

r-ferrier:
> There seems to be a bug (or undefined behaviour) when combinations of category filter and neighbourhood filter are used:
> 
> 1. apply neighbourhood filter
> 2. apply category filter
> 3. neighbourhood filter is removed
> 4. apply neighbourhood filter again
> 5. both filters operate (but the page is empty).
> 
> i also think there is a bug in the neighbourhood reset link where you want to remove the `neighbourhood_name=Moss+Side` part but you are only removing the arg value from the URL (so e.g. `http://manchester.lvh.me:3000/partners?Moss+Side`).

Argh yes. Ok I am going to try changing the behaviour so that both filters get evaluated every time either is submitted.

r-ferrier:
Ok for any still interested parties, I have refactored this again so that the filters are now living in one concern. 

If we want to add more filters at any point we'll probably want to start breaking out some of this logic again but for now, this is much easier to understand, because information from both filters is being assessed every time a change is made to either of them. Also makes the partners controller index much cleaner! 

Have rewritten the tests to assert this is working. One selenium test keeps crashing which I am fixing.

app/controllers/concerns/partner_filters.rb
```diff
@@ -1,30 +1,37 @@
 # frozen_string_literal: true
 
-class PartnerCategoryFilter
-  def initialize(site, params)
+class PartnerFilters
+  attr_reader :neighbourhoods, :neighbourhood_names
+
+  def initialize(current_site, neighbourhood_names, params)
+    @current_neighbourhood_name = params[:neighbourhood_name]
```
ivan-kocienski-gfsc, line 7: is there a reason to use the string of a neighbourhood and not an ID? this seems like a lot of extra work for something that is not in the A/C (?)

ivan-kocienski-gfsc (changes requested):
Lots of good stuff but a few points of concern

app/controllers/concerns/partner_filters.rb
```diff
@@ -65,4 +68,34 @@ def include_mode?
   def exclude_mode?
     @include_mode == 'exclude'
   end
+
+  def reset_categories(url)
```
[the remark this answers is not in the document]
r-ferrier: ok, will have a go at implementing

app/models/partner.rb
```diff
@@ -311,10 +311,43 @@ def owned_neighbourhood_ids
...
+    else
+      address&.neighbourhood&.name_from_badge_zoom(badge_zoom_level)
+    end
+  end
+
   def self.fuzzy_find_by_location(components)
     Partner.find_by('lower(name) IN (?)', components.map(&:downcase))
   end
 
+  def self.neighbourhood_names_for_site(current_site, badge_zoom_level)
```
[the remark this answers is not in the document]
r-ferrier: you're probably right but at this stage they are going to stay here!

app/controllers/concerns/partner_filters.rb
r-ferrier: There absolutely is - https://github.com/geeksforsocialchange/PlaceCal/pull/1999#issuecomment-1729665685 - also the way these names are assigned is way more complicated than you would think, which is why I moved all the name logic into the partner model, and decided to just pass the names themselves

app/models/partner.rb
```diff
@@ -311,10 +311,43 @@ def owned_neighbourhood_ids
...
+      partners.each do |partner|
+        name = partner.neighbourhood_name_for_site(badge_zoom_level)
+        partner_names << name if name
+      end
+    partner_names.uniq.sort
+  end
+
+  def self.for_neighbourhood_name_filter(query = Partner, badge_zoom_level, neighbourhood_name)
+    partners_with_name = []
+    all_partner_neighbourhoods =
```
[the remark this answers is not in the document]
r-ferrier: I ended up sticking to each here but only ever passing in Partner.all, and removing the redundant return values. Otherwise every test was breaking because for whatever 'unsolvable today' reason when you get a partner back from Partner.all in the tests, it does not come with parents attached.

app/controllers/concerns/partner_filters.rb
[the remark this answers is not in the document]
r-ferrier: I also ran into a tests failing problem with passing in two hashes here - rubocop just refused to allow it. So my compromise is passing in the params as they are, and duplicating them within the concern.

app/controllers/concerns/partner_filters.rb
r-ferrier: But then we just have to work back the logic for how a name is assigned in here, instead of through the name method?

app/controllers/concerns/partner_filters.rb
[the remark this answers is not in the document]
ivan-kocienski-gfsc: i think i was being a little unclear maybe. don't use the url string reset methods and only use the hash based `reset_category_params` methods.

then change all the reset links in the views to be:
```ruby
<% if filter.category_active? %>
<%= link_to 'Reset', url_for(filter.reset_category_params), class: "btn size-patch", role: "button" %>
<% end %>
<% if filters.neighbourhood_active? %>
<%= link_to 'Reset', url_for(filter.reset_neighbourhood_params), class: "btn size-patch", role: "button" %>
<% end %>
```

ivan-kocienski-gfsc (approved):

app/controllers/concerns/partner_filters.rb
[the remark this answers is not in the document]
r-ferrier: ah I didn't misunderstand this, I just forgot to implement those methods. They of course won't work without the permitted params, which I can't include without breaking the tests, so for now I am leaving this with the regex reset which I agree is not a perfect long term solution but as it's only ever dealing with filters on this one page is hopefully fine for now

PR #2112, opened 2023-10-30
Fix flaky system test

kimadactyl:
I think this fixes the cause of at least the last log page worth of flaky system tests - fingers crossed

### DIFF
--- a/test/support/select2_helpers.rb
+++ b/test/support/select2_helpers.rb
@@ -22,7 +22,7 @@ module Select2Helpers
 
   def assert_select2_single(option, node)
     await_select2 10
-    within :xpath, node.path do
+    within :xpath, node.path, wait: 10 do
       assert_selector '.select2-selection__rendered', text: option
     end
   end


aaaaargZombies:
hmm I remember Ivan setting a longer default wait time so I have questions
- Does this override the default, reducing the max wait to 10 from 60?
- does this not override the default and is redundant?
- is it working because the default isn't being applied and that's where the fix should be?

https://github.com/geeksforsocialchange/PlaceCal/pull/1573/files

kimadactyl:
> Does this override the default, reducing the max wait to 10 from 60?

I don't think so - `within` is a capybara method. It currently has no wait time. So I think this is just giving it time to properly render the _contents_ of the select2 area after the functionality is loaded. This is all intuited from reading the docs tho but I figured worth a try.

aaaaargZombies:
> Of course there could be a default `within` wait time and I just got lucky rerunning the tests a couple times?

I think that's the case but I'm not sure how to verify it. README says the default is 2s  but you can up it like Ivan did.

https://github.com/teamcapybara/capybara/tree/2.5_stable#asynchronous-javascript-ajax-and-friends

What Ivan probably should have done was clean up the other `wait` arguments when he added that

kimadactyl:
Screenshot 2023-10-31 at 13 35 29

grrrrrr ok maybe this is not in fact, it

it is very consistently this function tho that is failing

aaaaargZombies:
yes frustrating 

- I think we need to see IF the global default is being respected
- if it is why are some things taking >60s to load :grimacing:

kimadactyl:
I don't think its taking 60s to load - the failed tests fail in the same time the passing ones do. A bit weird. I opened an issue for it: https://github.com/geeksforsocialchange/PlaceCal/issues/2116

PR #2114, opened 2023-10-31
Neighbourhood admins can edit partners with service areas in their neighbourhood

### LINKED ISSUE #2106: CRUD any partners outside Norwich

r-ferrier:
Fixes #2106 

## Description

- Neighbourhood admins (NAs) can now edit partners who appear on their partner index because they have set their service area to be within a neighbourhood belonging to the NA
- An NA is effectively also a partner admin (PA) for every partner in their neighbourhood, with one small difference - they cannot set a partners address or service areas to be outside of their neighbourhood. This is to prevent them from being able to accidentally remove partners from their purview.
- PAs can still set their address or neighbourhood to anywhere, so if an NA is also a PA of one of their partners, they can set this address to anywhere, and the partner should still be visible to them in their dashboard.


app/models/user.rb
```diff
@@ -76,7 +76,17 @@ def owned_neighbourhood_ids
   end
 
   def admin_for_partner?(partner_id)
-    partners.pluck(:id).include? partner_id
+    return true if partners.pluck(:id).include? partner_id
```
kimadactyl, line 79:
```suggestion
    partners.pluck(:id).include? partner_id
```

app/models/user.rb
```diff
@@ -76,7 +76,17 @@ def owned_neighbourhood_ids
...
+  end
+
+  def neighbourhood_admin_for_partner?(partner_id)
+    return true if
```
kimadactyl, line 83: don't think you need this `return true if` block, it should work without?

app/policies/partner_policy.rb
```diff
@@ -20,7 +20,7 @@ def new?
   def update?
     return true if user.root?
```
kimadactyl, line 21: Can replace this whole function with

`user.root? OR user.admin_for_partner?(record.id) OR user.neighbourhood_admin_for_partner?(record.id)`

kimadactyl:
Just looking at the ticket, do you think the validation error is something we should give more help for? Like we know there's edge cases this might happen, should we put a step there for what to do next i.e. contact support@placecal.org or something if you think this is in error? A bit wary of people bouncing off this.

r-ferrier:
> Just looking at the ticket, do you think the validation error is something we should give more help for? Like we know there's edge cases this might happen, should we put a step there for what to do next i.e. contact [support@placecal.org](mailto:support@placecal.org) or something if you think this is in error? A bit wary of people bouncing off this.

Ah the error validation is to show that this was working as it should - but then I found this edge case with partners with service areas in your neighbourhood but addresses elsewhere. I think the message is clear!

app/policies/partner_policy.rb
r-ferrier: for some reason this breaks without the return true statements so I'm keeping them in. I do also like how clear it makes the hierarchy so no bad thing!

aaaaargZombies:
## review testing

as hulme neighbourhood_admin

- [ ] Neighbourhood admins can update/destroy partners in their neighbourhood or partners with service areas in their neighbourhood
    - [x] update partner with addreess in hulme
    - [x] update partner with service area in hulme
    - [ ] delete partner with addreess in hulme
    ```
    ActiveRecord::InvalidForeignKey at /partners/work-for-change
PG::ForeignKeyViolation: ERROR:  update or delete on table "partners" violates foreign key constraint "fk_rails_44ff563791" on table "events"
DETAIL:  Key (id)=(10) is still referenced from table "events".
    ```
    - [ ] delete partner with service area in hulme
    missing delete button (maybe this is fine - should you be able to delete them, maybe just remove from your service area. though they can just come back in.... )
admin lvh me_3000_partners_geeks-for-social-change_edit


- [x] Neighbourhood admins cannot view/update/destroy partners outside of their neighbourhood whose service areas are also outside the neighbourhood
- [x] Neighbourhood admins can create new partners within their neighbourhoods or partners with service areas in their neighbourhood
    -[x] create with address in hulme
    -[x] create with servicearea in hulme no address
    -[x] create with servicearea in hulme address in different neighbourhood

- [x] Neighbourhood admins cannot create new partners outside of their neighbourhoods whose service areas are also outside the neighbourhood
- [ ] Neighbourhood admins cannot edit a partner to change its address and service area to be outside of their neighbourhood - at least one of these must be in their neighbourhood

problems

- I can delete a service area for a partner with no address losing contact wit it
- I can change the address to be out of my nieghbourhood then delete a service area for a partner losing contact wit it
- If I try to delete an address by removing all fields it successfully saves but the address is re-instated (should this fail? maybe out of scope)


- [x] A neighbourhood admin who is also a partner admin of a partner in their neighbourhood can change its address and service area to be outside of their neighbourhood

aaaaargZombies (approved):
as discussed in discord, approving to allow bits that work to be merged in, sharding off the smaller issues.

PR #2120, opened 2023-11-02
Neighbourhood admins can see all, edit all, and delete some users within their neighbourhood

### LINKED ISSUE #2109: See users who are partner_admin of partners in Norwich

r-ferrier:
Fixes #2109 

## Description

- Makes it possible for neighbourhood admins (NAs) to see all of the users who admin for partners in their neighbourhood in the user index
- NAs can click to view and edit any of these users, including a list of all partners they admin for
- NAs can add or remove partners within their neighbourhood to the list of partners a user admins for
- They cannot add or remove partners that have no relationship to their neighbourhood

- NAs can see the option to delete a user when they have no partners attached with any relationship to any other neighbourhoods.

## Notes

- It is possible to remove all of the partners in the NA's neighbourhood from a user, thus 'deleting' the user from the NAs world. This user is not deleted but is detached from the user. We decided this was preferable to arbitrarily deleting users (too much risk of deleting users who are active in other neighbourhoods or partnerships) but there are going to be occasions where someone slips through the net. I've implemented an alert that will pop up and force you to confirm you're happy for the user to be removed from your view if you do this, and explain how to keep the user in your neighbourhood, so NAs will be made aware of what is happening, and we thought it better to see how this plays out in reality?

- validation for removing partners from users is all handled by a js controller - this is so that the feedback could be given in the form. I did this because I didn't want the user to get the feedback after submitting - I wanted them to be able to respond in real time - and none of these changes actually violate the db rules & may very well be reverted again as we continue to get a feel for what this user role should look like.

- You can only delete a user when it has no connection to any other neighbourhood - this is quite strict but felt important to prevent neighbourhood admins from being  able to go rogue!


app/javascript/controllers/user_partners_controller.js
```diff
@@ -0,0 +1,52 @@
...
+				selectedValues.includes(x)
+			);
+
+			if (
+				selectedPermittedValues.length <= 1 &&
+				permittedValues.includes(Number(event.params.args.data.id))
+			) {
+				if (
+					!confirm(
+						"Removing this partner will remove this user from your neighbourhood.\n\n Are you sure you want to remove it? \n\n  If you want to keep the user in the neighbourhood but still remove this partner from them, you'll need to make them an admin for another partner in your neighbourhood."
```
aaaaargZombies, line 31:
```suggestion
"Removing this partner will remove this user from your neighbourhood.\n\nAre you sure you want to remove it? \n\nIf you want to keep the user in the neighbourhood but still remove this partner from them, you'll need to make them an admin for another partner in your neighbourhood."
```
minor formatting thing

app/javascript/controllers/user_partners_controller.js
```diff
@@ -0,0 +1,52 @@
...
+				if (option.selected) {
+					return [...accumulator, Number(option.value || option.text)];
+				}
+				return accumulator;
+			}, []);
+		};
+
+		$(this.element).select2();
+
+		$(this.element).on("select2:unselecting", (event) => {
```
aaaaargZombies, line 18: smart

aaaaargZombies (approved):
worked well for me, one minor suggestion.

I notice that the JS is untested but the system tests seem to often be causing as much pain as they are preventing. So I will just mention it and leave it up to you.

r-ferrier:
> worked well for me, one minor suggestion.
> 
> I notice that the JS is untested but the system tests seem to often be causing as much pain as they are preventing. So I will just mention it and leave it up to you.

I decided not to test the JS - I tried to keep as much of the logic out of it as I could and made sure I tested all of those bits instead. This decision was mainly because our system tests are flaky as hell but also because it's entirely separate and related only to this one action, so it would be extremely difficult to accidentally change behaviour here & you would end up having to rewrite the tests again to support any changes you made.

PR #2121, opened 2023-11-03
neighbourhood_admins can create users in their neighbouhood

### LINKED ISSUE #2113: neighbourhood_admin can create new user IF they are assigned to a partner in their neighbourhood
## ac

- [ ] only root can create new user without a partner
- [ ] new user must be assigned to a partner to create
- [ ] neighbourhood_admin can only assign partners to users in their neighbourhood

aaaaargZombies:
fixes #2113 

- NA can create a user if the user has a partner in the neighbourhood of the admin
- Root user can create user with no partner

Reasoning.

- NA only has access to partners in their neighbourhood, so any partner IDs in the params must come from that neighbourhood
- if you select no partners you get `user: { partner_ids: [''] }` so we have to sanitize before checking empty

thoughts

- we could strengthen the check by going though the partner_ids and matching to the current users neighborhood to the partners neighbourhood but I think if we start letting the neighbourhood_admin access more partners then we should be rethinking permissions generally.


app/controllers/admin/users_controller.rb
```diff
@@ -124,5 +125,16 @@ def update_user_profile
         current_user.update_without_password(profile_params)
       end
     end
+
+    def validate_neighbourhood_relation
+      # we only allow the neighbourhood_admin to assign Partners from their own Neighbourhood so any ids here are proof of a relationship
+      return if current_user.root?
+      return unless current_user.neighbourhood_admin? && params[:user][:partner_ids].reject { |id| id == '' }.empty?
```
r-ferrier, line 132: Nice! Simple and effective

r-ferrier (approved):
Works great and is a pleasingly simple check!

PR #2127, opened 2023-11-07
Rollbar exception filter

### LINKED ISSUE #2126: Filter runtime exceptions from being logged to rollbar
## Description

Some exceptions that get logged to rollbar are quite numerous and not critical, investigate a way of configuring rollbar in application so some exceptions (like not finding a route to an asset) don't needlessly add noise to our mission critical logging infrastructure.

- [x] Look through rollbar and identify exceptions that can be ignored
- [x] Investigate means of configuring app to skip logging those exceptions

## Notes
- rollbar config docs: https://docs.rollbar.com/docs/gem-configuration-reference
- example exception to ignore: https://app.rollbar.com/a/gfsc/fix/item/placecal-production/106

ivan-kocienski-gfsc:
Fixes #2126 

Stop rails from logging really common routing errors to rollbar

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require Rails.root.join('lib/rollbar_exception_filter')
+
 Rollbar.configure do |config|
   # Without configuration, Rollbar is enabled in all environments.
   # To disable in specific environments, set config.enabled=false.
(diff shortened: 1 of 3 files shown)


r-ferrier (approved):
I think this looks good! Obvs impossible to properly test until we get it out there so let's merge it in and see what happens.

One question I have is did you trace back where the bad paths were coming from? Is this from missing assets/broken links on placecal instances?

ivan-kocienski-gfsc:
@r-ferrier i pulled those URLs out of rollbar issues (perhaps i could have worded that better in the helper/test)

PR #2131, opened 2023-11-08
Allow deletion of a partner by a neighbourhood admin when they are the only possible Neighbourhood admin for that partner

### LINKED ISSUE #2130: If an NA can see a delete button on a partner, they can delete that partner
## Description

A user can delete a partner IF:

- [ ] They are a partner admin for that partner
- [ ] They are a root user
- [ ] They are a neighbourhood admin for the neighbourhood of the partners' address and the partner has no service areas outside the neighbourhood admin's remit
- [ ]  They are a neighbourhood admin for one of the partners' service areas and the partner has no service areas or address outside the neighbourhood admin's remit

r-ferrier:
Closes #2130 

## Description

- Lets neighbourhood admins delete partners, but ONLY if those partners exist only within their neighbourhoods
- Lets partner admins delete partners
- Adds some explanatory text for how to contact us if a user wants to delete a partner and can't

## Notes

I can certainly see a future scenario where this becomes a nuisance - for example, one of your partner admins has decided their service should operate outside of your remit and you want to shut them down. Neighbourhood admins can control this by making themselves a partner admin and then deleting the partner, but this is also open to abuse. It would be good to consider in the next architecture/permissions cycle what we think the best approach to handling these sorts of situations might be!

### DIFF
--- a/app/policies/partner_policy.rb
+++ b/app/policies/partner_policy.rb
@@ -31,9 +31,8 @@ class PartnerPolicy < ApplicationPolicy
 
   def destroy?
     return true if user.root?
-    return false unless user.neighbourhood_admin?
-
-    user.owned_neighbourhood_ids.include?(record.neighbourhood_id)
+    return true if user.admin_for_partner?(record.id)
+    return true if user.only_neighbourhood_admin_for_partner?(record.id)
   end
 
   def setup?
--- a/app/views/admin/partners/_form.html.erb
+++ b/app/views/admin/partners/_form.html.erb
@@ -155,10 +155,12 @@
   <div class="row">
     <div class="col-sm-12">
       <%= f.submit "Save Partner", class: "btn btn-primary btn-lg" %><br><br><br>
-      <% if policy(@partner).destroy? && !@partner.new_record? %>
-        <%= link_to "Delete Partner", @partner, method: :delete, class: "btn btn-danger btn-sm", id: 'destroy-partner'  %>
-      <% else %>
-        <p>You do not have permission to delete this partner, please contact <%= mail_to "info@placecal.org", "info@placecal.org" %> if you need to delete this partner.</p>
+      <% unless @partner.new_record? %>
+        <% if policy(@partner).destroy? %>
+          <%= link_to "Delete Partner", @partner, method: :delete, class: "btn btn-danger btn-sm", id: 'destroy-partner'  %>
+        <% else %>
+          <p>You do not have permission to delete this partner, please contact <%= mail_to "info@placecal.org", "info@placecal.org" %> if you need to delete this partner.</p>
+        <% end %>
       <% end %>
     </div>
   </div>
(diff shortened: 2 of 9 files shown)


app/policies/partner_policy.rb
```diff
@@ -31,9 +31,8 @@ def edit?
 
   def destroy?
     return true if user.root?
-    return false unless user.neighbourhood_admin?
-
-    user.owned_neighbourhood_ids.include?(record.neighbourhood_id)
+    return true if user.admin_for_partner?(record.id)
+    return true if user.only_neighbourhood_admin_for_partner?(record.id)
```
ivan-kocienski-gfsc, line 35: (you could leave off the `return true if` part)

app/views/admin/partners/_form.html.erb
```diff
@@ -124,6 +124,44 @@
...
+
+  <hr>
+  <br>
+  <div class="row">
+    <div class="col-sm-12">
+      <%= f.submit "Save Partner", class: "btn btn-primary btn-lg" %><br><br><br>
+      <% if policy(@partner).destroy? && !@partner.new_record? %>
+        <%= link_to "Delete Partner", @partner, method: :delete, class: "btn btn-danger btn-sm", id: 'destroy-partner'  %>
+      <% else %>
+        <p>You do not have permission to delete this partner, please contact <%= mail_to "info@placecal.org", "info@placecal.org" %> if you need to delete this partner.</p>
```
ivan-kocienski-gfsc, line 161: uh- this text shows up even if its a new record. you probably want something like
```
<% if !@partner.new_record? %>
  <% if policy(@partner).destroy? 
    <%= link_to "Delete Partner", @partner, method: :delete, class: "btn btn-danger btn-sm", id: 'destroy-partner'  %>
  <% else %>
    <p>You do not have permission to delete this partner, please contact <%= mail_to "info@placecal.org", "info@placecal.org" %> if you need to delete this partner.</p>
  <% end %>
<% end %>
```

ivan-kocienski-gfsc (changes requested):
One tiny bit of confusing text in the UI

app/policies/partner_policy.rb
r-ferrier: have had weird issues with another policy where not returning true explicitly causes it to break so I'm going to leave this in? Also it kind of helps my brain to just see the list of return statements!

app/views/admin/partners/_form.html.erb
r-ferrier: ohhhhhh it's the else clause. doh.

app/policies/partner_policy.rb
ivan-kocienski-gfsc: oh i do like a good bit of symmetry as well, hehehe

ivan-kocienski-gfsc (approved):
Looks good to go

PR #2135, opened 2023-11-10
All neighbourhoods on partners are shown, editing these is limited for neighbourhood admins

### LINKED ISSUE #2123: [Bug]: neighborhood_admin service area display inaccurate
## Description

only neighborhoods owned by a NA (nieghbourhood_admin)  can be seen by the NA.

If a partner or root user assigns a service area outside of this scope it displays the correct number of service areas to the NA but not the correct neighbourhood

root users view

image

NA view

image

Good news, the NA cannot remove the out of bounds neighbourhood.

## Steps to reproduce

1. What you did first
1. What you did next

## Platform (if relevant)

r-ferrier:
Fixes #2123 

## Description

What a bug! Neighbourhood admins used to see an incorrect list of service areas for partners - it would replace actual neighbourhoods with neighbourhoods owned only by the admin.

They can now see all of the neighbourhoods, but they cannot remove or modify neighbourhoods they do not admin for. They CAN add neighbourhoods they do not admin for, but only ones that are already owned by the partner, ie: no change. This does cause the validation to fail, preventing a partner save, but they can refresh the page and everything is fine. I will raise a separate ticket to fix this in a future iteration as it is annoying but does not actually break anything, and the likelihood that admins are going to be trying to add the same service area multiple times is really slim.

Partner admins and root users can still see and modify every service area.

This logic required a change to the input, but since the actual logic is handled in a new user method and the controller, testing is just based on these methods.

I've also done a tiny bit of refactoring on a form whose layout was hurting my eyes, and removed an unused method from the partners controller.

### DIFF
--- a/app/controllers/admin/partners_controller.rb
+++ b/app/controllers/admin/partners_controller.rb
@@ -6,7 +6,6 @@ module Admin
     before_action :set_partner, only: %i[show edit update destroy]
     before_action :set_tags, only: %i[new create edit]
     before_action :set_neighbourhoods, only: %i[new edit]
-    before_action :set_service_area_map_ids, only: %i[new edit]
 
     def index
       @partners = policy_scope(Partner).order({ updated_at: :desc }, :name).includes(:address)
@@ -54,7 +53,6 @@ module Admin
           format.html do
             flash.now[:danger] = 'Partner was not saved.'
             set_neighbourhoods
-            set_service_area_map_ids
             render :new, status: :unprocessable_entity
           end
           format.json { render json: @partner.errors, status: :unprocessable_entity }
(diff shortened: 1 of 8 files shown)


aaaaargZombies:
I started to document the strange behaviour - this is confusing but doesn't let them break the permissions. so I think raising a seperate ticket is probably the right thing, this does move us towards a better solution.

I'll atatch the screen recording I made so it can be linked to the issue

https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/2a325b59-e33b-4f42-94d8-d34bb9f72d4e

aaaaargZombies (approved):
code looks fine, removing unused method good, refactor also good.

we are getting some very convoluted conditionals all over the shop but I don't think that's your fault, just the structure of the code.

As mentioned earlier I think that this works well on the happy path and breaks non-catastrophically if people take a wrong turn. 

A big improvement to before  when it was just always broken.

Issue #2128, opened 2023-11-07
NAs (and by extension partnership admins) can remove addresses from partners who have an address in their neighbourhood and a service area in their neighbourhood


r-ferrier:
Related
- #1914 
- #1219 

## Acceptance criteria

- [ ] As a root user I can remove a partner's address
- [ ] As a partner admin, I can remove a partner's address if they have a service area set
- [ ] As a neighbourhood admin or a partnership admin, I can remove a partner's address if they have a service area set, but if that service area is not in my scope, I receive a warning to confirm that the partner will no longer be visible to me if I proceed

## To be dealt with in a later issue
- [ ] Can users (who are not root or partner owners) remove addresses from partners when that address falls outside of their neighbourhood pool?
- [ ] If there is a calendar attached to the partner and its location strategy is set to 'Default', 'Room number' or 'Event override', the user is warned that their calendar will no longer have a location for their events and the location strategy is changed to 'None' (See also #2214 and #1219)

## Notes

The way addresses are assigned to partners means that once that record is present it cannot be removed (in the original specs partners MUST have an address and they shared addresses).

A solution may be to either
- if all the fields in the address part of the form are blank then delete the address record.
- put in a "remove address" button/action to explicitly delete the partners' address

r-ferrier:
This is definitely not possible. So, you can create a partner without an address, but even as root, you cannot remove an address once it is added. I think we need to tackle this in tandem with the importer issue mentioned above so that we don't introduce more bugs to normal workflows, so i'm changing the description of this issue to reflect that

r-ferrier:
blocked by #1219

ivan-kocienski-gfsc:
i just picked up #1219

r-ferrier:
After discussion this morning we think this is ready to be worked on again! Have put some notes in the description.

ivan-kocienski-gfsc:
Does that partner have any service areas? roots cannot clear an address if its the only location attached to that partner.

and yes this UI is confusing

r-ferrier:
Ah ok! That makes sense! I will close and verify the rest.

r-ferrier:
Verified

- As a root user I can remove a partner's address if they have a service area set

![Screenshot 2024-02-01 at 09 43 01](https://github.com/geeksforsocialchange/PlaceCal/assets/46496391/72aaf92d-0b4f-4295-b026-3b2ea841a930)

- As a partner admin, I can remove a partner's address if they have a service area set

[...]

PR #2136, opened 2023-11-13
Limit NA address power

### LINKED ISSUE #2119: neighbourhood_admins cannot assign addresses to partners outside their neighbourhood
In the same way that you cannot see serviceAreas outside of a neighbourhood you are the admin for.

admin lvh me_3000_partners_new (1)

You should not be able to assign an address that falls outside of the neighbourhood. This is prevented if there is no servicearea assigned but once the service area is in your neighbourhood you can start assigning addresses anywhere.

this can lead to two problems.
1. a neighbourhood_admin could change the address then remove the servicearea accidentally deleting it from their point of view
2. a neighbourhood_admin could start making partners using their servicearea tag but in any neighbourhood they like influencing/spamming a different neighbourhood outside their jurisdiction 

## AC

- [ ] a `neighbourhood_admin` cannot save a partners address outside of their own neighbourhood
- [ ] the above AC does not prevent a `neighbourhood_admin` from making other updates to a partner who has an address outside the neighbourhood but has opted in to the `neighbourhood_admin`s neighbourhood via the service area.

ivan-kocienski-gfsc:
Closes #2119 

Neighbourhood admins can only assign addresses to partners that fall within their neighbourhood pool

### DIFF
--- a/app/components/event/event_component.rb
+++ b/app/components/event/event_component.rb
@@ -43,6 +43,8 @@ class EventComponent < MountainView::Presenter
   end
 
   def date
+    return formatted_date(event.dtstart) if event.dtend.blank?
+
     if event.dtstart.to_date == event.dtend.to_date
       formatted_date(event.dtstart)
     else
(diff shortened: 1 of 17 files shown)


r-ferrier (approved):
I think this looks good. It meets acceptance criteria and the tests are all working well. Neighbourhood admins still can't remove addresses which is slightly limiting, but it's just about out of scope and we do have this ticketed (https://github.com/geeksforsocialchange/PlaceCal/issues/2128) , and I think this is a big improvement.

Issue #1707, opened 2023-02-23
Update to ruby 3.x


ivan-kocienski-gfsc:
Has to happen sooner or later.

kimadactyl:
Sure - but vote 'later' if it doesn't work in 20 mins!

katjam:
Closing as duplicate of #1786

PR #2156, opened 2024-01-08
Update to ruby v3.1.2

### LINKED ISSUE #1786: Ruby v2 nearing EOL (Dec 2023)
## Description
Upgrade to Ruby v3

## Acceptance Criteria

- [ ] Ruby version 3 or higher is specified in `.ruby_version` and `Gemfile`. Running `ruby --version` in PlaceCal directory returns version indicating such.

## Motivation
image

ivan-kocienski-gfsc:
Closes #1786, #1707 

Updates ruby, leaves the gems at their locked versions.

Minor tweaks to tests and rubocop config


r-ferrier (changes requested):
Unfortunately I'm running into a couple of problems with this!

The main issue is that it is crashing when I run this locally and try to visit the site. This could very much be a me problem - it looks like it might be the same weird binaries thing that was happening when we set up GI matrix - but I'll need to investigate and make sure that's the case before we can safely release this.

I'm also seeing a deprecation warning in the console on running the site:

'Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.'

If you could fix the deprecation I will try to work out what's happening that's preventing me from being able to run this locally and see if we need to add in any changes to the gemfile like we did on gi matrix.

ivan-kocienski-gfsc:
That sounds painful. Could be because we are using the new ruby with all the old gems locked down? so an older binary probably needs updating. I will take a look and see if I can find a candidate.

> If you could fix the deprecation I will try to work out what's happening that's preventing me from being able to run this locally and see if we need to add in any changes to the gemfile like we did on gi matrix.

I am hoping this will be fixed when we update the rest of the gems.

aaaaargZombies:
working for me but I do get this warning on tests

```
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
```

Is this something that needs addressing? I don't get it on ruby 2

ivan-kocienski-gfsc:
@aaaaargZombies looks like the problem is with a few gems interacting with the not-super-current ruby (3.1.2).

this feels like it could turn into a complicated mess that drags on when what we need to focus on is not having an EOL ruby.

i'd rather get this work on main today and fix all the numerous gem problems as a next step than try to craft the entire solution in one PR.

aaaaargZombies (approved):
ok let's give it a go and hope this is resolved soon

r-ferrier (approved):
bundler update has fixed deprecation warnings for me

PR #2180, opened 2024-01-16
Rename badly named column in model

### LINKED ISSUE #2179: Remove unused OrganisationRelationship model / table / code
## Description

In preparation for #1110: this model has attributes that are flagged as unsafe (specifically `ActiveRecord::DangerousAttributeError: object_id is defined by Active Record. Check to make sure that you don't have an attribute or method with the same name.`)

This model appears unused and unmaintained (in code) and the production table only has 10 entries in it, so the developers decided that it should be removed.

ivan-kocienski-gfsc:
Closes #2179 

That conflicts with a rails method in upcoming rails 7.

### DIFF
--- a/app/models/organisation_relationship.rb
+++ b/app/models/organisation_relationship.rb
@@ -1,10 +1,19 @@
 # frozen_string_literal: true
 
 class OrganisationRelationship < ApplicationRecord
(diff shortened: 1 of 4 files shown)


aaaaargZombies (approved):
LGTM

PR #2187, opened 2024-01-17
Added gems to gemfile

ivan-kocienski-gfsc:
To fix CI build problem.

Also fixing missing test assets


r-ferrier (approved):
Looks good, this is as per https://stackoverflow.com/questions/70500220/rails-7-ruby-3-1-loaderror-cannot-load-such-file-net-smtp

Once we're at or above rails 7.0.1 this can be reverted.

PR #2193, opened 2024-01-18
fix: missing sites from partners @sites

### LINKED ISSUE #2151: [Bug]: Error message on Partner edit page reporting that is is not appearing on any sites, but it is
## Description

Came from Rach who was updating a partner.

Also research #1722 which might be part of the same?

## Steps to reproduce

1. Edit parner: 

Screenshot 2023-12-08 at 15 03 39

aaaaargZombies:
fixes #2151 

We were checking the partners neighborhood ids against the sites neighborhood ids but this failed when the site had a neighborhood that encompassed other smaller neighborhoods. From the partner we would have needed to expand all the way to the root neighborhood and check if any sites existed.

It feels like there's probably some more efficient way of doing this,  we bounce back and forth between the sites and partners rebuilding the entire world just to scope it back down again. So open to an obvious answer I'm missing but also didn't want to start down a  rabbithole.

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -159,13 +159,14 @@ class Site < ApplicationRecord
     # @param [Partner]
     # @return [ActiveRecord::Relation<Site>]
     def sites_that_contain_partner(partner)
-      neighbourhood_ids = partner.owned_neighbourhood_ids
-      tag_ids = partner.partner_tags.pluck(:tag_id)
-
-      Site.left_outer_joins(:sites_tag, :sites_neighbourhoods)
-          .group(:id)
-          .where('neighbourhood_id in (?) or tag_id in (?)',
-                 neighbourhood_ids, tag_ids)
+      sites = Site.all.order(:name)
+      site_partners = []
+      sites.each do |site|
+        partner_ids = Partner.for_site(site).pluck(:id)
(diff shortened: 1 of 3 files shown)


app/models/site.rb
```diff
@@ -159,13 +159,14 @@ def find_by_request(request)
...
     # @return [ActiveRecord::Relation<Site>]
     def sites_that_contain_partner(partner)
-      neighbourhood_ids = partner.owned_neighbourhood_ids
-      tag_ids = partner.partner_tags.pluck(:tag_id)
-
-      Site.left_outer_joins(:sites_tag, :sites_neighbourhoods)
-          .group(:id)
-          .where('neighbourhood_id in (?) or tag_id in (?)',
-                 neighbourhood_ids, tag_ids)
+      sites = Site.all.order(:name)
```
ivan-kocienski-gfsc, line 162: I think your problem may be that the original code also dealt with tags as well as neighbourhoods.

aaaaargZombies: should that matter because here I compare a partner to a list of partners instead of inferring the inclusion based on shared attributes?

r-ferrier (approved):
Looks good. I get your point about there being a potentially dryer or faster way of doing this, but I do think this is easy to understand from the code which is always a plus!

PR #2194, opened 2024-01-18
Added explanation text to calendar form

### LINKED ISSUE #2069: Calendar admin page does not specify if contact info is public or private
## User story

As a Partner Admin or Organiser, I need to be able to know if contact information I'm adding to a partner's calendar will be public or private.

## Acceptance criteria

- [ ] The page for add/edit calendar makes clear where the information is shared

At present, the page to add or edit a calendar ( https://admin.placecal.org/calendars/*/edit  <-- wildcard for calendar number) has a section for contact details, but does not specify whether those details are private (i.e. a contact specifically for PlaceCal/GFSC internal use) or public (will be published, anyone interacting with PlaceCal can see them).

Asking a developer about this in a meeting yielded "I'm not sure". The page should make it clear so none of us get confused ever again.

## Implementation plan

Screenshot 2023-09-20 at 15 54 20

ivan-kocienski-gfsc:
Closes #2069 

Warning text at the bottom of the admin calendar form explaining how the public contact info fields are used on the public site.


aaaaargZombies (changes requested):
apparently `big` is discouraged

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/big

I wonder if we could link to an example on a live site?  Is there a live partner we could link to as an example. "at the bottom of the page" is a bit confusing unless you know it's an event page.

ivan-kocienski-gfsc:
Hmm thanks for the spot! Hard to remember sometimes which tags are new and which ones are obsolete.

I know what you mean about that but I don't think its a good idea to link to a live event as it may get deleted? Maybe a screenshot with an arrow pointing it out?

Something like:

Screenshot_2024-01-18_21-25-11

aaaaargZombies:
image

Maybe we should just let them know it's public like on the partner page.

ivan-kocienski-gfsc:
Are you talking about calendars and events or partners?

The partner page is not part of this ticket and the area is titled "Public Contact" with the description of "This is the information that will be shown to the public".

But I did have an idea about how to solve the "see this example" link you suggested (other than that godawful graphic).

aaaaargZombies:
> Are you talking about calendars and events or partners?

Sorry I was just looking at how we have dealt with this problem elsewhere in the site, screenshot is just an example not additional stuff for the ticket.

aaaaargZombies (approved):
That should clear up any confusion :heavy_check_mark:

PR #2196, opened 2024-01-18
Only show tagged partners to partnership admins

### LINKED ISSUE #2168: Partnership admins can only see partners with a service area or address in their neighbourhood scope, that also have the tag they are partnership admin for
## Acceptance criteria

- [ ] Partnership admins can see partners with a service area or address in their neighbourhood scope, that also have the tag they are partnership admin for
- [ ] Partnership admins cannot see partners that are only in their neighbourhood scope
- [ ] Partnership admins cannot see partners that are only in their tag scope
- [ ] Partnership admin can see partners they are directly  partner admin for

r-ferrier:
## Description

fixes #2168

This PR begins the work to reinstate partnership admins by making the only partners visible to them ones that have the tag that they own, and are in neighbourhoods within their scope. 

I began by defining a generic partnership admin as a tag admin who only owns partnership tags. Since it is technically impossible for a user to own any other type of tag, this is a bit belt and braces, but I thought it made it clearer what it meant to be a partnership admin. I then removed all references to tag admins and replaced them with references to the partnership admin.

I tweaked how we define a `neighbourhood_admin_for_partner` class method on the user to only return true for users without partnership tags, and added a `partnership_admin_for_partner` class method that will only return true for users who own a tag shared with that partner, and manage neighbourhoods that partner is in. I decided to take this approach so that we could leave the neighbourhood admin logic completely untouched, safe in the knowledge that all we were doing was removing anyone with a partnership_tag from this bracket.

I also added a new clause to the partner policy scope to replicate this logic, revived some broken tests we had for this, and added some new partnership admin tests to cover all of the above!

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -120,8 +120,8 @@ class User < ApplicationRecord
     partners.any?
   end
 
-  def tag_admin?
-    tags.any?
+  def partnership_admin?
+    tags.any? { |tag| tag[:type] == 'Partnership' }
   end
 
   def admin_roles
@@ -131,7 +131,7 @@ class User < ApplicationRecord
     types << 'editor' if editor?
     types << 'neighbourhood_admin' if neighbourhood_admin?
     types << 'partner_admin' if partner_admin?
-    types << 'partnership_admin' if tag_admin?
+    types << 'partnership_admin' if partnership_admin?
     types << 'site_admin' if site_admin?
 
     types.join(', ')
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -79,14 +79,16 @@ class User < ApplicationRecord
     partners.pluck(:id).include? partner_id
   end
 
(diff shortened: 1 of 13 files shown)


app/models/user.rb
```diff
@@ -155,6 +157,23 @@ def assigned_to_postcode?(postcode)
...
+  def partner_in_neighbourhood_scope?(partner_id)
+    neighbourhood_admin? &&
+      (
+        owned_neighbourhood_ids & (
+          Partner.find_by(id: partner_id).owned_neighbourhood_ids
+        )
+      ).any?
+  end
+
+  def partner_in_partnership_scope?(partner_id)
```
aaaaargZombies, line 169: seems like a really good idea to split these, if at some point in the future we just have partnership admins who aren't neighborhood admins this will be less painful

aaaaargZombies (approved):
nice, works well and I still get the partners I am an admin for who aren't tagged.

PR #2201, opened 2024-01-20
1791 enforce full domain in site settings

### LINKED ISSUE #1791: [Bug]: Enforce full domain in site settings
## Description

Needs to have https:// in front of it - otherwise gets generated as a relative link.

## Acceptance Criteria

- [ ] Domain needs to include protocol in the site settings

## Implementation Notes
- Whatever is easiest - automatically add https:// if it isn't there on save or give the user a hint and prevent saving if it isn't there.

GOOD:
image

BAD:
image

aaaaargZombies:
fixes #1791 

A small fix that required a lot of test wrangling

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -36,6 +36,7 @@ class Site < ApplicationRecord
   validates :name, :slug, :domain, presence: true
   validates :place_name unless :default_site?
   validates :hero_text, length: { maximum: 120 }
+  validates :domain, format: { with: %r{\Ahttps://[^\s,]+\z}, message: 'A domain must start with "https://"' }
 
   scope :published, -> { where(is_published: true) }
 
--- a/test/controllers/news_controller_test.rb
+++ b/test/controllers/news_controller_test.rb
@@ -22,15 +22,9 @@ class NewsControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should get index subdomain' do
-    get news_index_url(subdomain: @site.domain)
+    get news_index_url(subdomain: @site.slug)
     assert_response :success
     assert_select '.articles__article-card', 5
-
-    # counts
-    # assert_select 'p', { text: 'Found 5 articles.' }
-
-    # pagination
-    # assert_select 'p', { text: 'No more news items' }
     assert_select 'p', { count: 0, text: 'Older news items' }
   end
 
@@ -49,7 +43,7 @@ class NewsControllerTest < ActionDispatch::IntegrationTest
       art.update! published_at: @epoch - n
     end
 
-    get news_index_url(subdomain: @site.domain)
+    get news_index_url(subdomain: @site.slug)
     assert_response :success
 
     # this is capped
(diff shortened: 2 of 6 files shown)


app/models/site.rb
```diff
@@ -36,6 +36,7 @@ class Site < ApplicationRecord
   validates :name, :slug, :domain, presence: true
   validates :place_name unless :default_site?
   validates :hero_text, length: { maximum: 120 }
+  validates :domain, format: { with: %r{\Ahttps://[^\s,]+\z}, message: 'A domain must start with "https://"' }
```
ivan-kocienski-gfsc, line 39: We do also support `webcal://`.

I would probably just validate looking for `https://` and then anything after that is probably okay. Also there was some drama about supporting plain `http`.

ivan-kocienski-gfsc:
Looking at the PR and the issue I think the issue is incorrect: the field is called "domain" and is intended to be just a URL domain, which doesn't include the http bit or anything after the query string.

r-ferrier (approved):
I think this looks good, I prefer this route of validating on submit because it makes it clear what needs changing with the least visual clutter and only needs to fire if someone gets it wrong.

Annoying about the tests!

ivan-kocienski-gfsc (changes requested):
wait i think i found a bug with this.

i will write it up. just stopping this being merged quickly

aaaaargZombies:
change domain to url so clear what it's being used for. do migration.

ivan-kocienski-gfsc (changes requested):
As per our discussions

app/models/site.rb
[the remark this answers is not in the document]
ivan-kocienski-gfsc: Maybe also rename the field to URL (or site_url if that is too generic a name)

db/migrate/20240122152631_change_domain_to_url.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ChangeDomainToUrl < ActiveRecord::Migration[6.1]
+  def self.up
+    rename_column :sites, :domain, :url
```
ivan-kocienski-gfsc, line 5: you can actually just use the one method here `self.change` and rails will figure out the inverse of the rename depending if you are migrating up or down on its own

ivan-kocienski-gfsc (changes requested):
Unfortunately got some small things, one of them a showstopper.


You missed a domain->url switch in the `PartnershipCardComponent`

Screenshot_2024-01-23_09-09-34

(Might be worth writing a test for this page?)

aaaaargZombies:
@ivan-kocienski-gfsc  hopefully this resolves those last issues

aaaaargZombies:
I should also say I rebuilt transdimension elm app locally with no issues, I looked at the data folder and I think we just pull in partners based on the (partnership)tag not any site rule, probably because it covers the whole uk so geographic info is redundant.

ivan-kocienski-gfsc (approved):
Looks great! good work here

Issue #2054, opened 2023-09-18
[Bug]: neighbourhood_admin users can press update button but cannot change anything on edit user page


r-ferrier:
## Description

Edit page or maybe just the update button should not be visible when a user is not able to make any updates

aaaaargZombies:
This issue may be out of date. Currently a NA can

- update which partners a user has
- delete some users, potential regression on #2104 

![2024-01-20-14:49:33-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/ab02ed31-c00c-4e58-9c27-1f9c37c502d4)

r-ferrier:
marking as duplicate, dealt with in #2203

PR #2203, opened 2024-01-20
only root users can delete users

### LINKED ISSUE #2104: Only root users can delete users
## Description

Only root users should be able to delete users.

## Acceptance criteria

- [ ] A root user can delete any user
- [ ] A neighbourhood admin cannot delete any users
- [ ] A partnership admin cannot delete any users
- [ ] A citizen, editor or site_admin cannot delete any users

## Plan

- Investigate code & UI and find out if anyone else currently can delete users
-  If they can, add in checks to user policy to only permit root to delete users
- If they can see any, remove links to delete options from views accessible by other users

aaaaargZombies:
see #2054

fixes #2104

Nice to delete that conditional, something like that always gives me a bad feeling. Also removed a contradictory test, not sure how that and the test below it were both passing. I guess the problem with having tests for a refactor is that when you change how the system works a passing test is no longer an indication of desired behavior.

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -37,13 +37,7 @@ class UserPolicy < ApplicationPolicy
   end
 
   def destroy?
-    user.root? ||
-      (user.neighbourhood_admin? &&
-      !record.root? &&
-      !record.neighbourhood_admin? &&
-      !record.tag_admin? &&
-      record.partner_admin? &&
-      all_user_partners_in_admin_neighbourhood?(record, user))
+    user.root?
   end
 
   def permitted_attributes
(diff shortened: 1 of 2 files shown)


r-ferrier:
Ooh I don't know about this one! What about . . .
https://github.com/geeksforsocialchange/PlaceCal/issues/2109

Maybe we need to discuss this one at the meeting this morning. Looks like we created to contradictory issues at around the same time and I'm actually not sure what the preferred behaviour is here.

r-ferrier (approved):
We discussed this and have decided to limit deletion to root users. This is because:
- admins run the risk of deleting users who have responsibilities outside of the admin's scope
- admins can remove users from their scope by taking any of the properties they admin for away from them and if a user has no other responsibilities, this effectively deactivates their account

We think once we add in a 'permissions handshake' type system, allowing users to request access to things, this type of scenario will make more sense.

PR #2205, opened 2024-01-22
Allow partnership admins to delete partners

### LINKED ISSUE #2177: Partnership admins can delete partners that ONLY have service areas/addresses within their neighbourhood scope and tags that they admin for
This is to prevent the deletion of partners still operating in other neighbourhoods/ partnerships.

## Acceptance criteria

- [x] A partnership admin can delete a partner whose service area, address, and partnership tag they admin for
- [x] A partnership admin cannot delete a partner whose service area they do not admin for
- [x] A partnership admin cannot delete a partner whose address they do not admin for
- [ ] A partnership admin cannot delete a partner whose partnership tag they do not admin for

ivan-kocienski-gfsc:
Closes #2177 

When deleting a partner the user must
- own all the neighbourhoods this partner is in
- own all the partnership tags this partner has


app/models/partner.rb
```diff
@@ -312,6 +312,28 @@ def owned_neighbourhood_ids
     neighbourhood_ids
   end
 
+  # returns true if all
+  #  - address in user neighbourhood set
+  #  - service areas in user neighbourhood set
+  #  - tags in users partnership tag set
+  def owned_exclusively_by?(user)
```
r-ferrier, line 319: This is a method that is really asking if a user is allowed to access a resource, and given that all the other methods in the scope (ie. `.admin_for_partner?`) reside in the User model, I think it would make a lot more sense if this did too? 

If you have a look at the existing method `def only_neighbourhood_admin_for_partner?` we could also add a simple `def only_partnership_admin_for_partner?` method which could be checked first. Then the logic hierarchy would be clearer - partner admins and roots get the most control, followed by partnership admins, followed by neighbourhood admins.

In that method you would need to then check if a user is a `partnership_admin`, and then run the logic you've implemented here.

You'll also need to modify `only_neighbourhood_admin_for_partner` to check first if the user is a `partnership_admin || neighbourhood_admin` and not just a neighbourhood admin.

r-ferrier (changes requested):
Sorry, I think this could cause some snags because of overlap with the `only_neighbourhood_admin_for_partner?` method. We're trying to keep the two roles separate in the system, so it would be great if the policy uses a similar method that makes it clearer it's just checking for other `partnership_admins`, with the next method checking for other `neighbourhood_admins`. Hope this all makes sense!

app/models/user.rb
```diff
@@ -92,14 +92,37 @@ def neighbourhood_admin_for_partner?(partner_id)
...
+    user_neighbourhoods = owned_neighbourhood_ids
+    user_tags = tags.pluck(:id)
+
+    a = Set.new(user_neighbourhoods)
+    b = Set.new(partner.owned_neighbourhood_ids)
+
+    return false unless a.superset?(b)
+
+    # not needed as we check we're a partnership admin?
+    # return false if user_tags.empty? || partner.partnerships.empty?
```
r-ferrier, line 117: can these comments come out?

app/models/user.rb
```diff
@@ -92,14 +92,37 @@ def neighbourhood_admin_for_partner?(partner_id)
...
 
+  def only_partnership_admin_for_partner?(partner_id)
+    return unless partnership_admin?
+
+    partner = Partner.find(partner_id)
+
+    user_neighbourhoods = owned_neighbourhood_ids
+    user_tags = tags.pluck(:id)
+
+    a = Set.new(user_neighbourhoods)
```
r-ferrier, line 111: would be clearer if these variables had semantic names, same for below, (also prefer not repeating variable names within the same scope) but that's just my preference for clarity, so fine if you want to leave as is

r-ferrier (approved):
Thanks for making those changes, it will be much clearer if (/when!) we have to change any of this in the future what's going on where. I've left a couple more comments for changes that I would make, but these are not dealbreakers if you'd rather not!

PR #2207, opened 2024-01-22
Fix the tag editing capabilities of everyone on a partner

### LINKED ISSUE #2174: Partnership admins can edit everything about a partner, except removing or adding neighbourhoods outside of their neighbourhood scope or removing tags they do not admin for
## Acceptance criteria

Of the partners they can see, PAs can:

- [ ] Edit all details belonging to the partner, including who admins for it
- [ ] #2128
- [ ] Remove service areas if the address is still within their scope
- [ ] change address to a different one within their scope
- [ ] add service areas within their scope
- [ ] add tags they admin for
- [ ] remove tags they admin for (but if they do this, they should be warned that they will no longer be able to see the partner)

r-ferrier:
fixes #2174

## Description

- ensures root users can still add/remove any tag
- ensures partner admins can still add any category or facility tag and remove any partnership tags
- allows partnership admins to edit tags without accidentally removing tags they can't see in the process
- prevents partnership admins from removing partnership tags that don't belong to them
- warns partnership admins if tag removal is likely to boot a partner out of their scope

## Implementation notes

I've used a bespoke select2 style controller for this again, like we did for controlling the partners that users admin for. This is mainly so there is instant and obvious feedback, and because we don't actually need to make these moves impossible at a database level, we just want to ensure admin users are fully aware of how certain actions will impact their reach. 

I've added tests for the new methods in the helper but not for the JS logic for the same reasons mentioned in the related PR.

(related PR [here](https://github.com/geeksforsocialchange/PlaceCal/pull/2120))

### DIFF
--- a/app/controllers/admin/partners_controller.rb
+++ b/app/controllers/admin/partners_controller.rb
@@ -6,6 +6,7 @@ module Admin
     before_action :set_partner, only: %i[show edit update destroy]
     before_action :set_tags, only: %i[new create edit]
     before_action :set_neighbourhoods, only: %i[new edit]
+    before_action :set_partner_tags_controller, only: %i[new edit]
 
     def index
       @partners = policy_scope(Partner).order({ updated_at: :desc }, :name).includes(:address)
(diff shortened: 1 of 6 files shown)


ivan-kocienski-gfsc (approved):
Looks nice, good work.

(Didn't know you could `def policy_scope` in tests like that)

PR #2209, opened 2024-01-23
Hide untagged partner admins from partnership admins

### LINKED ISSUE #2169: Partnership admins can only see users that admin for a partner they can view
## Acceptance criteria

- [ ] Partnership admins can only see users that admin for the partners they can view
- [ ] Partnership admins cannot see users who admin for partners that are only in their neighbourhood scope
- [ ] Partnership admins cannot see users who admin for partners that are only in their tag scope

r-ferrier:
Fixes #2169 

## Description

- Shows partnership admins only users that are partner admins for partners within their scope
- Adds some user policy tests to assert the scope is behaving as it should

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -130,7 +130,6 @@ class UserPolicy < ApplicationPolicy
 
       else
         user_neighbourhood_ids = user.owned_neighbourhood_ids
-
         scope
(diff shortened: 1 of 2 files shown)


ivan-kocienski-gfsc (approved):
Looks good

PR #2210, opened 2024-01-23
restrict partnership_admin access to calendars by neighbourhood AND partnership tag

### LINKED ISSUE #2170: Partnership admins can only view calendars that belong to partners they can view
## Acceptance criteria

- [ ] Partnership admins can only see calendars that belong to the partners they can view
- [ ] Partnership admins cannot see calendars that belong to partners that are only in their neighbourhood scope
- [ ] Partnership admins cannot see calendars that belong to partners that are only in their tag scope

aaaaargZombies:
fixes #2170 

- adjust scope to account for tags
- adjust edit policy to prevent partnership_admins accessing calendars directly via URL

### DIFF
--- a/app/policies/calendar_policy.rb
+++ b/app/policies/calendar_policy.rb
@@ -14,15 +14,12 @@ class CalendarPolicy < ApplicationPolicy
   end
 
   def edit?
-    index?
+    update?
   end
 
   def update?
     return true if user.root?
-    return true if user.partner_admin? && user.partner_ids.include?(record.partner_id)
-
-    # return true if user.neighbourhood_admin? && user.neighbourhoods.include?(record.address.neighbourhood)
-    index?
+    return true if scope.map(&:id).include? record.id
   end
 
   def import?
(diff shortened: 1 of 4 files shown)


app/policies/calendar_policy.rb
```diff
@@ -14,15 +14,12 @@ def new?
...
+    update?
   end
 
   def update?
     return true if user.root?
-    return true if user.partner_admin? && user.partner_ids.include?(record.partner_id)
-
-    # return true if user.neighbourhood_admin? && user.neighbourhoods.include?(record.address.neighbourhood)
-    index?
+    return true if scope.map(&:id).include? record.id
```
r-ferrier, line 22: oh this is so much simpler! Why don't we always do this!

app/policies/calendar_policy.rb
```diff
@@ -42,16 +39,32 @@ def resolve
       return scope.all if user.root?
       return scope.none if !user.partner_admin? && !user.neighbourhood_admin?
```
r-ferrier, line 40: don't you need to add in a `!user.partnership_admin?` check here too?

app/policies/calendar_policy.rb
```diff
@@ -42,16 +39,32 @@ def resolve
...
-                  'service_areas.neighbourhood_id in (:neighbourhood_ids) OR '\
-                  'calendars.partner_id IN (:partner_ids))',
-                  neighbourhood_ids: user.owned_neighbourhood_ids,
-                  partner_ids: user.partner_ids
-                )
-                .distinct
+      if user.partnership_admin?
+        user_partnership_tag_ids = user.tags.map(&:id)
+        partnership_calendars =
+          Calendar.left_joins(partner: %i[address service_areas partnerships])
```
r-ferrier, line 45: nice!

r-ferrier (changes requested):
One change (maybe) needed?

app/policies/calendar_policy.rb
aaaaargZombies: yes I think this is a good catch, all the conditionals are getting a bit hairy round here

r-ferrier (approved):

PR #2213, opened 2024-01-23
Add partner admins into scope

### LINKED ISSUE #2169: Partnership admins can only see users that admin for a partner they can view

### LINKED ISSUE #2212: [Bug]: partnership_admin user scope does not include users they also admin a partner with
## Description

It is possible that a partnership_admin may also be working as a partner_admin for an partner that has not tag or neighbourhood relationship to their partnership.

They should still be able to see users which they share a connection to through joint partner_admin of a partner.

eg

PA for transdim is also involved in nottingham bog trotters. They should be able to see fellow nottingham bog trotters admins even though they have no connection to the transdim.

## Steps to reproduce

1. What you did first
1. What you did next

## Platform (if relevant)

r-ferrier:
Fixes #2169 
Fixes #2212

## Description

Adds in a missing part of the puzzle - allowing partner admins to also see all the users that admin for their partners, and in a way that works in conjunction with other admin types, so for example, if you're a neighbourhood admin for manchester and also a partner admin for a gardening group based in Liverpool, you can see every user that admins for a partner based in manchester as well as every user that admins for your gardening group.

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -114,29 +114,35 @@ class UserPolicy < ApplicationPolicy
 
       elsif user.partnership_admin?
         user_neighbourhood_ids = user.owned_neighbourhood_ids
+        user_partner_ids = user.partners.map(&:id)
         user_partnership_tag_ids = user.tags.map(&:id)
 
         scope
           .left_joins(partners: %i[address service_areas partner_tags])
           .where(
-            'partner_tags.tag_id IN (:tags) AND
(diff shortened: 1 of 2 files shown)


ivan-kocienski-gfsc (approved):
Looks good

PR #2215, opened 2024-01-24
improve error messages

### LINKED ISSUE #2175: Whether or not a user is an admin for a partner (if they remove all partners from a user this can make them invisible to the admin but we force them to confirm this in the UI)

r-ferrier:
Fixes #2175 

## Description

Honestly I cannot believe that this just works! We managed to already design a controller for this that simply compares a list of permitted 'removable' users with the user being removed by the user, which very helpfully evaluates correctly who will disappear from your sight upon removal regardless of whether you are a partnership, neighbourhood, or partner admin. Wonderful!

So, to address the needs of this ticket, I went down a teeny rabbit hole to change the error message wording to reflect the role of each user, but on reflection decided it would be good to keep this very loosely coupled and made much more sense from a users' perspective to use generic language rather than couching the whole thing in current PlaceCal terminology, so I just rewrote it to work with any admin.

### DIFF
--- a/app/javascript/controllers/user_partners_controller.js
+++ b/app/javascript/controllers/user_partners_controller.js
@@ -5,7 +5,7 @@ export default class extends Controller {
 
 	connect() {
 		const confirmRemove =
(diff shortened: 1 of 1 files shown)


ivan-kocienski-gfsc (approved):
Looks good. Much better wording. I also cannot believe this just worked!

PR #2218, opened 2024-01-29
Removing partner addresses

### LINKED ISSUE #2128: NAs (and by extension partnership admins) can remove addresses from partners who have an address in their neighbourhood and a service area in their neighbourhood

ivan-kocienski-gfsc:
Closes #2128

There is now a button to clear a partners address on the edit partner form:
Screenshot_2024-01-29_12-03-12

- Admins can remove a partners address IF
  - partner has address
  - AND partner has at least one service area
  - AND
    - user is root
    - OR user owns partner
    - OR user is a partnership admin
    - OR user is a neighbourhood admin
      - AND user has partner address neighbourhood in their neighbourhood pool
- Admins will be warned if removing a partner will de-list it from their partner list IF
  - admin is a neighbourhood OR partnership admin
  - AND user has no other neighbourhoods within the partners' service areas


app/javascript/controllers/partner_address_controller.js
```diff
@@ -0,0 +1,67 @@
...
+		"partner_address_attributes_city",
+		"partner_address_attributes_postcode",
+	];
+
+	connect() {}
+
+	disconnect() {}
+
+	do_clear_address(event) {
+		console.log("do_clear_address");
```
r-ferrier, line 24: this will print into the browser console so should probably be removed

app/javascript/controllers/partner_address_controller.js
```diff
@@ -0,0 +1,67 @@
...
+
+		event.preventDefault();
+		console.log("do_clear_address");
```
r-ferrier, line 27: same for this one!

.rubocop.yml
```diff
@@ -55,6 +55,12 @@ Style/HashSyntax:
 Rails/ActionOrder:
   Enabled: false
 
+Metrics/ClassLength:
+  Exclude:
+    - "test/**/*.rb"
```
r-ferrier, line 60: why was this necessary - is there a good reason we don't want the tests validating too?

app/models/partner.rb
```diff
@@ -16,7 +16,7 @@ class Partner < ApplicationRecord
   has_and_belongs_to_many :users
   has_many :calendars, dependent: :destroy
   has_many :events
-  belongs_to :address, optional: true
+  belongs_to :address, optional: true, dependent: :destroy
```
r-ferrier, line 19: nice!

app/models/partner.rb
```diff
@@ -247,6 +247,47 @@ def has_service_areas?
     service_areas.any?
   end
 
+  def can_clear_address?(user = nil)
```
r-ferrier, line 250: I think this is unnecessary - if you can edit a partner at all, you can clear the address - so I would just remove this check. Also you've already defined access over in the policy.

app/models/partner.rb
```diff
@@ -247,6 +247,47 @@ def has_service_areas?
...
+    return false if user.root?
+    return false if user.admin_for_partner?(id)
+
+    user_hood_ids = user.owned_neighbourhood_ids
+    # this shouldn't happen (?) - a user has no neighbourhoods
+    #   but can still remove a partners address? if that is the case
+    #   then they aren't seeing the partner by neighbourhood anyway
+    #   so warning them that removing the address will remove the
+    #   partner seems wrong
+    return true if user_hood_ids.empty?
```
r-ferrier, line 272: not possible to be able to edit a user without owning any neighbourhoods or meeting the first two criteria, so this can just be removed

app/policies/partner_policy.rb
```diff
@@ -37,6 +37,10 @@ def destroy?
     return true if user.only_neighbourhood_admin_for_partner?(record.id)
   end
 
+  def clear_address?
+    index?
```
r-ferrier, line 41: i know that logically this will return the same result right now, but I think you should use `edit?` here instead of `index?` as the defining logic for who gets to edit this, as this should really be the deciding scope, and it will get changed in future!

app/views/admin/partners/_address_fields.html.erb
```diff
@@ -0,0 +1,40 @@
...
+
+    <% if partner.can_clear_address?(current_user) %>
+	<div data-partner-address-target="addressInfoArea">
+	    <p>Address in neighbourhood <%= link_to_neighbourhood(partner.address.neighbourhood) %>.</p>
+	    <p>
+		<%= link_to 'Clear Address',
+		       '#',
+		       class: "btn btn-secondary btn-sm",
+		       data: { action: "click->partner-address#do_clear_address" } %>
+		<strong>WARNING:</strong> clicking this button will permanently clear this partners address!
```
r-ferrier, line 35: this feels like overkill given there is also a UI warning - think we should lose this!

r-ferrier (changes requested):
Unfortunately this is a bit broken. Two things:

- A partnership or neighbourhood admin should not be allowed to remove an address unless it's in their neighbourhood - this was to stop them taking over control of partners that other users also admin for.

- once an address has been removed from a partner, it can never be put back! It's also now not possible to create a partner with an address.

See screenshot below:

Screenshot 2024-01-29 at 16 27 48

There's a few other smaller comments about the code I've left, let me know if you have any qs. I think it's possible it might be better for us to leave this now until the next sprint, especially as the permissions surrounding who can do what will become much simpler then, but the foundation work for this is really helpful and we can definitely extend upon this when we get to it!

app/models/partner.rb
r-ferrier: realise I've just contradicted myself here above - the check that needs to be in here is, instead of just seeing if you're a neighbourhood admin or a partnership admin, checking to see if you admin for the neighbourhood used in the address or are a partner admin or root - these are the only people who should be able to remove an address

.rubocop.yml
ivan-kocienski-gfsc: Ugh robocop loves to complain about everything. This is a test that your class is getting too long, which is common in minitest, so I disabled it for the minitest directory.

(rubocop was complaining that the PartnerTest class was too big)

app/policies/partner_policy.rb
ivan-kocienski-gfsc: I did think after submitting this should be `update?`

app/models/partner.rb
ivan-kocienski-gfsc: okay so i kept all the logic as before but extended the checks so if you are a neighbourhood admin then you must have the address neighbourhood in your neighbourhood pool if you want to remove the address.

i tried putting in the same check for partnership admins but you can't give a partnership admin a neighbourhood without making them a neighbourhood admin, so it felt moot.

r-ferrier: This doesn't quite work - I'll try to explain why!

The first 5 statements are all working fine.

Then we get to  two lines you should take out:

`return false unless user.neighbourhood_admin? || user.partnership_admin?`
and
`return true if user.partnership_admin?`

If you're able to see this screen and you don't own a partner and aren't a root, you're one of these, and even if you are, just being a neighbourhood_admin shouldn't give you access to any partner - just access to partners in your neighbourhood. We only want to return true if you admin for the address - you shouldn't be allowed to remove addresses other partnership or neighbourhood admins have added, because that will eject this partner from their neighbourhood. Does that make sense?

(Would probably also remove `return false if user_hood_ids.empty?` because the following line will immediately return false if it is indeed empty, and it will make it tidier)

.rubocop.yml
r-ferrier: I'm not sure whether this is a good idea or not! @aaaaargZombies @katjam do we want rubocop in our test suite as well? It seems to me like there could be bad knock on effects of just taking it out, don't want us to start committing lots of messy code just cos it's in tests, but also, tests take up so much of out time as it is. Could we not just make another partner test for this specific behaviour if it's making the original too big?

r-ferrier (changes requested):
That looks like the address disappearing problem is fixed which is great but partnership/neighbourhood admins can still remove addresses that don't belong to them, which shouldn't be possible - have left some notes in the code as to why this might be.

.rubocop.yml
ivan-kocienski-gfsc: This is just disabled for only one test, not the entire rubocop bit at all.

r-ferrier: ohhhh I'm sorry i missed that this is specifically for `Metrics/ClassLength:` - I thought you were just excluding all tests from rubocop! In that case is probably fine

r-ferrier (approved):
wahey, this works! Looks great, so nice to be able to implement this in this sprint ✨

PR #2219, opened 2024-01-30
Mute rollbar routing error warnings

### LINKED ISSUE #1666: DB timeout on production (via Rollbar)
https://rollbar.com/gfsc/placecal-production/items/239/

Database is timing out. We should find out why. How often is it happening.

r-ferrier:
Fixes #1666 

## Description

We're still logging tonnes of routing error warnings every month which is causing us to burn through our rollbar allowance in a few days.

The previous solution was only firing for specific 'bad' routes, but every route that ends in a 404 is a bad route and we're getting such a wide variety of them through that for now, just muting all of them is more useful to our error logging than trying to pick and choose which we care about. 

As an example, here are all the errors that were logged this morning:

Screenshot 2024-01-30 at 14 18 34

We should probably also look at serving a 404 page/ rerouting to home instead of just raising this error! (https://github.com/geeksforsocialchange/PlaceCal/issues/759)


ivan-kocienski-gfsc (approved):
Looks good to me. Hopefully rollbar calms down now and can become more useful

PR #2232, opened 2024-02-01
Add controllers to create action to prevent controller breaking if validation fails

### LINKED ISSUE #2229: [Bug]: select2 input on tags breaks for partner if validation fails
## Description

see #2217 

admin placecal-staging org_partners_new (1)

r-ferrier:
Fixes #2229

## Description

This js controller also needed to be used on create, which is the action used following a validation fail on new. Works now!


r-ferrier:
this work was also undertaken in #2237 so am closing this as duplicate work

PR #2236, opened 2024-02-01
Show admins the sites this partner appears on

### LINKED ISSUE #2231: show information about which sites a partner appears on to all users with view permission for that partner
## User story

As a partner admin, partnership admin, or neighbourhood admin, it would be useful to see which sites the partner I admin for appears on.

## A/C
- [ ] If an admin can see an edit partner page, then they can see a list of sites this partner is visible on (as just text)

ivan-kocienski-gfsc:
Closes #2231 

This is just a list of links to the sites (as the original behaviour) not links to the partner page of that site.

### DIFF
--- a/app/views/admin/partners/edit.html.erb
+++ b/app/views/admin/partners/edit.html.erb
@@ -2,33 +2,34 @@
 <h1 class="page-header">Edit Partner: <em><%= @partner.name %></em></h1>
 
 <% if current_user.root? -%>
-  <p>ID: <%= @partner.id %></p>
-  <%# Show the Neighbourhood the partner is related to via Address %>
-  <% if @partner.address&.neighbourhood&.present? -%>
-    <p>Address in neighbourhood <%= link_to_neighbourhood(@partner.address.neighbourhood) %>.</p>
-  <% end -%>
+    <p>ID: <%= @partner.id %></p>
+    <%# Show the Neighbourhood the partner is related to via Address %>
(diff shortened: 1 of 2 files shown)


app/views/admin/partners/edit.html.erb
```diff
@@ -2,33 +2,34 @@
 <h1 class="page-header">Edit Partner: <em><%= @partner.name %></em></h1>
 
 <% if current_user.root? -%>
```
r-ferrier, line 4: this section of info is still hidden to everyone except root users - think we should probably just show all of this info to anyone who can view a partner in the admin!

r-ferrier (changes requested):
Sorry, I appreciate this wasn't clear from the issue but I just meant for all the metadata at the top of the partner to be visible to all admin users, and not just roots. If you could remove the root restriction for the rest of it this will be good to go.

app/views/admin/partners/edit.html.erb
ivan-kocienski-gfsc: but it's not part of the issue A/C. perhaps i could extend/alter the issue to include all this stuff?

do you also want the ID readout?

r-ferrier (approved):
Looks good!

app/views/admin/partners/edit.html.erb
r-ferrier: I think you wrote the AC! I'll add it in.

PR #2246, opened 2024-02-14
Tidied up .env.example, made PG port configurable

### LINKED ISSUE #1383: Update .env.example with all configurable variables
## User story

As a developer setting up the project I need see examples for configurable vars so that I can get my instance set up.

## Acceptance criteria

- [x] If it is possible to override an .env var, that var name is in the .env.example
- [x] If the possible values for the .env var are not obvious, there is an example of a value of the right type
- [x] There are no unused variables in the example

Missed out in https://github.com/geeksforsocialchange/PlaceCal/issues/1351
image

## Implementation plan

ivan-kocienski-gfsc:
Closes #1383


r-ferrier (approved):
looks good!

PR #2249, opened 2024-02-15
Clean up addresses

### LINKED ISSUE #2005: Clean out broken Addresses
## Description

There are 1,037,176 addresses of which 1,037,017 have no partner (only 159 have partners) and 14 have no address.

Some addresses 
- have bad postcodes and should be deleted. e.g. 'M15', 'M16', 'M144' and 'UK'.
- are missing neighbourhoods
- are missing partners

---

## notes

- events can have address
- it looks like some addresses are being remade (maybe by the importer) 1 address has around 700,000 duplicates
- requires research to see why we have to many and come up with a plan to remove as many as possible

ivan-kocienski-gfsc:
Closes #2005 

A rake task `db:clean_bad_addresses` that deletes addresses that have no place, partner or event.
Minor tidy up of Address and Event models.

### But...

It looks like the calendar importer orphans addresses as it runs. Further investigation is needed as to where/why this is happening. (#2254)


ivan-kocienski-gfsc:
Random thoughts from the placecal-ops channel:

this is a bit of a rabbit hole. just found that addresses have a link to calendars BUT the calendar table has no FK, so if you event.calendars it just raises a "missing field" exception.
[16:00]
we also appear to (or did at one point) recycle addresses, so one address has hundreds of events, which i am not sure is correct given how we decided that each partner will have its own address. not sure what the rationale (utility) for having events share addresses is.
[16:03]
like- an event can have its own address, or it can get the address from the partner and that is it. maybe we allow people to find events in their (the sites) neighbourhood that have no associated partner?

app/models/address.rb
```diff
@@ -12,7 +12,7 @@ class Address < ApplicationRecord
 
   has_many :events
   has_many :partners
-  has_many :calendars
+  # has_many :calendars
```
aaaaargZombies, line 15: how come we are removing this relation? and if we are removing it should we delete rather comment, do we need to do any migration related tasks?

aaaaargZombies:
```sh
$time rake db:clean_bad_addresses 
found 1161163 addresses
  1160881 orphaned addresses found
rake db:clean_bad_addresses  4.67s user 0.37s system 0% cpu 48:49.77 total
```

```irb
irb(main):001> Address.all.length
  Address Load (18.7ms)  SELECT "addresses".* FROM "addresses"
=> 282
```

big improvement in speed (and num of addressed)

aaaaargZombies (approved):
:seedling: time for a spring clean

PR #2251, opened 2024-02-16
make up!

aaaaargZombies:
I really liked `make up` / `make down` on the GI site, how do people feel about making it consistent here too?


r-ferrier (approved):
yes, this makes total sense!

Issue #1319, opened 2022-06-16
Migrate CSS to cssbundling-rails


kimadactyl:
## User story

As a developer, I want all the CSS in one place and using a modern build process, in order that things compile quickly, its easy to know where to look for code, and we can use variables and mixins properly.

https://github.com/rails/cssbundling-rails

## Acceptance criteria

- [x] CSS all moved out of mountain view and main sprockets directory into a nice clean area
- [ ] CSS built by a sep process in the Procfile
- [ ] CSS structure is well documented and directory structure and inheritance easy to understand
- [ ] Mountain View doesnt complain about missing assets (bonus round - maybe out of scope)
 
## Implementation notes & questions

I think we use Dart Sass for this out of the listed options but I don't really understand them?

Variables and mixins need including in each file individually atm - its not the worst and doesnt impact actual performance but its a bad smell. Hopefully this removes the need for that.

## Implementation plan

_To be written by the developer_

kimadactyl:
with hindsight i am a bit worried this was an old kim busywork "feels really nice but doesnt actually acheive much" job but i think itd be nice. deprecating mountain_view feels more important, assume youre already on that tho. up to you obvi but yeah its funny with distance how things you thought were completely needed are perhaps not.

dont get me wrong there are some real pros to it. i think its nice to just use native node tools to do the frontend stuff, it makes a cleaner sep of concerns in the applicaiton structure, it  means you can have proper console output for sass errors bc of how Procfile thingy works, and i think it makes it easier to swap sass compliation library in the future. poss other beenfits too

but yeah. like How Important Are Those Things?

katjam:
> **proposal** we switch to `cssbundling-rails` once the last mountain_view component is gone, this will then be a small ticket.

Sounds like a good plan. Shall we mark as blocked until then?

kimadactyl:
Asking around people do seem very keen on Tailwind and i think it would make the component logic, themeing etc a lot simpler.

Possibly this makes sense to do as part of #1812 

This is comptible with cssbundling-rails but there may be other ways to do this.

kimadactyl:
I am gonna close this as I think the thinking here is very stale and the functionality needs re-deciding from first principles.

PR #2253, opened 2024-02-16
import our view_component stylesheets explicitly 

### LINKED ISSUE #2252: [Bug]: explicitly impirt `view_component` stylesheets into root scss file
## Description

We're accidentally relying on some automatic processing of our stylesheets built into mounatin_view. As we want to move off that system we should move our stylesheets somewhere they can be directly imported - see https://github.com/GenderedIntelligence/genderedintelligence.co.uk/ as an example

## Steps to reproduce

1. What you did first
1. What you did next

## Platform (if relevant)

aaaaargZombies:
fixes #2252 

## description

I took a look at #1319  and think removing `mountain_view` might be a required step to getting everything building through the standard asset pipeline. 

That raised the issue that we were using `mountain_view` to import our `view_component` styles (and why we were surprised we couldn't colocate styles in the GI project) 

This work is a first step in weaning us off `mountain_view` and putting a structure in place for adding future components... but It was a Friday special where I wrote the AC and PR on my own so perhaps Monday eyes will uncover something important I missed.


ivan-kocienski-gfsc (approved):
Nice, very tidy

PR #2255, opened 2024-02-19
remove all traces of collections from the system

### LINKED ISSUE #835: Allow general access to & fix Collections or remove them
e.g.
Multi select fucked when there's tons of options (see collections)
Collection maps don't show addresses with no matching Partner

r-ferrier:
Fixes #835 

## Description

This is a PR to complete the work to remove collections. After deleting the two collections we still had, I realised how much the code has changed since they were last used. 

This is a feature that has fallen out of use, and that we are holding onto just in case it becomes useful to us again. A side effect of us maintaining this connection is that it can unintentionally shape the way we design other parts of the system, as we mirror patterns found here, or worse, make use of it elsewhere.

I think this kind of approach is causing us a lot of maintenance headaches, and this is a great example of something that would be easier to rebuild to better suit our current set up if we wanted to do something like this again. 

I'm putting this in as a draft as I realise this is more than I said I would do to collections! I just wanted to see how much work it would be to remove them and investigate how much of the system they touched, and after doing this, it seemed like it would be worth putting this work in as a PR at least so we could discuss it and maybe merge. Putting this in as one PR will also provide a handy guide to anyone working on something like this again in the future who wants to see what we had in place last time.


ivan-kocienski-gfsc:
Yes, i have been wanting to do this for the longest time!

As long as everyone is happy.

One question: do we want to do something with those routes? Like- redirect them to the home page or something?

r-ferrier:
We've decided to leave this in draft just while we sit with the idea as a team, and can hopefully put this in at the start of the next release.

ivan-kocienski-gfsc:
We are not doing this at this time and it likely needs rethinking and reworking.

Something that should be picked up by future devs if/when they come to clean up the code.

PR #2262, opened 2024-02-20
selfhost fonts

### LINKED ISSUE #2248: Host fonts ourselves
We don't want to have a cookie notif if poss and i think this is the only thing. Removes nasty google tracking!

aaaaargZombies:
fixes #2248 

sorry this is a lot of files, it basically removes the link/imports from the head of the html document / duplication in the fonts.css file. adds fonts.scss and a bunch of assets and their licenses.


ivan-kocienski-gfsc (approved):
This looks a little large but I assume the licence stipulates it's an all-or-nothing type affair.

Will approve!

PR #2264, opened 2024-02-20
Drops unused `partners_places` and `places` tables

### LINKED ISSUE #2258: Remove places tables
As part of #1950 

## A/C
- [ ] `places` table is removed
- [ ] `partner_places` is removed
- [ ] verify these tables are no longer referenced in code

ivan-kocienski-gfsc:
Closes #2258 

The `places` table used to map to the `Place` model that was removed a long time ago.
All the `place` relations (in Calendar and Event models) just point to `Partner` objects now, so it is very safe to remove these.

Here are all the model references to places:

```ruby
# app/models/calendar.rb
belongs_to :place, class_name: 'Partner', optional: true

# app/models/collection.rb
events.order(:dtstart).includes(:place)
#  (this is referencing the event place, see Event model)

# app/models/event.rb
belongs_to :place, class_name: 'Partner', optional: true

scope :in_place, ->(place) { where(place: place) }
# (references the above relation, which is a Partner)

scope :by_partner_or_place, ->(partner) { in_place(partner).or(by_partner(partner)) }
# (chains existing scopes, one defined just below)

# Filter by Place
scope :in_place, ->(place) { where(place: place) }
# (references the above relation, which is a Partner)

```


r-ferrier (changes requested):
I haven't checked yet but I think this could break some things - some other models have relationships to places, for example calendars and events, and there are lots of references in the code still - maybe these need to all come out before we can drop the tables?

r-ferrier:
What about bits like:

https://github.com/geeksforsocialchange/PlaceCal/blob/3f3032d00dfa62f8d3f9ab0822bfcbcbe9dc1004/app/views/admin/calendars/show.html.erb#L8-L10

What happens to these if the calendar contains a reference to a place id that is still in the database?

I feel like removing the tables is a great idea but I'm worried we might make it even harder to understand other bits of the code if we still have these references to a long gone model hanging around afterwards - I feel like this work should all be in the same PR. 

What do you think @aaaaargZombies? I could very well be wrong on this!

r-ferrier:
Ah, I see what you are saying @ivan-kocienski-gfsc - the names persist but they're referring to other things? There's still some parts of the code looking for `place_id` too

https://github.com/geeksforsocialchange/PlaceCal/blob/3f3032d00dfa62f8d3f9ab0822bfcbcbe9dc1004/app/models/partner.rb#L207-L210

Are these bits also disconnected?

ivan-kocienski-gfsc:
> Ah, I see what you are saying @ivan-kocienski-gfsc - the names persist but they're referring to other things? There's still some parts of the code looking for `place_id` too
> 
> https://github.com/geeksforsocialchange/PlaceCal/blob/3f3032d00dfa62f8d3f9ab0822bfcbcbe9dc1004/app/models/partner.rb#L207-L210
> 
> Are these bits also disconnected?

Yes. the line you point to specifically says `events.place_id = partners.id` so it is not using the places table at all :sweat_smile:

r-ferrier (approved):
This looks to be working fine for me and was a great shout to destroy.

I'm going to make a separate ticket about refactoring/renaming attributes we're calling 'place' that are used to represent partners because I think that would really help me to understand the system a lot better.

PR #2271, opened 2024-02-21
(WIP) of removing MountainView

### LINKED ISSUE #2270: remove mountain_view gem / code
Convert existing mountain view components to ViewComponents

```
- FreePublicWifiComponent < MountainView::Presenter
- NavigationComponent < MountainView::Presenter
- ComputerAccessComponent < MountainView::Presenter
- PaginatorComponent < MountainView::Presenter
- EventListComponent < MountainView::Presenter
- ErrorComponent < MountainView::Presenter
- HeroComponent < MountainView::Presenter
- ProfileComponent < MountainView::Presenter
- OpeningTimesComponent < MountainView::Presenter
- PlacePartnerPreviewComponent < MountainView::Presenter
- AdminIndexComponent < MountainView::Presenter
- StepsComponent < MountainView::Presenter
- AddressComponent < MountainView::Presenter
- DashboardCardComponent < MountainView::Presenter
- EventComponent < MountainView::Presenter
- PlaceComponent < MountainView::Presenter
- AdminEditComponent < MountainView::Presenter
- BreadcrumbComponent < MountainView::Presenter
- ContactDetailsComponent < MountainView::Presenter
```

ivan-kocienski-gfsc:
Closes #2270 

This is a very small amount of progress towards removing mountainview.

The process is (something like):
- rename component directory like `address` -> `address_component`
- rename all the files in `address_component` directory appropriately
- move the`address_component.rb` file in to just `/app/compondents` (if there is one)
- create (copy) a ViewComponent template into the `address_component.rb` file
- open the view partial and convert all the `properties[:street]` to just `street` variables
- copy over all the variables from the partial into the component class, making `attr_reader` declerations, set up constructor to take all those variables as parameters, being careful to note which attributes are optional
- copy over any business logic from the MountainView helper into the ViewComponent class
- update all the call-sites in the view

repeat that ~50 times


kimadactyl:
This is already a chunky PR - lets re-add the `mountain_view` depend and get this merged when we are next on it. No need to do it all in one go!

kimadactyl:
Are you not confident merging it in as is? It looks like good work to date just not yet ready to remove the gem?

ivan-kocienski-gfsc:
This was never production code and only research that was pushed so it could be picked up later on.

Have agreed with the dev team this is very rough and I wouldn't be happy merging without more work done.

aaaaargZombies:
seems like there's been some ongoing confusion / differences of opinions about merging this PR so for the next person who looks into this here is a status update.

- Ivan removed the `mountain_view` gem and tried converting one component to `view_component` to see what breaks and how long it took.
- I generated a list of components to convert see #2270 which yielded 29!
- It was decided that we didn't have / couldn't justify the time to re-implement the same functionality, instead we should migrate a component when we have to modify it based on a raised issue.
- once these have been cleared we can remove the gem.
- then #1319  will be unblocked.

PR #2272, opened 2024-02-21
Removal of active_link_to gem

### LINKED ISSUE #2259: [Bug]: Remove active_link_to gem and write own simple solution
## Description

This gem https://github.com/comfy/active_link_to is now very out of date and is only being used in two places in our code:

https://github.com/geeksforsocialchange/PlaceCal/blob/6b59d83d9959cb84fd962d96cab11edc5d201900/app/components/footer/_footer.html.erb#L18-L22

and

https://github.com/geeksforsocialchange/PlaceCal/blob/6b59d83d9959cb84fd962d96cab11edc5d201900/app/components/navigation/_navigation.html.erb#L24-L26

This should be easy to refactor and then we can remove the gem

This came out of https://github.com/geeksforsocialchange/PlaceCal/issues/2159

ivan-kocienski-gfsc:
Closes #2259 

A small gem that is used in the navigation menu links to add 'active' classes to anchor links.

Implemented in a helper now


app/helpers/application_helper.rb
```diff
@@ -51,4 +51,21 @@ def image_uploader_hint(uploader_field)
       number_to_human_size(uploader_field.size_range.max)
     )
   end
+
+  # ported from https://github.com/comfy/active_link_to/blob/master/lib/active_link_to/active_link_to.rb
+  def active_link_to(title, url, data: nil)
```
aaaaargZombies, line 56: I like this - it's such a small amount of code better to have it here where we can keep an eye on it.

aaaaargZombies (approved):
Bon!

Issue #2268, opened 2024-02-21
We have attributes saved as 'place' and 'place_id' on some models that are actually referring to partners


r-ferrier:
## Description

`Place` is a record that used to exist in the system. We still have lots of references to it that are now pointed at partners, which makes understanding the flow of the system confusing at times. It would be really helpful for my own comprehension if we could rename these attributes to better reflect the models they are actually referencing.

kimadactyl:
Events can have Places, which are references to Partners.

This was less confusing when Partners and Places were separate things in the code base but now yeah is very confusing.

Before starting this we should define desired functionality again.

PR #2275, opened 2024-02-22
Partner address calendar importer fix

### LINKED ISSUE #1219: [Bug]: Deleting an address from a partner will cause importer issues with any calendar with that partner set as a default location
## Description

This is theoretical but putting this here before I forget!

## Steps to reproduce

1. What you did first
1. What you did next

## Platform (if relevant)

ivan-kocienski-gfsc:
Closes #1219 

Two checks added:
- stop admins from removing a partners address if that partner is the default location for any calendar
- validate (and stop) the calendar importer task if the strategy needs a default location (place address) and none is set up.


app/models/partner.rb
```diff
@@ -17,6 +17,7 @@ class Partner < ApplicationRecord
   # Associations
   has_and_belongs_to_many :users
   has_many :calendars, dependent: :destroy
+  has_many :calendar_places, class_name: 'Calendar', foreign_key: :place_id, inverse_of: :place
```
aaaaargZombies, line 20: This place / partner muddling is getting pretty confusing (for me at least). I'm wondering if there's some way we can make it clearer? 

I'm not entirely sure what this represents. the name suggests it's other partners(places)? that have been linked to the calendar the partner owns but the test set up makes it look like it's a list of calendars.

Is it all the calendars that list this partner as the location of their events, rather than the partner the calendar belongs to?

aaaaargZombies (commented):
OK I think I've understood this through some modification to the tests to see what works and what doesn't. This looks very useful and something we should have already been doing.

Is there a way to make this a validation on the calendar and partner so users get a warning when they are creating a combination that will is doomed to fail?

I'm not going to approve or request changes as I'm going to be off next week and just mopping up my remaining hours for this week so will let you and Rosie come to a conclusion, I'm sure I'll be happy with whatever you two decide.

app/models/partner.rb
r-ferrier: I agree on this but think we can handle them separately - I've made this ticket #2268 to document

r-ferrier:
Hmm this one is tricky! I think (one day) we probably should resolve this by forcing the calendar importer to default to a sensible strategy when an address is not existing, but that feels like something further down the line and part of a calendar fixing task. Given that it is possible for people to import events without any address and without assigning one, that seems more sensible to me than not letting them do it.

Personally I think leaving the error in the system is preferable to preventing people from taking this action. Here's why:
- It means they can easily remove an address from a service if it changes
- If their calendar fails to import they can see this, and they can see the reason why it fails (no address)

I'd be up for us adding this change in, but only if we can add some kind of validation to explain why users can't remove the address and what action they need to take to allow them to remove it.

ivan-kocienski-gfsc:
Closing this because otherwise it will just hang around.

The team decided that this problem would be better fixed if we reviewed the importer at a higher level and had a more substantial refactor of how it worked (to make it more robust and give better user feedback).

PR #2279, opened 2024-02-27
Fix/revert datatables removal

r-ferrier:
This code is not working so I am putting it back the way it was before, with one more check to assert the variable is not `== ""` before calling destroy on it. This will hopefully prevent system tests from breaking quite so much!


ivan-kocienski-gfsc (approved):
I hope this helps our poor suffering CI!

r-ferrier:
> I hope this helps our poor suffering CI!

me too! 😬

PR #2289, opened 2024-03-06
Update the way partner moderation works to remove the sharing of emails

r-ferrier:
## Description

- Small edits to the email & UI info that is shown to partner admins if their partner is hidden to remove the details of the user who moderated them. Now they will have to email support@placecal.org to settle any problems that arise.
- Change to the email sent to support@placecal.org as a notification that this action has been taken to include the information about who took the action.


aaaaargZombies (approved):
grand!

PR #2315, opened 2024-03-18
Removal of mountainview styleguide engine

### LINKED ISSUE #2309: [Bug]: Fix openstruct error
see rollbar for more in depth info:

https://app.rollbar.com/a/gfsc/fix/item/placecal-production/400

## A/C

- [ ] MountainView styleguide (i.e. [here](https://placecal.org/styleguide/styleguide/event)) doesn't exist (can't crash)

ivan-kocienski-gfsc:
Closes #2309 

As it was crashing on production


aaaaargZombies:
what about views that won't be reachable https://github.com/geeksforsocialchange/PlaceCal/blob/f7b4c7543269762174382dcc9b2a6872854cdd5c/app/views/layouts/mountain_view.html.erb#L0-L0

ivan-kocienski-gfsc:
i did try to remove more of mountain view in various places like in the layout where we reference stylesheets and javascripts, or like in the stylesheet asset definition. 

But the original issue was to fix a crash that was happening on production and being caught by rollbar. It was not about removing mountain view.

I didn't want to get into a whole mess removing mountainview as we are still using it and removing that code may brake all our MV components.

So it's just about removing the one specific bits of MV that was causing problems.

aaaaargZombies (approved):
That makes sense, we are trying to phase out mountainview incrementally.

merge away!

PR #2317, opened 2024-03-18
make place requirements clearer on partner forms

### LINKED ISSUE #2310: [Bug]: It is not clear when creating a partner that you can make one with just a service area
## Description

It's not clear from the ui that you can create a partner with just a service area and not an address - we need to fix this in the hint text and in the validation. At the moment it makes it look like postcode is required.

It should be clear that a partner needs to provide *either* an address or a service area, but that a street address is not required. 

## Acceptance Criteria

- [ ] There is a clear description showing that your partner Address or Service Area (or both)
- [ ] There is no asterisk indicating that post code field is required
- [ ] There is no asterisk on the first line of address field.

aaaaargZombies:
fixes #2310 

I've moved the place based parts of the form together and added text explaining that one is required. Updated with honor's description for people using the admin UI. Hopefully this makes it a bit clearer.

image

### DIFF
--- a/app/views/admin/partners/_form.html.erb
+++ b/app/views/admin/partners/_form.html.erb
@@ -37,9 +37,7 @@
     <br>
     <hr>
     <h2>Place</h2>
-    <p>
-    Partners need to be connected to <b>atleast one place</b>, this can either be the address of the building they opperate from or the area they deliver services too. If this is more than one place you can add multiple service areas.
-    </p>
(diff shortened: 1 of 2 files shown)


honor-gfsc:
I would suggest something like:

Partners need to be associated with at least one place. This can be a fixed address if they usually operate from one location, or a service area which can be a Ward or District. Keep this information as specific as possible – limited to where this partner actually runs events rather than the geographical spread of the people they are trying to reach. 

This is probably a bit long and maybe the explanation is too much but we can edit down.

ivan-kocienski-gfsc (approved):
This seems like a sensible layout, let's hope it is easier to intuit for all users

Issue #2066, opened 2023-09-20
Replace generic 'tags' inputs with ones for partnerships, facilities and categories


kimadactyl:
## User story

As a developer I am getting confused about the current combined tags field and would like it to be split into the seperate bits in order that we use proper terminology more.

As a partner admin and site admin the current input is pretty confusing as some of the options are not real

In other words it should be an architectural irrelevance that these things inherit from tags and currently they are being talked about interchangably.

A lot of this will be covered when #2061 and #2051 are done, this ticket is a failsafe to make sure its done completely.

## Acceptance criteria

- [ ] #2333
- [x] #2334
- [x] #2335
- [x] #2336
- [ ] #2337

## Implementation notes & questions

This is supposed to be a quick easy fix so I'm going to implement the suggested UI

```ruby
<%= f.association :partnerships (...) %>
<%= f.association :categories (...) %>
<%= f.association :facilities (...) %>
```

And raise a draft PR so we can see how many other tickets it raises and make an informed judgement about time/cost/benefit. Essentially doing some active research. Fingers crossed it will be plain sailing ⛵🏝️🌞

## Implementation plan

_To be written by the developer_

aaaaargZombies:
![2024-01-11-14:22:01-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/0f3e909d-6765-40f9-b3f5-c1b262495db3)

split this input into 3 inputs

kimadactyl:
OK so it's my understanding that this should be as simple as replacing the one tags input with three like this. I'm sure the validations need redoing but I think the fact they're backed with a single database table is irrelevant.

If I'm wrong then I understand this is a significantly more complex ticket!

```ruby
<%= f.association :partnerships (...) %>
<%= f.association :categories (...) %>
<%= f.association :facilities (...) %>
```

r-ferrier:
For future context: this was discussed at a meeting this afternoon. There's potentially quite a bit of unpicking here required to make this happen. I raised some concerns that me and Ivan had about what the evidence is that this is required, how much time we think it could take, and if we think this is introducing further complexity to an already overloaded part of the system. We will try to make time to investigate this next week after prioritising the issues in the milestone based on the feedback from miro & the session notes.

katjam:
Closing as has been sharded off.

PR #2320, opened 2024-03-21
WIP: research splitting out tag imputs

aaaaargZombies:
research for #2066

OK I picked the simplest option to implement as a test run: sites only have partnerships.

 I've got it to a state where simply scoping the tags down from the full set to partnerships doesn't reduce the amount of failing tests (55 initially, down to about 15 now) but starts to introduce new one.  eg.
 
 ```ruby
  test 'show computer access card when partners are tagged for it' do
    tag = create(:tag, name: 'computers')
    @site.partnership << tag

    partner = build(:partner)
    partner.service_area_neighbourhoods << @neighbourhood
    partner.tags << tag
    partner.save!

    get 'http://hulme.lvh.me'

    assert_select '.help__computer_access'

    url = partner_path(partner)
    selector = ".help__computer_access a[href='#{url}']"
    assert_select selector
  end

  test 'show public wifi card when partners are tagged for it' do
    tag = create(:tag, name: 'wifi')
    @site.partnership << tag

    partner = build(:partner)
    partner.service_area_neighbourhoods << @neighbourhood
    partner.tags << tag
    partner.save!

    get 'http://hulme.lvh.me'
    assert_select '.help__free_public_wifi'

    url = partner_path(partner)
    selector = ".help__free_public_wifi a[href='#{url}']"
    assert_select selector
  end
```

[view code in context](https://github.com/geeksforsocialchange/PlaceCal/blob/4399ac0e5f48335dbb9f4b510d3524d8b1f2d082/test/integration/sites_integration_test.rb#L59-L92)

This is the point where we will start to need to reconsider the underlying implementation rather than just what is happening with the admin UI. For example it's not clear to me why a `Site` would be receiving facility tags.

Given that we have 3 other places to implement this and the Partners is the most complicated involving some custom validation based on the current UI I think this is the best place to stop and open up discussion for next steps.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -54,7 +54,7 @@ class ApplicationController < ActionController::Base
 
     if site
       events = events.for_site(site)
-      events = events.with_tags(site.tags) if site.tags.any?
+      events = events.with_tags(site.partnership) if site.partnership.any?
     end
 
     events = events.in_place(place) if place
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -6,16 +6,16 @@ class PagesController < ApplicationController
 
   def home
     @neighbourhoods = Site.published.select do |site|
-      site.tags.none? { |tag| tag.type == 'Partnership' }
+      site.partnership.none? { |tag| tag.type == 'Partnership' }
     end
   end
 
   def find_placecal
     @neighbourhoods = Site.published.select do |site|
-      site.tags.none? { |tag| tag.type == 'Partnership' }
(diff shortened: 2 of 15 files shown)


kimadactyl:
Ouch, OK I see why this is so complicated now - its less the backend storage and just about how much this field is referenced.

I would say the only one that really matters is the partner edit/show form as its in use all the time - appreciate that might be the most complex tho.

It it acutally breaking when you do this, or are the tests just mad?

(Facilities could come in handy if anyone wanted us to make a site for accessible loos or free wifi, but that feels like a side issue)

PR #2321, opened 2024-03-21
Rake task that lists and removes duplicate events

### LINKED ISSUE #2287: [Bug]: Events displaying multiple times
## Description

Import a calendar (Google) - some events are showing up more than once in PlaceCal listing.

Maybe related to:
https://github.com/geeksforsocialchange/PlaceCal/issues/1802
https://github.com/geeksforsocialchange/PlaceCal/pull/1827#discussion_r1184733827
https://github.com/geeksforsocialchange/PlaceCal/issues/1596

https://norwich.placecal.org/events/2024/3/15
image

## Steps to reproduce

1. View events for Queer Norfolk

The events are only showing once in the source calendar, but displaying in duplicate in PlaceCal

ivan-kocienski-gfsc:
Closes #2287 

There was a long investigation in to how the event importer operates when dealing with duplicates but as far as can be figured out there is a discrepancy between how the Event model defines a duplicate and how the EventResolver defines a duplicate. The Event model validation also only runs on save and not update. 

I think it may have to do with some odd data we get from the remote source: like if a user moves an event around on their calendar. But this would require finding or recreating the conditions and stepping through the EventResolver.

In the interim this PR provides a rake task `rails db:find_and_clean_duplicate_events[]` which will list and remove duplicate events. When run with no arguments it only shows you something like
```
Found 5 events with duplicate IDs
37r53ov39h6m1n948thtpp61ro@google.com has duplicate count: 1 for date 2024-03-27 11:00:00 +0000
  [377603]
76ma3pl0c9mqgc6rcs5nmea55o@google.com has duplicate count: 2 for date 2022-10-18 00:00:00 +0100
  [317332,317333]
5nesefucdvre4pq56f4u2d18e4@google.com has duplicate count: 2 for date 2024-03-15 17:00:00 +0000
  [367369,368285]
3nmemfvb3mnclgsa8cqt2cdauo@google.com has duplicate count: 1 for date 2024-07-20 10:00:00 +0100
  [355460]
4ao2d36r8ngo5s04i4v7dhf7fv_R20210809T180000@google.com has duplicate count: 1 for date 2021-12-13 19:00:00 +0000
  [308662]
```

But when run like `rails db:find_and_clean_duplicate_events[true]` it also destroys the duplicates.

This rake task could be run every day or hour or something.

## A better database snapshot restore

Have also implemented `rails db:restore_local_2` that uses more railsy actions, so instead of calling `dropdb` directly it invokes `rails db:drop` and it also reads all its configuration arguments from the rails configuration object, rather than having them fixed in code.

I didn't want to unilaterally replace `db:restore_local` without first getting some feedback from other developers.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -186,6 +186,43 @@ namespace :db do
 
     $stdout.puts "Restoring DB dump file #{filename} to local dev DB. (May take a while.) ..."
     puts `dropdb placecal_dev && createdb placecal_dev && pg_restore -d placecal_dev #{filename}`
+
+    if $CHILD_STATUS.success?
+      $stdout.puts '... done.'
+    else
+      warn 'Failed to restore DB dump to local dev DB!'
+      warn 'Please manually check to see whether local DB dev still exists.'
+      exit
+    end
+  end
+
+  desc 'a better version of db:restore_local'
+  task restore_local_2: :environment do
+    # this does the same thing as the previous task but in a less hacky way by
+    #    utilising rails tasks and configurations.
+
+    filename = ENV.fetch(DB_DUMP_ENV_KEY, nil)
+    raise "Could not find #{filename} file!" unless File.exist? filename
+
+    $stdout.puts "Restoring DB dump file #{filename} to local dev DB. (May take a while.) ..."
+
+    db_config = Rails.configuration.database_configuration[Rails.env]
+
+    begin
+      Rake::Task['db:drop'].invoke
+
+    rescue ActiveRecord::ProtectedEnvironmentError
+      # https://github.com/rails/rails/issues/34041#issuecomment-426097998
+      puts "*** DB drop failed, you may have pending migrations ***\n\n"
+      raise
+    end
+
+    Rake::Task['db:create'].invoke
+
+    command = "pg_restore -h #{db_config['host']} -p #{db_config['port']} -d #{db_config['database']} -U #{db_config['username']} #{filename}"
+    puts "running: #{command}"
+    system command
+
     if $CHILD_STATUS.success?
       $stdout.puts '... done.'
     else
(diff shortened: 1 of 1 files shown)


r-ferrier:
Ok, I *think* I understand what this is doing! I can run the task fine locally without any args, but I'm unable to get it to run with args - is there something I'm doing wrong here?

`rails db:find_and_clean_duplicate_events[true]`
just returns
`zsh: no matches found: db:find_and_clean_duplicate_events[true]`

Also wanted to check on the `db:restore_local_2` task - we currently use the regular `db:restore_local` as part of the setup in our makefile - should this be changed to this new task?

ivan-kocienski-gfsc:
weird. you have the correct syntax. is that a z-shell thing? just checked and `rails db:find_and_clean_duplicate_events[true]` works fine (in bash on linux)

Hmm. It's not a good look to have the two similarly named rake tasks. If you change over the makefile to use the 'new' one does it all still work for you? if that is the case then i will replace the original with the new code.

r-ferrier:
ok great, that works!

`db:restore_local` is used as part of the `dump_production_and_restore_other` rake task, which is then in turn used by the makefile - do you use this task at all and should it be updated in here?

ivan-kocienski-gfsc:
i do not use the makefile beyond running the default test task.

if you change out the `restore_local` for the `restore_local_2` task it should keep working as before. i think i'd prefer if only one of those tasks remained as they do the same thing (but the second is more inline with rails configurations).

but i cannot check this on my machine as i don't use dockerised postgres at all.

perhaps this could be merged in and a new ticket to resolve the two tasks could be made up so we can get the fixes in this PR live?

lib/tasks/db.rake
```diff
@@ -179,13 +179,30 @@ namespace :db do
     end
   end
 
-  desc "Restore db dump file #{DB_DUMP_ENV_KEY}=<filename> to local dev DB"
+  desc 'a better version of db:restore_local'
```
aaaaargZombies, line 182: we should add a better description, sorry I could have done that on my PR

aaaaargZombies (approved):
OK rake task works and make file works: LGTM!

PR #2323, opened 2024-03-21
Show the correct calendars on the site calendars page

### LINKED ISSUE #2312: [Bug]: Calendar listing for partnership site (seems to) include all calendars for geographical area
## Description
Viewing the listing for a partnership based site's calendars

## Steps to reproduce (on staging - not verified in production)

1. view https://admin.placecal-staging.org/sites/consciouscollectivemcr

I expected to only see calendars for the 4 partners in Conscious Collective but this list seems much too long for that

r-ferrier:
Fixes #2312
## Description

- Changes the scope for `Calendar.that_appear_on_site` to take into account a site's partnership tags as well as the neighbourhoods it covers
- Adds a test to assert this is working as expected and to avoid hitting this type of regression again if these rules change again in future


app/models/calendar.rb
```diff
@@ -45,18 +45,28 @@ class Calendar < ApplicationRecord
...
-      .joins('JOIN events ON events.calendar_id = calendars.id')
-      .joins('JOIN partners ON events.partner_id = partners.id')
-      .joins('LEFT JOIN addresses ON partners.address_id = addresses.id')
-      .joins('LEFT JOIN service_areas ON service_areas.partner_id = partners.id')
-      .where(
-        '(addresses.neighbourhood_id IN (:site_neighbourhoods)) OR (service_areas.neighbourhood_id IN (:site_neighbourhoods))',
-        site_neighbourhoods: site_neighbourhoods
-      )
-      .distinct('calendars.id')
+    site_neighbourhood_ids = site.owned_neighbourhoods.map(&:id)
```
ivan-kocienski-gfsc, line 48: also i think we use `site.owned_neighbourhood_ids` here (?)

Issue #2342, opened 2024-03-26
[Bug]: 500 error if saving a calendar with a bad/missing source url value


ivan-kocienski-gfsc:
## Description

On admin calendar form, if the source URL is blank when saving, it crashes out with a `500` page.

See [rollbar](https://app.rollbar.com/a/gfsc/fix/item/placecal-production/411)

This is also the case if the URL is malformed (from [comment](https://github.com/geeksforsocialchange/PlaceCal/pull/2345#pullrequestreview-1966544359)):

![2024-03-28-15:21:26-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/76148761-66af-49ec-9e46-889fd53b1c28)

kimadactyl:
Is that the same as this? https://github.com/geeksforsocialchange/PlaceCal/issues/2328

ivan-kocienski-gfsc:
I don't think so as that calendar has a source url set

ivan-kocienski-gfsc:
Sorry kim, but yes it looks like it is!

aaaaargZombies:
## verified

![image](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/56729f59-ec81-4d3d-bcce-5ed06d876665)

PR #2345, opened 2024-03-26
Form value persistance

### LINKED ISSUE #2301: [Bug]: Form fields are lost if validation fails
## Description

We should investigate dynamic in-page form field validation.

From notion: https://www.notion.so/gfsc/Kim-s-timeline-for-completion-12-3-24-d0e8ef7a03c44ba08e56d4477e3fea5d
Improve field validation process — can we have fields autosave as you go, etc? I think hotwire is meant to be good for this.

ivan-kocienski-gfsc:
Closes #2301, #2342 

Fix admin form value persistence
    
*Tags*: the type field can/must be set on new Tags only
    
*Calendars*: partner organiser selector
    
*Users*: partners selector

Note: none of the slug values persist because of our use of a gem to generate that value (on model save). It looks like it doesn't cope with having a user-provided slug value (that is then persisted across requests.


kimadactyl:
apols to shoehorn in a tiny layout thing but it never seems worth making a sep ticket for!

could we use this opportunity to fix the whitespace to the left of "Connected entities" please? this got knocked out of whack at some point and would be nice to have it put back, it should be flush left with the save partner button.

Screenshot 2024-03-28 at 11 56 47

app/controllers/admin/users_controller.rb
```diff
@@ -78,7 +78,7 @@ def create
         flash[:success] = 'User has been created! An invite has been sent'
         redirect_to admin_users_path
       else
-        Rails.logger.debug @user.errors.full_messages
+        @partners = collect_partners
```
aaaaargZombies, line 81: I expected this to be `@partner` not a list of partners for the new view / form. What am I missing here?

aaaaargZombies (changes requested):
I might be doing something extremely daft but most of what I'm testing is failing

---

### Partner

- Slug / NO
- image upload (?) NO - preview remains which is confusing.
    

### Tags

- Slug / no
- Type / YES
- Assigned Users / NO (input disappears!?) 

### Site

- Url / no YES
- image upload (?) NO - preview remains which is confusing.

### Calendar

- partner organiser / no  - complete crash

### Users

- partners / no - select2 also having a bad time

----

this is what happens with the calendar :|

2024-03-28-15:21:26-screenshot

aaaaargZombies:
> apols to shoehorn in a tiny layout thing but it never seems worth making a sep ticket for!

I've pulled this into a new issue #2351 as there's enough to work on here without muddying the water, This PR is much more likely to raise future issues because of it's complexity so it would nice to have it merged in free of unrelated changes.

Also, it means I can start working on it sooner rather than waiting for Ivan to finish this.

ivan-kocienski-gfsc:
Forgive me but I forgot to add into the release comment that non of the slug fields are persisted as we are using a gem to auto-generate those values and it's got some non-obvious logic to generate those slugs automatically and that doesn't seem to play well with having those values set manually by value explicitly by the user.

app/controllers/admin/users_controller.rb
ivan-kocienski-gfsc: This variable is set up so the form view has a list of partners with some of those partners pre-selected. I think it is used when creating a partner: it allows us to have a link that takes the admin to the "New user" form with the just-created partner pre-selected.

ivan-kocienski-gfsc:
Looks like that calendar crash you captured is because the calendar URL starts with `htts://` and not `https://` although why this wasn't picked up sooner is a problem

aaaaargZombies:
> Looks like that calendar crash you captured is because the calendar URL starts with `htts://` and not `https://` although why this wasn't picked up sooner is a problem.
> 
> This #2342 issue looks like it covers it

ok great I just tried to force an invalidation but must have hit a different know bug, glad it's covered elsewhere

aaaaargZombies:
> 
> ### Users
> 
>     * partners / no - select2 also having a bad time
> 

This is retaining the inputs on validation failure and I have created separate ticket for the select2 issue here  #2357

aaaaargZombies:
> ### Tags
> 
>   * Slug / no
> 
>   * Type / YES
> 
>   * Assigned Users / NO (input disappears!?)

OK this is now working for me with the users, discussed the slug issue, it remains if it has previously been set but not on a new tag. This is an issue with the Gem we use for slugs.

aaaaargZombies (approved):
Approving if you are happy to pull out a couple of issues into new tickets
- slug persistence
- handle images more gracefully

Those things can then be worked on independently and we won't hold up this useful work going in.

PR #2347, opened 2024-03-28
seperate out tag inputs on partners

### LINKED ISSUE #2333: Partners have separate facility, partnership and category inputs
## description

It is hard to explain to organizers why there are different types of tags and they all live in the same input, there is discussion in this pr  #2320 about #2066

It has been decided that the priority is to try and split out the partners tag inputs into separate fields.

## AC

- [ ] there is an input for category tags 
- [ ] category input has validation for max 3 selected
- [ ] there is an input for partnerships 
- [ ] partnership input has JS validation that warns a user about removing a tag from a partner
- [ ] there is an input for facility tags

## implementation plan

Investigate if it is possible to leave partners with a `tags` relationship and just update the inputs.

If not restructure all the places tags are checked on partners.

aaaaargZombies:
fixes #2333

-  there is an input for category tags
- category input has validation for max 3 selected
- there is an input for partnerships
- partnership input has JS validation that warns a user about removing a tag from a partner
- there is an input for facility tags

## notes

I have left as much of the `tag` associations in place so that we didn't need to restructure much of placecal, the forms add to the respective STI options and that's picked up by the tags. I did a small amount of tidying of the helper function file pointed tests in it's direction but I stopped short of ripping out the partnership scoping on the tag model, that should probably done after all the other entities have been shifted over. I think in the future I would avoid STI as it has created a lot of tight coupling between things that aren't really related.

I expect people will have opinions on the exact wording of the helper text in the UI but I tried to put something sensible in as a jumping off point.

image

### DIFF
--- a/app/controllers/admin/partners_controller.rb
+++ b/app/controllers/admin/partners_controller.rb
@@ -71,8 +71,6 @@ module Admin
 
       mutated_params = permitted_attributes(@partner)
 
-      before = @partner.hidden
-
       @partner.accessed_by_user = current_user
 
       # prevent someone trying to add the same service_area twice by mistake and causing a crash
@@ -86,9 +84,6 @@ module Admin
       hidden_in_this_edit = mutated_params[:hidden] == '1' && !@partner.hidden
 
       mutated_params[:hidden_blame_id] = current_user.id  if hidden_in_this_edit
-      Rails.logger.debug '0' * 80
-      Rails.logger.debug mutated_params
-      Rails.logger.debug '0' * 80
 
       if @partner.update(mutated_params)
         # have to redirect on associated service area errors or form breaks
--- a/app/policies/partner_policy.rb
+++ b/app/policies/partner_policy.rb
@@ -54,12 +54,13 @@ class PartnerPolicy < ApplicationPolicy
              { calendars_attributes: %i[id name source strategy place_id partner_id _destroy],
                address_attributes: %i[id street_address street_address2 street_address3 city postcode],
                service_areas_attributes: %i[id neighbourhood_id _destroy],
-               tag_ids: [] }]
+               tag_ids: [], category_ids: [], facility_ids: []  }]
 
     attrs << :slug if user.root?
     attrs << :hidden if user.root? || user.neighbourhood_admin? || user.partnership_admin?
     attrs << :hidden_reason if user.root? || user.neighbourhood_admin? || user.partnership_admin?
     attrs << :hidden_blame_id  if user.root? || user.neighbourhood_admin? || user.partnership_admin?
+    attrs << { partnership_ids: [] }  if user.root? || user.partnership_admin?
     attrs
   end
 
--- a/app/views/admin/partners/_form.html.erb
+++ b/app/views/admin/partners/_form.html.erb
@@ -153,7 +153,6 @@
     <%= f.association :facilities,
         label: "Facilities",
         hint:"What infrastructure does this partner provide.",
-        collection: options_for_partner_partnerships(),
         collection: Facility.all,
         input_html: { class: 'form-check', data: { 
           controller: 'select2',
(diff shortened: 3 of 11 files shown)


app/controllers/admin/partners_controller.rb
```diff
@@ -71,8 +71,6 @@ def update
 
       mutated_params = permitted_attributes(@partner)
 
-      before = @partner.hidden
-
```
aaaaargZombies, line 75: this wasn't getting referenced anywhere so just a bit of house keeping

ivan-kocienski-gfsc:
Just managed to create a partner with

Screenshot_2024-04-04_12-09-09

ivan-kocienski-gfsc (changes requested):
I found only the one bug

app/controllers/admin/partners_controller.rb
```diff
@@ -165,6 +164,18 @@ def setup
 
     private
 
+    def validate_category_limit
```
ivan-kocienski-gfsc, line 167: Lordy, remove this!

1. validations should only happen in the model (one place for data business logic)
2. the rule here is checking for a maximum of 5 and the model says it should be 3 (which is correct?)
3. this only happens on create, but not update (different behaviours depending on what the user is doing)
4. it also stops all the other validations running so if there are any other problems with the other fields they will not be shown.

app/policies/partner_policy.rb
```diff
@@ -54,12 +54,13 @@ def permitted_attributes
              { calendars_attributes: %i[id name source strategy place_id partner_id _destroy],
                address_attributes: %i[id street_address street_address2 street_address3 city postcode],
                service_areas_attributes: %i[id neighbourhood_id _destroy],
-               tag_ids: [] }]
+               tag_ids: [], category_ids: [], facility_ids: []  }]
```
ivan-kocienski-gfsc, line 57: Do we still need `tag_ids: []`? that may present a bypass if somehow something submitted a partnership ID as a basic tag, without the user being a partnership admin?

app/views/admin/partners/_form.html.erb
```diff
@@ -143,14 +143,32 @@
...
+        input_html: { class: 'form-check', data: { 
+          controller: 'select2',
+        } } %>
+    <hr>
+    <%= f.association :categories,
+        label: "Categories",
+        hint:"Partners may have up to 3 category tags, the public will be able to filter by these",
+        collection: Category.all,
+        input_html: { class: 'form-check', data: { 
+          controller: 'select2',
```
ivan-kocienski-gfsc, line 171: Maybe we could (as a future issue) limit this in the browser?

ivan-kocienski-gfsc (changes requested):
This looks like great work!

Mostly okay except for one showstopper that just needs to be taken out.

I am also going to make a suggestion that is not directly linked to these modifications about the messaging that I found (below).

ivan-kocienski-gfsc:
Just a small tweak here but the error message was a bit confusing

When saving with too many categories the error message is "Partner.tags can contain a maximum of 3 Category tags"

But I think that [line](https://github.com/geeksforsocialchange/PlaceCal/blob/6b34264ee85b29cba33cae558a8f0ca137d26284/app/models/partner.rb#L489) should be
```ruby
 errors.add :categories, 'Can only have maximum of 3 selections'
```

So it reads better and highlights the field.

(I noticed there are a few validation methods that add messages to `base` here when they should be pointing to the specific field, but I am going to make an issue about those)

r-ferrier:
Have done some digging into partner validation. The custom validation, `three_or_less_category_tags` is getting called on create and update, but it's not firing because categories do not exist yet on create - only `tag_ids` do. If you stick a log into the validator just above the return statement to spit out the content of `categories` and `tag_ids` you can see it firing as it should on create and update, and can see what you're working with!

aaaaargZombies:
closing as I made a new branch with a cleaner git history see #2378

PR #2350, opened 2024-03-28
Partner form live validtion

### LINKED ISSUE #2341: Deal with making new partners with the same name as an existing partner better
## User story

As an organiser, I want to know if a new partner I am adding already exists before I fill in the entire new partner form, in order that I don't waste my time filling in info that already exists.

Motivation: the new partner form is very long, and organisers can't see every partner in the database, or might not look properly. It's currently frustrating to fill in that entire form to be told that {partner name} already exists. We should attempt to mitigate this much sooner.

Screenshot_2024-03-08_at_11 58 04

## Acceptance criteria

- [ ] Partner name field gives much quicker validation OR appears as a separate first step

Do we want to retain the uniqueness constraint on partner names? Not too unbelievable to think there would be more than one org with the same name -- but it would signficantly complicate this dupe detecting.

## Implementation plan

ivan-kocienski-gfsc:
Closes #2341 

When entering a name on the admin partner page an indicator is displayed if the name is already taken.

Screenshot_2024-03-28_13-57-52


app/javascript/controllers/partner_form_validation_controller.js
```diff
@@ -0,0 +1,59 @@
...
+		const nameProblemFeedbackElement = document.getElementById(
+			"partner-name-feedback"
+		);
+
+		this.inputFunction = _.debounce(() => {
+			this.sourceTarget.classList.remove("is-invalid");
+			nameProblemFeedbackElement.style.display = "none";
+
+			const nameValue = this.sourceTarget.value;
+			if (nameValue.length < 1) {
```
aaaaargZombies: we don't allow names less than 5 characters so they can't collide till you hit that amount

app/controllers/admin/partners_controller.rb
```diff
@@ -144,6 +144,12 @@ def clear_address
       end
     end
 
+    def lookup_name
+      found = params[:name].present? && Partner.where(name: params[:name]).first
```
aaaaargZombies, line 148: I'm able to type `caracol` without a warning but I get a rejection because `Caracol` already exists, is there a way to check for partners that isn't case sensitive.

app/views/admin/partners/_form.html.erb
```diff
@@ -15,7 +15,9 @@
       <% end  %>
     <% end  %>
     <h2>Basic Information</h2>
-    <%= f.input :name, class: "form-control" %>
+    <%= f.input :name, class: "form-control", input_html: { 'data-controller': "partner-form-validation", 'data-partner-form-validation-target': 'source' } %>
+    <p class="invalid-feedback" id="partner-name-feedback">This partner name has already been taken.</p>
```
aaaaargZombies, line 19: I'm getting a double warning 

image

app/javascript/controllers/partner_form_validation_controller.js
```diff
@@ -0,0 +1,59 @@
...
+					if (!data.name_available) {
+						this.sourceTarget.classList.add("is-invalid");
+						nameProblemFeedbackElement.style.display = "block";
+					}
+
+					this.sourceTarget.focus();
+				});
+		}, this.inputDebouncePeriod);
+
+		this.sourceTarget.oninput = (event) => {
```
aaaaargZombies, line 56: It might be a bit clearer to follow the stimulus convention for actions here, I'm assuming stimulus will remove the event for us on disconnect If we use it's built in system for that too.

https://stimulus.hotwired.dev/reference/actions

app/javascript/controllers/partner_form_validation_controller.js
```diff
@@ -0,0 +1,62 @@
...
+
+		const nameField = this.sourceTarget;
+		const originalValue = nameField.value;
+
+		this.inputFunction = _.debounce(() => {
+			this.sourceTarget.classList.remove("is-invalid");
+			nameProblemFeedbackElement.style.display = "none";
+
+			const nameValue = this.sourceTarget.value;
+			if (nameValue.length < 1 || nameValue === originalValue) {
```
aaaaargZombies, line 37: we don't allow partner names to be less than 5 chars so we can't collide below that value.

app/javascript/controllers/partner_form_validation_controller.js
```diff
@@ -0,0 +1,62 @@
...
+				.then((response) => response.json())
+				.then((data) => {
+					if (!data.name_available) {
+						this.sourceTarget.classList.add("is-invalid");
+						nameProblemFeedbackElement.style.display = "block";
+					}
+
+					this.sourceTarget.focus();
+				});
+		}, this.inputDebouncePeriod);
```
aaaaargZombies, line 54: this is is undefined so we're not debouncing, maybe just pop the number in here instead of trying to declare it above?

aaaaargZombies (changes requested):
I've left a few notes but this is a nice solution, just a bit of polishing needed.

app/javascript/controllers/partner_form_validation_controller.js
ivan-kocienski-gfsc: Fixed.

app/controllers/admin/partners_controller.rb
ivan-kocienski-gfsc: fixed

app/views/admin/partners/_form.html.erb
ivan-kocienski-gfsc: Yeah i think this would be a tricky fix as the JS would have to figure out if the warning was already given from the failed request as plain old HTML.

app/javascript/controllers/partner_form_validation_controller.js
ivan-kocienski-gfsc: fixed

app/javascript/controllers/partner_form_validation_controller.js
ivan-kocienski-gfsc: fixed

app/views/admin/partners/_form.html.erb
kimadactyl: could it link to the partner with the same name or is that a huge faff with other implications around who can see the clashing partner? would be one way to make the warnings a bit diff at least. but i think this is fine as is too.

ivan-kocienski-gfsc:
@aaaaargZombies I bought back the debouncer. I think this is now done.

app/views/admin/partners/_form.html.erb
ivan-kocienski-gfsc: That is a neat idea but I think it may be best written in a new issue with clear specs

aaaaargZombies (approved):
LGTM

PR #2356, opened 2024-04-01
mega ical

### LINKED ISSUE #2178: [Bug]: When trying to subscribe to one site's combined ical feed, I end up subscribing to the ENTIRETY OF PLACECAL
## Description

I wanted to subscribe to see all events on PlaceCal Norwich in my calendar as an .ical feed. I went to the Events page of PlaceCal Norwich and clicked 'Subscribe to with iCal', like I would if subscribing to the feed of any individual partner. I then loaded this into my calendar and it pulled in every event from every PlaceCal site, not just PlaceCal Norwich. 

Screenshot below of what it did to my calendar, lol

Screenshot 2024-01-16 at 12 08 01

aaaaargZombies:
fixes #2178

the ical feed was being pulled from `events.all`, also found we were never declaring the title for the link to the feed.


app/controllers/events_controller.rb
```diff
@@ -32,8 +33,13 @@ def index
       end
       format.text
       format.ics do
+        events = Event.all
```
ivan-kocienski-gfsc, line 36: Would it be possible to put in a small test here? Just something that makes a few events and then checks the response is an ical feed? (maybe parse the result and check it's getting back the correct number of events?)

ivan-kocienski-gfsc (changes requested):
Looks good buuuuuuuut one suggestion

app/controllers/events_controller.rb
aaaaargZombies: done!

katjam:
@aaaaargZombies In the interest of getting this milestone verified asap, I am merging on your behalf.

PR #2369, opened 2024-04-08
Updated rollbar gem

### LINKED ISSUE #2367: [Bug]: 404 pages crash
## Description

When visiting a page like https://placecal-staging.com/foofooofoo we are not shown the "404 page not found"

Digging in to this on local host, if you remove the static html pages from the `public/` folder and hit that path the dev server just haults. So something is very wrong deep inside rails (probably because of our code?).

It's also not being logged in rollbar (on staging)

## A/C
- [ ] Visiting an unrecognised path serves the standard 404 page.

ivan-kocienski-gfsc:
Closes #2367

Bit of a rabbit hole this one: when visiting unknown routes like `/foofoofoo` or `/styleguide/styleguide/event` the server would 500 with no backtrace (probably because there was an error in the error handler). 

Turns out the rollbar gem (v3.3.0) likely had a bug in it that is fixed in the latest version (v3.5.2)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -375,7 +375,7 @@ GEM
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
     rexml (3.2.6)
-    rollbar (3.3.0)
+    rollbar (3.5.2)
     rubocop (1.62.0)
       json (~> 2.3)
       language_server-protocol (>= 3.17.0)


aaaaargZombies (approved):
I'm getting the dev 404 screen so presumably we'll get something not crashy in staging / production

image


Either way good to update this gem and I'm surprised it's having this effect.

PR #2378, opened 2024-04-09
partner tag inputs

### LINKED ISSUE #2333: Partners have separate facility, partnership and category inputs

aaaaargZombies:
fixes #2333

-  there is an input for category tags
- category input has validation for max 3 selected
- there is an input for partnerships
- partnership input has JS validation that warns a user about removing a tag from a partner
- there is an input for facility tags

## notes

I have left as much of the `tag` associations in place so that we didn't need to restructure much of placecal, the forms add to the respective STI options and that's picked up by the tags. I did a small amount of tidying of the helper function file pointed tests in it's direction but I stopped short of ripping out the partnership scoping on the tag model, that should probably done after all the other entities have been shifted over. I think in the future I would avoid STI as it has created a lot of tight coupling between things that aren't really related.

I expect people will have opinions on the exact wording of the helper text in the UI but I tried to put something sensible in as a jumping off point.

image


ivan-kocienski-gfsc (approved):
Great job here Alfie, this is superb effort.

PR #2379, opened 2024-04-09
Queue new calendars for immediate import

### LINKED ISSUE #2304: Improve UI for adding a calendar by explaining the importer is running in the background
## Acceptance Criteria

- [ ] When I update a calendar, it automatically queues for import
- [ ] I will see a confirmation message telling me the calendar is being imported and that I should check back in a few minutes

https://www.notion.so/gfsc/Kim-s-timeline-for-completion-12-3-24-d0e8ef7a03c44ba08e56d4477e3fea5d?pvs=4#92b415e5ca5d440f8840401d7dc94d1f

Maybe we can improve the UI here by letting the user know something like:
"The importer is now running, check back in 10 minutes to see if your calendar has been imported" 

Or maybe we could change the display names for the labels to something more informative than 'in worker'?

r-ferrier:
Fixes #2304 

## Description

After a bit of digging it seemed like the easiest solution here was to improve the notification shown in the header and to also immediately import a calendar's events on creation. That means that the user doesn't get to specify the date to import from on their first import (it will just default to now) but they can change that at any point after the import, and changing it to a date in the future will remove any events that have already been imported if they didn't want current events showing for any reason. We could add a field in the form to let them control when the import begins from on their first import, but that just felt like another thing for users to have to think about at the create stage when most of them probably don't really care about this, so I think this would be better as a new feature if it comes up again.

### DIFF
--- a/app/controllers/admin/calendars_controller.rb
+++ b/app/controllers/admin/calendars_controller.rb
@@ -42,7 +42,11 @@ module Admin
       authorize @calendar
 
       if @calendar.save
-        flash[:success] = 'Successfully created new calendar'
+        if @calendar.queue_for_import! true, DateTime.now
+          flash[:success] = 'New calendar created and queued for importing. Please check back in a few minutes.'
+        else
+          flash[:danger] = 'Successfully created new calendar but there was a problem importing events.'
+        end
         redirect_to edit_admin_calendar_path(@calendar)
       else
         flash.now[:danger] = 'Calendar did not save'
--- a/app/controllers/admin/calendars_controller.rb
+++ b/app/controllers/admin/calendars_controller.rb
@@ -42,12 +42,8 @@ module Admin
       authorize @calendar
 
       if @calendar.save
-        if @calendar.queue_for_import! true, DateTime.now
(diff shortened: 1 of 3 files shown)


ivan-kocienski-gfsc (changes requested):
This looks almost okay but I think I'd be happier if this logic was in the model as an `after_save` hook. It can then be tested that saving a new calendar puts it into the queue

I am not sure if that the `queue_for_import!` method returns a truthy value if it succeeds (or a falsey value if it fails).

Also if the calendar doesn't successfully get queued (which ideally should never happen) then is showing a message in the flash the most helpful feedback?

r-ferrier:
ok, have made those changes as requested - decided automatically queuing for import after any editing is always good - no one wants to be waiting around for their changes to be picked up after they've just made changes to something.

r-ferrier:
ok I've made use of the `after_create` hook instead to prevent this queueing on edit - mainly because the number of tests failing when I tried to change this was so overwhelming and the way they're being run, would be hard to change, so I decided given the timeframe that just having this update on create is enough!

ivan-kocienski-gfsc (approved):
Looks good to me

PR #2382, opened 2024-04-10
Hiding partners from the GraphQL API that are set to `hidden`

### LINKED ISSUE #2220: only partners that is `hidden == false` is visible on public transdim site
Verify/fix that API needs to behave like the display logic.

Security issue.

## A/C
- [ ] only visible partners are shown through the GraphQL API

ivan-kocienski-gfsc:
Closes #2220 

## Considerations
Can a partner be viewed through an event even if not visible?


r-ferrier (approved):
Nice! This looks great.

PR #2385, opened 2024-04-11
Prevent events from linking up to the wrong partner

### LINKED ISSUE #2348: [Bug]: Events are being matched up to partners through their address incorrectly
## Description
See #1511 for context

The following event has been matched to a partner whose location is different from the event address.

 https://climatejustice.placecal.org/events/378308

They share a street address but have different postcodes, so we think there may be a bug in how this is resolving.

r-ferrier:
Fixes #2348 

## Description

The event resolver has two strategies that can link up an event to a partner that is different to the partner they belong to.

These strategies previously tried to:
A) match with a partner by checking if the name and postcode in the event matched any partners in our DB

if this was unsuccessful they then tried to:
B) match with a partner by checking if the address matched any partners' addresses in our DB

if this was unsuccessful they then tried to:
C) create a new address from the event

Route B was broken - it would return true if just the first line of the address matched. 

I've decided to simply remove this option because I think it's poorly thought out - in our system it could be possible for multiple partners to exist at the same address and this only returns one. However partner names are unique, so if the address contains a matching partner name & postcode then we can be certain we are associating it with the intended partner.

I've also tidied up the code a little bit to remove some of the noise.

### DIFF
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -74,7 +74,7 @@ class CalendarImporter::EventResolver
 
   def event_strategy(partner, address: nil)
     if data.has_location?
-      partner = Partner.fuzzy_find_by_location(event_location_components)
+      partner = Partner.find_from_event(event_location_components, data.postcode)
       address = partner&.address || Address.build_from_components(event_location_components, data.postcode)
     elsif partner.present?
       raise Problem, WARNING2_MSG
@@ -84,7 +84,7 @@ class CalendarImporter::EventResolver
 
   def event_override_strategy(partner, address: nil)
     if data.has_location?
-      partner = Partner.fuzzy_find_by_location(event_location_components)
+      partner = Partner.find_from_event(event_location_components, data.postcode)
       address = partner&.address || Address.build_from_components(event_location_components, data.postcode)
       raise Problem, INFO1_MSG if partner.nil? && address.nil?
     elsif partner.present?
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -72,65 +72,27 @@ class CalendarImporter::EventResolver
     data.partner_id = calendar.partner_id
   end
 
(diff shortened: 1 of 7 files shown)


app/jobs/calendar_importer/event_resolver.rb
```diff
@@ -72,65 +72,27 @@ def determine_location_for_strategy
...
-      place ||= address&.partners&.first
-
-      # NOTE: What happens if address has zero partners and the earlier assignments fail?
-      #       'place' is nil in this instance
-      # NOTE: Address.search can also return Nil if there are no event_location_components, or
-      #       if the address failed to save
-      # Both of these will cause the event to fail validation with "No place or address could be created/found (etc)"
-
-    elsif place.present?
+      partner = Partner.fuzzy_find_by_location(event_location_components)
```
katjam, line 77: I think this is good for now because we have a workaround, that the event provider has to know to match exact name. It's better than erring on the side of probably as we were before. Is there somewhere in the interface to document that this has changed? Or will we update that when we finish untangling the event resolvers?

katjam, line 77: I might be missing something but it seems like this is only looking at the partner name, not postcode.

r-ferrier: argh yes. Have fixed!

r-ferrier: yes, agree. I think we should address this when we untangle the event resolvers/ sort out the UI - it should probably explicitly state that events will be matched to partners

aaaaargZombies:
I think this looks good but I'm hesitant to approve until @katjam has had a chance to see the corrections they noted.

katjam (approved):
I'm happy with this as long as there is a ticket open to rewrite the UI help text in the Location Strategy section with acceptance criteria that includes how to construct an Event address so that it gets matched with a partner & that the partner has to be open to including 3rd party events.

r-ferrier:
This ticket covers UI changes needed for this and the subsequent related event/partners ticket: https://github.com/geeksforsocialchange/PlaceCal/issues/2403

PR #2386, opened 2024-04-11
show events on index if possible

### LINKED ISSUE #1987: show future events on site index
## User story

As a site admin I need to be able to set the default view to be weekly so that the site appears busy and in-use even with relatively few partners.

## Acceptance criteria

- [ ] a site with future events always shows events on first load of the events index
- [ ] default to showing all future events
- [ ] fall back to day as the default if future events is `>200`
- [ ] if showing future events hide the date pills
- [ ] add "show all" option to filter menu

Possible fixes from honor

- the default view for events is configurable to be weekly or monthly by the site admin and doesn't reset to daily every time the listings page is visited
- alternatively, it shows something smart depending on event count? (e.g. weekly view if theres less than 10 events a week, monthly if less than 20 events a month, etc). poss too fiddly to work out an algo?

The later option sounds most useful but we need to investigate alternative forms of pagination, https://transdimension.uk/events `future events` option seems like a sensible but would require some re-working of the UI

image

Probably add `future` after `weekly` and dropping the pagination at the top ?

Set `day | week | future` based on min `n` events in controller.

## Implementation plan

aaaaargZombies:
fixes #1987

Minimal intervention to show events on first page load.

Filter option now includes a "Show all" option, this is the pages default state unless there are more than 200 events then you have to explicitly select it.

On some instances - Manchester especially - showing all is very slow with over 6000 events.

## potential areas for future improvement (out of scope in this PR)

- work to surface stale data being created by automated systems left monitored ( see [comment](https://github.com/geeksforsocialchange/PlaceCal/issues/1987#issuecomment-2049467630))
- provide a pagination that isn't date based. ie next 100 events.
- better handling of blank days for example
  - skip to next day/week with an event link on pages that load with no data 
  - inlcude `n` of events indicator in pagination link 
  - don't include nav links in the paginator to empty days.
- re-think filter / pagination UI, @honor-gfsc has noted it is non-obvious.


app/controllers/events_controller.rb
```diff
@@ -14,10 +14,15 @@ class EventsController < ApplicationController
...
-    @period = params[:period].to_s || 'day'
+    # Duration to view - default to future view
+    @period = params[:period] || 'future'
     @repeating = params[:repeating] || 'on'
     @events = filter_events(@period, repeating: @repeating, site: current_site)
+    # Duration to view - default to day view if there are too many future events
+    if params[:period].to_s == '' && @events.count > 200
+      @period = 'day'
+      @events = filter_events(@period, repeating: @repeating, site: current_site)
+    end
```
r-ferrier, line 25: This is a little picky buuuut this kind of variable reassignment makes it much harder to understand what is going on and also you don't need to reassign events as it's identical - any chance you could refactor this to something more like:

```
should_default_to_day = params[:period].to_s == '' && @events.count > 200
@repeating = params[:repeating] || 'on'
@events = filter_events(@period, repeating: @repeating, site: current_site)
@period = should_default_to_day ? 'day' :  (params[:period] || 'future')
```

r-ferrier: Also also - maybe we should default to weekly here rather than daily? Lots of places have repeating events and just one on lots of days which still makes it look a bit crap by default - for example torbay

r-ferrier (approved):
I think this works well, just a few qs & suggestions but nothing essential - up to you!

app/components/paginator/paginator_component.rb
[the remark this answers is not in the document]
aaaaargZombies: yes it's because someone redefined `period` to a different value here

https://github.com/geeksforsocialchange/PlaceCal/blob/10ada4821ef691a73777fa8fa98ba04dd925e843/app/components/paginator/paginator_component.rb#L52-L54

so now `step` is a unit of time and `period` is the string/key used consistently throughout. I hope that makes sense.

app/controllers/events_controller.rb
aaaaargZombies: I'll default to week, that makes it more likely to not give a blank, thanks.

> you don't need to reassign events as it's identical

I don't think that's the case because the `@events` inside the `if` block will shift the period from `future` to `day` so should *hopefully* produce a different result.

I'm afraid I'm struggling to get your snippet to produce the same result

In your code block I can't generate the `should...` var without first declaring `@events`, then again I need to set the `@period` before assigning `@events` or they won't be parametrized by it.

r-ferrier: ah yes! Makes sense.

Issue #2364, opened 2024-04-04
[Bug]: fix text on partner missing location error message


ivan-kocienski-gfsc:
## Description

![Screenshot_2024-04-04_14-12-34](https://github.com/geeksforsocialchange/PlaceCal/assets/109223824/2f3aac39-7c3e-4b43-aed8-33d1f0f4ba20)

"...This usually means one of *two* things:"

aaaaargZombies:
## verified

![image](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/d755baba-5dec-45ac-81f2-0945774ad6de)

PR #2388, opened 2024-04-11
Address fields are not required on the partner form

### LINKED ISSUE #2310: [Bug]: It is not clear when creating a partner that you can make one with just a service area

aaaaargZombies:
fixes #2310 

- text approved from previous pr  #2317 
- removed required status from 1st line of address and postcode field

## Notes

If someone misses out both these fields when adding the address, and they have a service area, it fails silently (refreshing the page but no warning). An unlikely event but it does not allow the creation of a malformed address.

There is a related issue #2364 that is outside the scope of the PR

### DIFF
--- a/app/views/admin/partners/_address_fields.html.erb
+++ b/app/views/admin/partners/_address_fields.html.erb
@@ -8,7 +8,9 @@
 
   <%= address_form.input :street_address,
       class: "form-control address_1 address_field",
-      label: 'Street address' %>
+      label: 'Street address',
+      required: false
+    %>
 
   <%= address_form.input :street_address2,
       class: "form-control address_2 address_field",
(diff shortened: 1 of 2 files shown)


r-ferrier (approved):
I think this is more helpful even with the silent fail possibility, but maybe add this into the other issue so it's more of a 'fix validation in general' kind of ticket? I feel like a flash should tell you at least that an address has not been saved.

aaaaargZombies:
OK will merge this as is and have opened this issue #2391 so we can decide where it sits compared to our other priorities.

Issue #2305, opened 2024-03-13
[Bug]: It's not clear to users that their calendar is failing/ how to reimport it


r-ferrier:
This issue is about the error state for a calendar that has failed import

## Story
Maybe we could improve the UI around this again - some more informative text? A clearer definition of what the problem is and how it can be fixed by the admin user.

## Further Context
https://www.notion.so/gfsc/Kim-s-timeline-for-completion-12-3-24-d0e8ef7a03c44ba08e56d4477e3fea5d?pvs=4#bb7290e3d3e449df95bcd0cac435d34f

Blocked by #2304 which is the loading state for this UI component.

## Acceptance Criteria
- [x] Once saved, the status of the calendar is visible at the top of the page
- [x] I can see details about the status of my calendar near the lozenge
- [x] The details about status include lozenge, date updated, description (including notices & any error details)
- [x] Placement & design of error description annotices TBC... we might want to look at formatting notices as a separate issue
- [x] Once saved, I can always see re-import button & trigger re-import
- [x] I **cannot** schedule a timed import for my calendar

If the calendar has:
- [ ] Not finished importing - `in_queue` or `in_worker` (loading) lozenge & notice to check back in 10 minutes
- [ ] Event problem - `idle` (success) lozenge & I see notices for the failing events & a message suggesting I fix then re-import
- [ ] Invalid response - `error` (error) lozenge & notice to (TBC)
- [ ] Inaccessible feed - `bad_source` (error) lozenge & notice to (TBC)
- [ ] ActiveRecord error - `error` (error) lozenge & notice to (TBC)
- [ ] #2399

## Plan
- [x] Change status lozenge copy from (idle --> success, in_queue --> loading, in_worker --> loading, error --> error, bad_source --> error)
- [x] Button copy from "Queue for import now" to Re-import calendar
- [ ] more tbc...

Placement of status block:
![image](https://github.com/geeksforsocialchange/PlaceCal/assets/861172/3ff14682-ab03-4c6d-a24c-2ad24bd57124)

Status block changes:
![image](https://github.com/geeksforsocialchange/PlaceCal/assets/861172/20b5d49b-219f-4695-8f8a-8f85e7a52b07)

ivan-kocienski-gfsc:
TODO: 

How can a calendar fail? What actions can be taken by the user?

r-ferrier:
@katjam & @ivan-kocienski-gfsc to discuss the approach to this and work out AC

[...]

PR #2392, opened 2024-04-15
Calendars clear importer problems on source value change

### LINKED ISSUE #1601: [Bug]: Calendar update status incorrectly persists after URL is changed
## Description

Trying to update the Yard's calendar and it's showing me incorrect older error messages

## Steps to reproduce

1. Replace the URL with a new feed
2. Press save
3. Error shows up - which is the old error
 
This does reset after next import but I think it's very confusing when people are trying to debug this

Screenshot 2022-12-01 at 10 24 46

Errors should reset when the URL field is updated

## A/C
- [ ] When a calendar has an error message on it, changing the source URL and saving will clear that message (and any associated fields)

ivan-kocienski-gfsc:
Closes #1601

### DIFF
--- a/app/models/calendar.rb
+++ b/app/models/calendar.rb
@@ -22,6 +22,7 @@ class Calendar < ApplicationRecord
 
   validate :check_source_reachable
 
+  before_save :clear_status_on_source_change
   before_save :update_notice_count
 
   # Output the calendar's name when it's requested as a string
@@ -271,5 +272,13 @@ class Calendar < ApplicationRecord
     errors.add :source, "The source URL returned an invalid code (#{e})"
   rescue CalendarImporter::Exceptions::UnsupportedFeed => e
     errors.add :source, 'Unable to autodetect calendar format, please pick an option from the list below'
+  end
+
+  def clear_status_on_source_change
+    return unless source_changed?
+
+    self.critical_error = nil
+    self.notices = nil
+    self.last_import_at = nil
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/calendar.rb
```diff
@@ -272,4 +273,12 @@ def check_source_reachable
...
     errors.add :source, 'Unable to autodetect calendar format, please pick an option from the list below'
   end
+
+  def clear_status_on_source_change
+    return unless source_changed?
+
+    self.critical_error = nil
+    self.notices = nil
+    self.last_import_at = nil
+  end
```
aaaaargZombies, line 283: This looks like it works though is it missleading to imply the issue has been fixed when it's not been tested yet? @kimadactyl  is this what you wanted?

I suppose the alternative is not clearing the issue and making it clear that you need to wait for the importer to run to findout if the changes have been effective?

happy for this to go in, just double checking before we commit to it.

kimadactyl: yeah hmmm lots of ways you could argue this

i think it prob makes most sense to change the messages to something like "Source URL changed, waiting for queue" and not to blank the last_import?

ivan-kocienski-gfsc: If you feel this is the desired behaviour then the A/C need to be (possibly discussed) and updated

aaaaargZombies (approved):
Approving based on the AC in the ticket - will discuss at dev standup if it's good AC then we can merge straightaway if everyone is on the same page.

Apologies for throwing doubt on the AC after the work has been done, I appreciate that's frustrating.

kimadactyl:
> Apologies for throwing doubt on the AC after the work has been done, I appreciate that's frustrating.

I think the AC were written after the code was done.

It's fine as is though.

aaaaargZombies:
After discussion Ivan pointed out we have this ticket that is about handling failing imports so the concerns can be covered there #2305

PR #2395, opened 2024-04-15
More secure feedback message on password reset

### LINKED ISSUE #2225: [Bug]: Should not be indicating which emails have accounts
## Description

When a user applies to reset password, the response should not reveal whether their email exists in the system

image

ivan-kocienski-gfsc:
Closes #2225 

If a password reset account is found (or not) show the same feedback message. This way no private information is being leaked in public.

The reset now looks like this regardless of whether or not the email exists:
Screenshot_2024-04-15_14-13-09

I also checked the log in page and confirmed that does not have the same problem (the message is always the same for either email or password)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -62,7 +62,13 @@ Metrics/ClassLength:
   Exclude:
     - "test/**/*.rb"
 
+Metrics/CyclomaticComplexity:
+  Exclude:
+    - app/controllers/application_controller.rb
 
+Metrics/PerceivedComplexity:
+  Exclude:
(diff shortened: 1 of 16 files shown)


aaaaargZombies:
image

ivan-kocienski-gfsc:
I fixed the A/C so this is okay (and it is okay)

The important thing is that we verify we aren't potentially leaking any information here

aaaaargZombies (approved):
LGTM - agreed the leaking the email was more important than the specific wording and I like the current one, to the point.

PR #2402, opened 2024-04-16
Partners control event connections

### LINKED ISSUE #2381: Give partners control over third party event listings
## Description

From conversations with Rakesh: Having an events grouped by partner is useful, having events grouped by place is not useful. The link between partner and event should be based on who created the event, not where the event is based.

## Acceptance criteria
- [ ] as a partner admin I can decide whether or not my partner displays events from other partners' calendars
- [ ] If a partner allows 3rd party events to be listed under them and there is an Event from another partner with their address, public can see the event under both partners
- [ ] As a partner admin, it is clear to me from the UI that I can toggle on and off 3rd party events listings for all partners, not individual partners
- [ ] As a partner admin, when I toggle on and off 3rd party event listing allowed, the events appear and disappear immediately (i.e. the relationship of the event with my partner is not determined on what my setting is at the time of import)
- [ ] The calendar importer only parses the raw event data into the database. e.g. It does not make links between entities.
- [ ] In the case that an event is additionally listed by a "host venue" partner outside the geographical area of the partnership/neighborhood admin, the admin will not be able to see the partner in their admin interface (though obviously they can see the listing on the public site)

## Related
- [ ] #2403

## Plan
- [ ] A partner has a field `can_be_assigned_events` with a description `Events that have the same address as this partner will be assigned to this partner on PlaceCal` (copy tbc)
- [ ] If `partner.can_be_assigned_events` then when the calendar importer runs, it is allowed to make connections between partners and events if they share the same address
- [ ] The calendar importer matches addresses from events to partners, and not to conceptual places
- [ ] Addresses are not shared but are matched on the fly by the code

## Blocked by this/ should be fixed as part of this
- [ ] https://github.com/geeksforsocialchange/PlaceCal/issues/2348
- [ ]  #2372

r-ferrier:
Fixes #2381 

## Description

- Adds a field to partners 'can be assigned events' that enables users to decide whether or not their partner can be linked up to events imported from other calendars
- Events imported with a strategy that takes the address from an event no longer attach the event to partners whose address matches - they just give the event their own address
- In the front end, when an event's address is shown:
  - if it was imported to display at the partner's location then that partner is shown OR
  - if it matches the name and postcode of a partner who can be assigned events, then that partner is shown OR
  - the event's own address is shown
- changes & adds some tests to assert the above is working as it should

### still to do:
- [ ] give every event its own address in the event importer
- [ ] assign the address to display based on the calendar importer strategy rather than assigning 'places' in the importer
- [ ] refactor event.place and calendar.place as methods on these classes that return the associated partner if a calendar's import strategy allows it

### DIFF
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -75,6 +75,7 @@ class CalendarImporter::EventResolver
   def event_strategy(partner, address: nil)
     if data.has_location?
       address = Address.build_from_components(event_location_components, data.postcode)
+      partner = nil
     elsif partner.present?
       raise Problem, WARNING2_MSG
     end
@@ -84,6 +85,7 @@ class CalendarImporter::EventResolver
   def event_override_strategy(partner, address: nil)
     if data.has_location?
       address = Address.build_from_components(event_location_components, data.postcode)
+      partner = nil
       raise Problem, INFO1_MSG if address.nil?
     elsif partner.present?
       address = partner.address
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -6,6 +6,8 @@ class CalendarImporter::EventResolver
   WARNING2_MSG = 'Could not determine where this event is. A default location is set but the importer is set '  \
                  'to ignore this. Add an address to the location field of the source calendar, or choose '      \
                  'another import strategy with a default location.'
+  INFO1_MSG = 'This location was not recognised by PlaceCal, please update the location field of the source ' \
+              'calendar, or choose another import strategy with a default location'
 
   attr_reader :data, :uid, :notices, :calendar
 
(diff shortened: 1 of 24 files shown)


app/jobs/calendar_importer/event_resolver.rb
```diff
@@ -180,10 +161,6 @@ def save_all_occurences
       unless event.update(attributes)
         notices << event.errors.full_messages.join(', ')
       end
-
-      if event.address_id.blank? && calendar.strategy == 'event'
```
ivan-kocienski-gfsc, line 184: I think I remember you saying that events will no longer require an address.

(So can only be matched through partner)

r-ferrier:
I think I'm going to merge this and then I can do the calendar verification ticket and the bug I raised afterwards. I'm still unsure about the exceptions I'm raising in here - it's not clear to me if these stop an event from being loaded or throw the whole calendar off, but doing the testing with this on staging should help to clarify these things, so I'm going for it. Can always roll it back if we don't like it!

PR #2405, opened 2024-04-16
 handle failing partner address gracefully

### LINKED ISSUE #2391: [Bug]: partner lacks warning when failing on address
## Description

small amount of tech debt introduced in https://github.com/geeksforsocialchange/PlaceCal/pull/2388 , We removed the required fields on address but they still get validated on the address model.

It would be good to flash a warning if they fail rather than just not save the partner.

## Steps to reproduce

1. create a partner with no service area
1. only add in the middle fields for address
2. save the partner

- Recieve a warning about how to write a valid address
- Recieve a warning about needing either address or service area

## Platform (if relevant)

aaaaargZombies:
fixes #2391 

We basically had two ways of intercepting the validation and both were covering half the use cases but not working in unison which was quite confusing. Now we always validate address creation from a partner but we never attempt to create an address if all the fields are blank.

This means we don't get silent failure if someone puts the first line of their address in line two by mistake, we get warnings at the top of the page and highlighted fields that need attention and we retain user input even if it's bad.

### DIFF
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -51,7 +51,17 @@ class Partner < ApplicationRecord
 
   accepts_nested_attributes_for :calendars, allow_destroy: true
 
-  accepts_nested_attributes_for :address, reject_if: ->(c) { c[:postcode].blank? && c[:street_address].blank? }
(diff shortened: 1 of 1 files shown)


r-ferrier (approved):
LGTM!

PR #2411, opened 2024-04-17
show all should show all future events

### LINKED ISSUE #2410: [Bug]: "show all" doesn't show all events if date range in url
## Description

This might be desired behaviour but I think it's confusing.

If you select a date using the pagination then change to `future` / "show all" in the form, it shows you the future from that date. This may be no events even though events exist on the system.

Should probably change from to future or have date reset.

https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/9234169e-001c-4bf4-814c-38bd3059972d

## Acceptance Criteria

- [ ] When I select all events it shows me from today even if I have been browsing by dates in the future

aaaaargZombies:
fixes #2410 

If period is `future` we always consider the current_day to be today regardless of url params.

## thoughts

I briefly toyed with resetting the url because I thought it would be cleaner but I realized it meant you couldn't toggle between views without losing state.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -64,7 +64,7 @@ class ApplicationController < ActionController::Base
     events = events.one_off_events_first if repeating == 'last'
     events =
       if period == 'future'
-        events.future(@current_day).includes(:place)
+        events.future(@today).includes(:place)
       elsif period == 'week'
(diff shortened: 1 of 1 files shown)


ivan-kocienski-gfsc (approved):
Looks good!

PR #2412, opened 2024-04-17
simple create partner form

### LINKED ISSUE #970: Make the create partner form more user friendly
## Description

The create partner process is very long. It's currently possible to spend a lot of time entering data before finding that you can't create a partner in that area, or that one exists with the same name.

Some of this has been addressed by: #2352, #2350

image

## Acceptance criteria

- [ ] The new partner form has only required fields: Name, Place & Partnership (if permitted)
- [ ] The Create partner re-directs to edit for that new partner

## Future considerations (TBC)
- [ ] There are 2 save buttons on new partner form (save & create another partner) (save & edit this partner) or does that complicate things
- [ ] #2408

aaaaargZombies:
fixes #970

- just essential details for partner creation on the new partner form
- redirects to full form / edit page after creation

## notes

I added in partnerships to the AC / form because if a partnership_admin creates a partner without connecting to their partnership they will not be able to access them (I think it also fails verification). This only shows for partnership_admins and roots.

It wasn't specified in the AC but I removed the `setup` related code as this is now a duplicate of `new`


ivan-kocienski-gfsc (approved):
Looks good!

Nice work removing the setup thing.

PR #2413, opened 2024-04-18
Tweak to how partner sites are linked

### LINKED ISSUE #2288: [Bug]: Link in Partner header to site where it appears needs to be absolute
## Description
Using the link in the header to view PlaceCal instance where a Partner appears

image

Is coming up as: https://admin.placecal.org/partners/ambitions-graveyard/norwich.placecal.org

Should be: https://norwich.placecal.org

## A/C
- [x] partner site links should be valid URLs
- [x] text leads with "View this partner on: ... (opens in new tab)"
- [x] links open in new tab
- [x] there is text saying "open in new tab"

ivan-kocienski-gfsc:
Closes #2288 

Have updated link view:
- change of text
- will now open in new tab

Also note that the bug was actually due to bad data, which has now been fixed on production and staging. This will still be invalid for any local snapshots devs use.


aaaaargZombies (approved):
LGTM

PR #2421, opened 2024-04-22
Be more permissive about calendar event URLs

### LINKED ISSUE #2417: [Bug]: URL fields from non-Google calendars are not imported
## Description

When adding URL information to a non-google .ics feed this is not picked up by the PlaceCal importer

## Platform (if relevant)

Example non-Google .ics feed https://ambition.rip/public_calendar.ics
PlaceCal event which has not pulled through the URL https://norwich.placecal.org/events/380062

## A/C

- [ ] events imported from an ical feed containing a `URL` field have the information from this field imported and stored
- [ ] events with online information are flagged as such in the events index
- [ ]  change "stream" to "button" on button on Rosie's screenshot in comment below

ivan-kocienski-gfsc:
Closes #2417 

When importing iCal feeds, detect any URL in the URL field and mark it as `indirect`

### DIFF
--- a/app/jobs/calendar_importer/events/ics_event.rb
+++ b/app/jobs/calendar_importer/events/ics_event.rb
@@ -50,42 +50,44 @@ module CalendarImporter::Events
 
     def online_event?
       # Either return the google conference value, or find the link in the description
-      link = @event.custom_properties.fetch 'x_google_conference', nil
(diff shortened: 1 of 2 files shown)


r-ferrier (approved):
i like it! let's get it in and we can validate it.

PR #2424, opened 2024-04-23
2373 calendar last updated

### LINKED ISSUE #2373: Show the last time a calendar's data was updated
## User story

As an organiser, I want to see when calendars were last updated by the partners (i.e, when events were last added or edited at the source level), in order that I can identify who hasn't updated them for a long time and know which listings are fresh

_Generated at [Cycle 29 scoping](https://www.notion.so/gfsc/Scoping-PlaceCal-Accuracy-Reliability-Consent-cycle-d04f48d75fd84acfbf3b9b9bdaca86db?pvs=4)_

## AC

- [ ] I can see the date my calendar was last updated in the calendar status section
- [ ] On the calendar status, if it was last updated more than 6? months ago, I can see a warning asking me if it is current
image
- [ ] On the calendar index page, the last updated field shows the time of checksum, not when the calendar was saved
image
image

## desirable outcomes pending research

- [ ] Calendars show the last time the source data was changed (where possible)
- [ ] It's possible to view this information in the calendars index and show pages
- [ ] We clean up the nomenclature a bit - the current "last updated" probably refers to when the model was last modified, not when the source data was changed
 
Note this is probably quite complex and source-dependant. I think some functionality is better than none. We can do based on the hash of the input changing, on the last non-repeating event, etc.

Our current approach is to spawn repeating events up to a year in advance.

## Implementation plan

- add date of checksum to modal and record when we update.
- seperate out completion of importer and updating of checksum
- prevent attempts to hash `nil` on failed `download_calendar`
- sanatize ical feeds
- surface this info on admin UI

re-ticket as not essential for AC

- ignore unused feeds for now, open techdebt tickets to ensure they meet the same standard as the above tested sources.
- ideally on index and allow sort by date so stale calendars can be weeded out.
- should probably handle checksums on a per parser basis in case there's sources that aren't capable of producing usable checksums
- there should be a method of communicating if a parser has a valid method for generating a checksum

aaaaargZombies:
fixes #2373 

- added date of checksum to modal and record when we update.
- set this on all existing calendars to be the same as the import date as that's when the checksum would have been made before
- update the checksum as soon as we can / separate it out from the importer completion
- I thought I would need to stop it hashing a failed network request but the whole job stops running if that happens so no problem
-  remove request date from ical feeds, only effecting google but that makes up a very high percentage of our feeds.
- add a test to show importer / checksum increment independantly
- update ui to surface info
  - I changed "events changed" to "importer ran" which differs from the AC but that date can be incremented even if the import finds no new events.
  - I referenced "source data" / "source" as I thought that might make more sense than "checksum"
  - warning shows up on the calendar form (not index) if either import or checksum date are older than  6 months

We unfortunately won't get a sense of how out of date stuff is for a while as the two figures will only start diverging from release date.

form view

image

sites calendar index

image

Calendars index view

image

6 month warning

image

### DIFF
--- a/app/views/admin/calendars/_form.html.erb
+++ b/app/views/admin/calendars/_form.html.erb
@@ -5,7 +5,7 @@
 </p>
 
 <%# don't show importer stats if it's never run %>
-<% unless @calendar.last_import_at == nil %>
+<% if @calendar.last_import_at.present? %>
   <%= render('importer_overview', calendar: @calendar) %>
 <% end %>
 <hr>
--- a/app/views/admin/calendars/_form.html.erb
+++ b/app/views/admin/calendars/_form.html.erb
@@ -4,7 +4,10 @@
       class: "btn btn-sm btn-primary" unless @calendar.new_record? %>
 </p>
 
-<%= render 'importer_overview', calendar: @calendar %>
+<%# don't show importer stats if it's never run %>
+<% unless @calendar.last_import_at == nil %>
+  <%= render('importer_overview', calendar: @calendar) %>
+<% end %>
 <hr>
 
 <div id="js-calendar-form">
--- a/db/migrate/20240422092628_set_calendar_checksum_date.rb
+++ b/db/migrate/20240422092628_set_calendar_checksum_date.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class SetCalendarChecksumDate < ActiveRecord::Migration[7.1]
+  def change
+    Calendar.update_all('checksum_updated_at=last_import_at')
+  end
+end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -373,6 +373,7 @@ Rails/SkipsModelValidations:
     - 'lib/tasks/fixes/users.rake'
     - 'db/migrate/20240125175508_partner_hidden_attribute_set_value.rb'
     - 'db/migrate/20240411200641_partner_can_be_assigned_events_set_value.rb'
+    - 'db/migrate/20240422092628_set_calendar_checksum_date.rb'
 
 # Offense count: 22
 # This cop supports unsafe autocorrection (--autocorrect-all).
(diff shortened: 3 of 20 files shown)


db/migrate/20240422092628_set_calendar_checksum_date.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class SetCalendarChecksumDate < ActiveRecord::Migration[7.1]
+  def change
```
ivan-kocienski-gfsc, line 4: this should just run on `up` not `change`

app/views/admin/calendars/_form.html.erb
```diff
@@ -4,7 +4,10 @@
       class: "btn btn-sm btn-primary" unless @calendar.new_record? %>
 </p>
 
-<%= render 'importer_overview', calendar: @calendar %>
+<%# don't show importer stats if it's never run %>
```
ivan-kocienski-gfsc, line 7: why not?

the partial only renders if the calendar has a model behind it and if the importer has never run then it should just be in 'idle' mode.

app/views/admin/calendars/_form.html.erb
```diff
@@ -4,7 +4,10 @@
...
+<% unless @calendar.last_import_at == nil %>
```
ivan-kocienski-gfsc, line 8: i think `<% if @calendar.last_import_at.present? %>` might be more descriptive?

ivan-kocienski-gfsc (changes requested):
Just some small things I noticed

app/views/admin/calendars/_form.html.erb
aaaaargZombies: if it's not imported yet then `last_import_at` is `nil` and all the date related code goes wonky

ivan-kocienski-gfsc (approved):
This looks okay.

PR #2436, opened 2024-04-25
Simplify error text

### LINKED ISSUE #2364: [Bug]: fix text on partner missing location error message

r-ferrier:
Fixes #2364

## Description

I've tried to rewrite this as simply as possible to make it easier to understand what could be wrong from a glance - the previous message felt very long. Happy to can this if people think it's not an improvement.

I haven't touched the partner hidden thing - we're just opening a new ticket for that as a separate issue instead, because it needs some design thinking.


ivan-kocienski-gfsc (approved):
Much nicer

PR #2439, opened 2024-04-25
Handle events with URLs in 'location' field

### LINKED ISSUE #2437: Event location URLs
## User story

As a user who has set up a calendar with events that have URLs as their location, I want these URLs to be links on the event / partner pages

## A/C
- [ ] any event that has a URL location should have a link to that URL and be marked as "online" (regardless of PlaceCal calendar strategy).

ivan-kocienski-gfsc:
Closes #2437 

If an event has a URL as its location field, make it in to an Online event. 

Also the location is still the URL (on the show event page) but now it is a link

Screenshot_2024-04-25_15-26-08

### DIFF
--- a/app/components/address/address_component.rb
+++ b/app/components/address/address_component.rb
@@ -12,6 +12,10 @@ class AddressComponent < MountainView::Presenter
       return address_lines.join(', <br>').html_safe
     end
 
+    uri = URI.parse(raw_location)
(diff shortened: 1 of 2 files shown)


aaaaargZombies (approved):
This looks good, will be nice to highlight more events with online options

PR #2456, opened 2024-05-01
Fix partner creation validations

### LINKED ISSUE #2454: [Bug]: Partnership admins cannot create new partners
## Description

trying to create a new partner. 2 things seem to be going wrong:

1. Partnership is not auto-filled
2. Partner can't be saved - see error

Screenshot 2024-05-01 at 11 41 20

## Steps to reproduce

1. Logged in as a partnership admin (sid)
2. Created a new partner
3. Pressed save

1. Partnership is auto-filled
2. Partner can be saved

## Platform (if relevant)

kimadactyl:
- Added back in a gem that drops console into the error pages (unclear why this was removed)
- Did some refactoring to reduce duplicate code
- Fixed validation error
- Renamed methods to make clearer
- Create partner form now properly pre-populates

Fixes #2454

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,6 +94,7 @@ end
 
 group :development do
   gem 'better_errors'
+  gem 'binding_of_caller'
   gem 'foreman'
   gem 'graphiql-rails'
   gem 'letter_opener'
--- a/app/javascript/controllers/partner_tags_controller.js
+++ b/app/javascript/controllers/partner_tags_controller.js
@@ -4,8 +4,6 @@ export default class extends Controller {
 	static values = { permittedTags: [String] };
 
 	connect() {
-		console.log("PartnerTagsController.connect");
-
 		const getSelectValues = (select) => {
 			return [...(select && select.options)].reduce((accumulator, option) => {
 				if (option.selected) {
--- a/app/javascript/controllers/partner_tags_controller.js
+++ b/app/javascript/controllers/partner_tags_controller.js
@@ -4,6 +4,8 @@ export default class extends Controller {
 	static values = { permittedTags: [String] };
 
 	connect() {
+		console.log("PartnerTagsController.connect");
+
 		const getSelectValues = (select) => {
 			return [...(select && select.options)].reduce((accumulator, option) => {
 				if (option.selected) {
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -112,7 +112,7 @@ class Partner < ApplicationRecord
 
   validate :three_or_less_category_tags
 
-  validate :partnership_admins_must_add_tag, on: %i[create]
+  validate :partnership_admins_must_add_partnership, on: %i[create]
 
   validate :must_give_reason_to_hide
 
@@ -507,13 +507,13 @@ class Partner < ApplicationRecord
     errors.add :categories, 'Partners can have a maximum of 3 Category tags'
   end
 
-  def partnership_admins_must_add_tag
+  def partnership_admins_must_add_partnership
     return if accessed_by_user.nil? # HACK: to stop factory breaking tests
     return unless accessed_by_user.partnership_admin?
 
-    if tags.any?
-      accessed_by_user.tags.each do |t|
-        return true if tags.include? t
+    if partnership_ids.any?
+      accessed_by_user.tags.pluck(:id).each do |t|
+        return true if partnership_ids.include? t
       end
     end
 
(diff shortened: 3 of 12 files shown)


kimadactyl:
@katjam, this fixes the issue but I can't work out what's wrong with the preselection thing.

it seems to be loading correctly and then select2 is overwriting it to be blank. however, it works on the edit page - just not the new page. any ideas? i can't see what's different!

https://github.com/geeksforsocialchange/PlaceCal/blob/2bea21081710aa34c24b9ca11111c71b9dc33dbd/app/javascript/controllers/partner_tags_controller.js#L4

kimadactyl:
Never mind - fixed it! Ready for review and merge :)

Gemfile
```diff
@@ -94,6 +94,7 @@ end
 
 group :development do
   gem 'better_errors'
+  gem 'binding_of_caller'
```
katjam, line 97: Remove this is for debug only, not production.

app/javascript/controllers/partner_tags_controller.js
```diff
@@ -4,6 +4,8 @@ export default class extends Controller {
 	static values = { permittedTags: [String] };
 
 	connect() {
+		console.log("PartnerTagsController.connect");
```
katjam, line 7: Remove debug.

app/models/partner.rb
```diff
@@ -507,13 +507,13 @@ def three_or_less_category_tags
...
-  def partnership_admins_must_add_tag
+  def partnership_admins_must_add_partnership
     return if accessed_by_user.nil? # HACK: to stop factory breaking tests
     return unless accessed_by_user.partnership_admin?
 
-    if tags.any?
-      accessed_by_user.tags.each do |t|
-        return true if tags.include? t
+    if partnership_ids.any?
+      accessed_by_user.tags.pluck(:id).each do |t|
```
katjam, line 515: So on the user still has `tags` which may match a `partnership.id` not `partnerships` like partner. Do we want to keep it like this or is it confusing?

katjam: Is there any danger it matches the id of another type of tag? Just riffing here - haven't dug into the code change from when we split the partner tags. I', assuming all other entities kept the generic tag relation instead.

app/views/admin/partners/new.html.erb
```diff
@@ -52,24 +52,15 @@
...
-            "partner-tags-permitted-tags-value": permitted_options_for_partner_tags 
-          } } %>
-      <hr>
-    <% end %>
+    
     <br>
     <div class="row">
       <div class="col-sm-12">
-        <%= f.submit "Save Partner", class: "btn btn-primary btn-lg" %><br><br><br>
+        <%= f.submit "Save and continue...", class: "btn btn-primary btn-lg" %><br><br><br>
```
katjam, line 63: Nice.

test/models/partner_test.rb
```diff
@@ -420,7 +420,7 @@ class PartnerTest < ActiveSupport::TestCase
 
   test 'partnership_admin can create a partner that is part of their partnership' do
     pa = create(:partnership_admin)
-    partner = create(:partner, :accessed_by_user => pa, :tags => [pa.tags.first])
+    partner = create(:partner, :accessed_by_user => pa, :partnerships => [pa.partnerships.first])
```
katjam, line 423: So this hasn't been working since https://github.com/geeksforsocialchange/PlaceCal/pull/2378 when we did the tag refector?

Gemfile
kimadactyl: This will only install for development don't worry!

app/javascript/controllers/partner_tags_controller.js
kimadactyl: Will do, there was one with and one without so will switch to both without instead.

app/models/partner.rb
kimadactyl: I think this can poss be refactored to remove reference to `tags` yes but for now it works and we already have a ticket open to resolve removing all refs to `tags` :)

test/models/partner_test.rb
kimadactyl: I think so.

kimadactyl: There are no system tests for creating a new partner as far as I can tell. I think the current test suite separation of concerns needs some attention.

katjam (approved):
Let's give it a whirl.

PR #2457, opened 2024-05-01
Strip whitespace around partner, calendar and address information

### LINKED ISSUE #2450: [Bug]: Remove white space around partner name
## Description

We have one "The Proud Trust" and one "The Proud Trust ". This has allowed it to be duped. Suggest we trim whitespace from before and after partner names.

Screenshot 2024-04-29 at 15 47 23

Partner names can't have white space before or after them.

### LINKED ISSUE #2452: [Bug]: Trim whitespace around address lines
## Description

Noting there's some whitespace issues here on partner addresses

Screenshot 2024-04-30 at 11 28 26

Screenshot 2024-04-30 at 11 31 21

## Steps to reproduce

1. What you did first
1. What you did next

Address lines are stripped of whitespace

kimadactyl:
This is causing a few issues:

- Partners can be created with dupe names but for a space
- Unnessecary validation messages are popping up for url fields with a whitespace before or after when we can just deal with this for people
- Addresss are outputting slightly strangely if they have spcaes at the end of lines

This is just a bit of a QoL tidyup improvement that will solve some of the issues we're seeing from live partners.

Fixes #2450 
Fixes #2452

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,6 +73,7 @@ gem 'graphql'
 gem 'rack-cors', require: 'rack/cors'
 
 # Utilities
+gem 'auto_strip_attributes'
 gem 'bootsnap', require: false
 gem 'enumerize'
 gem 'friendly_id'
--- a/Gemfile
+++ b/Gemfile
@@ -94,6 +94,7 @@ end
 
 group :development do
   gem 'better_errors'
+  gem 'binding_of_caller'
   gem 'foreman'
   gem 'graphiql-rails'
   gem 'letter_opener'
--- a/app/javascript/controllers/partner_tags_controller.js
+++ b/app/javascript/controllers/partner_tags_controller.js
@@ -4,6 +4,8 @@ export default class extends Controller {
 	static values = { permittedTags: [String] };
 
 	connect() {
+		console.log("PartnerTagsController.connect");
+
 		const getSelectValues = (select) => {
 			return [...(select && select.options)].reduce((accumulator, option) => {
 				if (option.selected) {
--- a/app/models/partner.rb
+++ b/app/models/partner.rb
@@ -112,7 +112,7 @@ class Partner < ApplicationRecord
 
   validate :three_or_less_category_tags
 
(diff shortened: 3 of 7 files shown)


kimadactyl:
oops - i think i have muddled up my commits a bit. i think the tests will pass when the other branch is merged though!

Gemfile
```diff
@@ -94,6 +95,7 @@ end
 
 group :development do
   gem 'better_errors'
+  gem 'binding_of_caller'
```
katjam, line 98: We should remove this gem it's a debug utility, not intended for production. https://rubygems.org/gems/binding_of_caller/versions/1.0.0?locale=en

app/javascript/controllers/partner_tags_controller.js
```diff
@@ -4,6 +4,8 @@ export default class extends Controller {
 	static values = { permittedTags: [String] };
 
 	connect() {
+		console.log("PartnerTagsController.connect");
```
katjam, line 7: Remove debug?

app/models/partner.rb
```diff
@@ -112,7 +114,7 @@ class Partner < ApplicationRecord
 
   validate :three_or_less_category_tags
 
-  validate :partnership_admins_must_add_tag, on: %i[create]
+  validate :partnership_admins_must_add_partnership, on: %i[create]
```
katjam, line 117: This is from the other PR I think. Maybe we close this and open a new one with just the intended changes... so we don't have a tangle later.

katjam (changes requested):
Re-open PR with just the strip whitespace changes.

I am assuming the default behaviour of the auto_strip_attributes gem is to trim whitespace top and tail, but maybe we should also add that as explicit config so we can't accidentally stripe whitespace from inside strings later.

app/models/partner.rb
kimadactyl: Yeah all this is. I don't mind re-doing it if it's really needed, but won't this all get automatically fixed when main is updated? Keen to understand git a little better (and yes be a bit lazy if I can lol)

PR #2464, opened 2024-05-02
v0.23.1

kimadactyl:


katjam (approved):
Go for it! I will verify on Monday.

PR #2473, opened 2024-05-15
fix: reduce event limit to switch to weekly view a touch

kimadactyl:
I think this is a little high atm and feedback from organisers has been wanting to switch earlier.

At some point we can add in a feature to pick this manually but for now let's just knock it down a touch.


katjam (approved):
Good call. Let's do a release in a couple weeks.

honor-gfsc:
I still think 150 is really high, and without manually counting I don't know how to verify this.

kimadactyl:
what would you prefer?

honor-gfsc:
like 20!! a page worth!

PR #2476, opened 2024-05-16
Show weekly paginator if there are no events this week but are some in the future

### LINKED ISSUE #2475: [Bug]: Mixer Mcr events are not showing up on their partner page
## Description

I noticed this new partner I'm interested in: https://trans-dimension-manchester.placecal.org/partners/mxer-mcr

I can see they have events importing, and they are visible on the main events listing.

They are not visible on the partner page
Screenshot 2024-05-15 at 20 01 56
Screenshot 2024-05-15 at 20 02 07
Screenshot 2024-05-15 at 20 02 48

## Steps to reproduce

1. See events in backend
2. View partner page
3. See events are not there

Partner events are listed on partner page

kimadactyl:
So what was happening was:

* Event count was more than 20, so going into weekly view
* The weekly view had no listings this week
* If there was no events for the current view, the paginator wasn't showing up

This fix does two things:

* Correctly shows paginator even if no events are visible
* Ups the threshold to move to weekly view a little so it won't show for this partner any more anyway but will ensure it's always visible if paginator is set to true.

Ideally we would implement the "show next event" feature we have on the index page as well, but I can't work out how that works and this fixes the bug itself.

Fixes #2475

Screenshot 2024-05-16 at 15 52 58
Screenshot 2024-05-16 at 15 53 08


katjam (approved):
This seems like a sensible step. As we discovered with testing Trans Dimension, the browsing of events experience could do with some structured trials & design thinking

honor-gfsc:
When trying to verify this, I found a partner which had more than 20 events which was not displaying as the weekly view https://trans-dimension.placecal-staging.org/partners/dalston-superstore 

I'm struggling to find a partner where this use case exists in order to verify this fix

kimadactyl:
guess this would be verified here:

https://trans-dimension-manchester.placecal-staging.org/partners/mxer-mcr

but need the cert... sigh

PR #2490, opened 2024-05-27
Delete .rdoc_options

kimadactyl:
This is empty - suggest we just delete?


katjam (approved):
Sure. If we need to configure, we can add it back.

PR #2492, opened 2024-05-27
Fix wrong internet providers showing up

### LINKED ISSUE #2365: [Bug]: partnership sites showing partners not affiliated in the wifi / computer access part of instance homepage
## Description

If you visit  https://consciouscollectivemcr.placecal.org/ you can see the partners on the homepage are not tagged with the relevant partnership.

kimadactyl:
* Adds byebug gem back into test gems so you can console out of a test (as suggested by the [official docs](https://guides.rubyonrails.org/v6.1/debugging_rails_applications.html#debugging-with-the-byebug-gem)).
* Removes faulty `:for_site_with_tag` method, just chain `:for_site` and `:with_tags` instead
* Update tests

Note I am not 100% confident the tests are doing what they should be doing. They're hard to interpret and read and are deffo due a refactor - factories are adding in way too much context. I've tested this manually a bunch for now though.

Fixes #2365


app/controllers/sites_controller.rb
```diff
@@ -25,12 +25,12 @@ def robots
   def set_places_to_get_computer_access
     tag = Tag.find_by(slug: 'computers')
 
-    @places_to_get_computer_access = Partner.for_site_with_tag(current_site, tag)
+    @places_to_get_computer_access = Partner.for_site(current_site).with_tags(tag)
```
katjam, line 28: Do we need to add the visible flag back in to this query?

app/models/partner.rb
```diff
@@ -176,38 +177,6 @@ class Partner < ApplicationRecord
...
-
-    # skip everything if site has no neighbourhoods
-    return none if site_neighbourhood_ids.empty?
-
-    # TODO; move this scope part that is very similar to the `for_site` method above
-    #   into a shared scope (with possible tests)
-    query
-      .left_joins(:address, :service_areas)
-      .where(
-        'NOT hidden AND (service_areas.neighbourhood_id in (:neighbourhood_ids) OR addresses.neighbourhood_id in (:neighbourhood_ids))',
```
katjam, line 204: This query was additionally checking if the partner should be visible.

katjam (commented):
Just double checking if we need to add that visibility check back in. Otherwise looks ok to try.

app/controllers/sites_controller.rb
kimadactyl: Good thinking - looks like the `for_site` method does check visibility too tho. Which does raise the question if it _should_ - but poss a tech debt question for antoher day?

katjam (approved):
One more teeny thought is that actually we do maybe want to show partners by just area in those sections - because internet & wifi access are kind of agnostic of interest group. But maybe also this is a safeguarding issue?

kimadactyl:
I think the logic here breaks down a bit for interest-based groups over wider areas, and I don't think we should have partners listed that aren't visible on the main partners page, that just sounds confusing. 

Might be more interesting to think about making these blocks modular and togglable? Let's keep an eye out for if anything that comes up that people are looking to highlight?

honor-gfsc:
I think this should be verified by someone with more technical knowledge

kimadactyl:
working on:

- hulme
- moss side
- gmsc (no listings)
- norwich (no listings)
- manchester
- torbay

PR #2494, opened 2024-05-28
Improve site metadata

### LINKED ISSUE #2491: Improve our open graph previews
## User story

Currently, our OG data is a bit all over the place, the defaults are being used in way more places than they should be, site images are not being used over the default one, for example. I think this just looks a bit scuffed basically and with a bit of polish it'll look a lot more professional.

While https://github.com/geeksforsocialchange/PlaceCal/issues/2077 is a bit out of reach for now it would be nice to somewhat polish this so at least the OG text matches the page title and the like.

## Acceptance criteria

- [ ] Overall goal: **all citizen-facing PlaceCal pages have sensible OG/metadata settings**
- [ ] Set site tagline as site home, event and partner index description
- [ ] Set partner summary as partner description
- [ ] Set event description as event description
- [ ] og:images are correct dimensions where possible (1200x600)
- [ ] Set hero image as og:image, if set, on site home index
- [ ] Set hero image as og:image, if set, on site partner index
- [ ] Set hero image as og:image, if set, on site event index
- [ ] Set partner logo as og:image, if set, on site partner show page 
- [ ] Set partner logo as og:image, if set, on site event pages
- [ ] Remove refs to "in your area" and other text (holdover from when we were soley a location-based service)
- [ ] Canonical / meta URLs are the full instance URL to the page
- [ ] Add rake task to regenerate og:images

Note images will need to be regenerated after update, docs here: https://github.com/carrierwaveuploader/carrierwave?tab=readme-ov-file#recreating-versions

## Implementation plan

kimadactyl:
* Fixed opengraph data on main site pages
* Removed some boilerplate copy or moved into locale
* Improved test coverage
* Removed default string for taglines on database end
* Refactored and moved around some things to get better MVC separation of concerns

Fixes #2491

**Note: will need to manually regenerate og:images after merge for sites and partners.** [Instructions are in the carrierwave docs](https://github.com/carrierwaveuploader/carrierwave?tab=readme-ov-file#recreating-versions).

# Previews

NB: the slay the spire screenshot is the site image.

Previewer is showing og:image, og:title, og:description and og:site_name, in that order.

Previews generated by [OGraph Previewer Chrome plugin](https://chromewebstore.google.com/detail/ograph-previewer/ggcfeakcnodgcmmllfdbmngekljbhiim), couldn't find a good Firefox one unfortch.

## Site index

Screenshot 2024-06-10 at 11 31 23

## Event index

Screenshot 2024-06-10 at 11 32 07

## Event show

NB: Trinity House is the organising partner's name. 

Electing not to use partner's image as most are a kind of horrible resolution or quality.

Screenshot 2024-06-10 at 11 39 44

## Partner index

Screenshot 2024-06-10 at 11 32 42

## Partner show

NB: Screenshot is the partner's image, if set. These tend to be low quality / bad so not using more widely for now.

Screenshot 2024-06-10 at 11 33 21


app/components/hero_section_component.rb
```diff
@@ -3,7 +3,8 @@
 class HeroSectionComponent < ViewComponent::Base
   def initialize(image_path:, image_credit:, title:, alttext:)
     super
-    @title = title.presence || 'PlaceCal is a community events calendar where you can find everything near you, all in one place.'
+    # TODO: remove hard coded string from component
+    @title = title.presence || "PlaceCal is a community events calendar where you can find out everything that's happening, all in one place."
```
kimadactyl, line 7: @katjam - when i try and replace this with a locale string, it gives me an error that that function isn't availabe here and some instructions if that's what I really want to do. is this to prevent an antipattern do you think, like should I be approching this in a different way, or should i just get on with it and include the translation library here so i can use the string? does that make sense?

katjam: I think you should include the translation library.

Headers are a weird case I guess because patternwise, you are right, the translations should be evaluated in the html templates.

kimadactyl: right, cool. but then it would be very anti-DRY to have the translation string fed through every time its called by the template. glad you agree its an edge case i couldnt quite make sense of the error message

Procfile.dev
```diff
@@ -1,3 +1,4 @@
 web: bin/rails server -p 3000 -b 0.0.0.0
 js: yarn build --watch
 job_worker: bin/rails jobs:work
+yard: yard server --reload
```
katjam, line 4: Checking this is intentionally included. Seems unrelated to this PR.

README.md
```diff
@@ -79,14 +79,19 @@ You can run the tests & rubocop with:
 make test
 ```
 
-## Documentation for Developers
+## Documentation
```
katjam, line 82: Checking this change is intentionally included. Seems unrelated to this PR.

app/models/event.rb
```diff
@@ -181,6 +181,11 @@ def blame
     "Something wrong with this listing? Contact #{name} <#{email}> with reference {url}"
   end
 
+  def og_title
+    str = "#{summary}, #{date}, #{time}"
```
katjam, line 185: What happens if an event doesn't have a time? Or do they always?

app/models/site.rb
```diff
@@ -125,6 +125,14 @@ def stylesheet_link
     end
   end
 
+  def og_image
+    hero_image&.opengraph&.url ? hero_image.opengraph.url : false
```
katjam, line 129: Would it be better to return a default url than `false`. In the case of false, what happens when the tag is rendered? 

Same question for `og_description`.

app/uploaders/hero_image_uploader.rb
```diff
@@ -2,14 +2,18 @@
 
 # app/uploaders/hero_image_uploader.rb
 class HeroImageUploader < DefaultUploader
-  process resize_to_fit: [2260, 700]
+  process resize_to_fill: [2260, 700]
```
katjam, line 5: I'm assuming this is in response to images not being full width, but is it safe to crop here if we don't supply the images?

app/views/layouts/application.html.erb
```diff
@@ -2,33 +2,14 @@
 <html>
   <head>
     <%= csrf_meta_tags %>
-
     <%= stylesheet_link_tag 'application', media: 'all', 'data-turbo-track': 'reload' %>
     <% if @site&.stylesheet_link %>
       <%= stylesheet_link_tag @site.stylesheet_link, media: 'all', 'data-turbo-track': 'reload' %>
     <% end %>
     <%= stylesheet_link_tag 'print', media: 'print', 'data-turbo-track': 'reload' %>
-
```
katjam, line 11: Nice refactor!

test/integration/partners_integration_test.rb
```diff
@@ -97,9 +96,9 @@ class PartnersIntegrationTest < ActionDispatch::IntegrationTest
     get from_site_slug(@tagged_site, partners_path)
     assert_response :success
 
-    assert_select 'title', count: 1, text: "Partners in your area | #{@tagged_site.name}"
-    assert_select 'div.hero h4', text: "Neighbourhood's Community Calendar"
-    assert_select 'div.hero h1', text: 'Partners in your area'
+    assert_select 'title', count: 1, text: "Partners | #{@tagged_site.name}"
+    assert_select 'div.hero h4', text: @neighbourhood_site.tagline
```
katjam, line 100: Should this be `@tagged_site.tagline`?

test/integration/partners_integration_test.rb
```diff
@@ -84,9 +83,9 @@ class PartnersIntegrationTest < ActionDispatch::IntegrationTest
   test 'region site page shows descendent partners' do
     get from_site_slug(@region_site, partners_path)
     assert_response :success
-    assert_select 'title', count: 1, text: "Partners in your area | #{@region_site.name}"
-    assert_select 'div.hero h4', text: "Neighbourhood's Community Calendar"
-    assert_select 'div.hero h1', text: 'Partners in your area'
+    assert_select 'title', count: 1, text: "Partners | #{@region_site.name}"
+    assert_select 'div.hero h4', text: @neighbourhood_site.tagline
```
katjam, line 87: Should this be `@region_site.tagline`?

db/schema.rb
```diff
@@ -277,7 +277,7 @@
     t.string "hero_image"
     t.string "hero_image_credit"
     t.string "footer_logo"
-    t.string "tagline", default: "The Community Calendar"
```
katjam, line 280: Confirming - we are removing this because we now have a default description that gets used as the tagline if the tagline does not exist. Is that right?

test/integration/collections_integration_test.rb
```diff
@@ -13,7 +13,6 @@ class CollectionsIntegrationTest < ActionDispatch::IntegrationTest
     assert_response :success
 
     assert_select 'title', count: 1, text: "#{@collection.name} | #{@site.name}"
-    assert_select 'div.hero h4', text: 'The Community Calendar'
```
katjam, line 16: Should we check tagline or default description here or is the h4 gone now?

katjam (approved):
There are potentially a couple typos in partners integration tests that should be fixed unless they all use neighbourhood site.

Rest of my comments are flags in case unintentional - but this is OK to merge once you are happy.

Procfile.dev
kimadactyl: Yeah it is - I'm just trying to improve the dev tooling as I go, tbh this was meant to make me want to improve the inline documentation too but ran out of energy for that

app/uploaders/hero_image_uploader.rb
kimadactyl: I mean we have to as this is the size of the area it takes on the page anyway - so the crop is being done in browser either way

test/integration/collections_integration_test.rb
kimadactyl: So this is testing the raw https://placecal.org/partner/partnername page with no subdomain. Partners should never be viewed in this context really, at least with our current logic -- perhaps we could make a kind of global index page later but yeah. This is what I mean from tests needing a bit of a top down reorganise.

app/models/site.rb
kimadactyl: This is set globally in the `_meta.html.erb` partial and not per-model :)

app/models/event.rb
kimadactyl: They always do, we do have an open ticket for proper all day event support tho somewhere.

Procfile.dev
kimadactyl: I think my rationale here is either we are using this gem/process or we're not - having it as kind of a secret feature noone ever remembers to use is a bit weird.

test/integration/partners_integration_test.rb
kimadactyl: good catch

honor-gfsc:
Screenshot 2024-06-24 at 16 39 42
Screenshot 2024-06-24 at 16 40 42
verified

PR #2513, opened 2024-06-20
Update Ruby to 3.1.6

kimadactyl:
I think upgrading and downgrading ruby might have left staging in a weird state. A bit out my depth but I think it may have uninstalled a version recently deprecated in the buildpack.

This updates to 3.1.6 which is the one reccomended in the Heroku docs.

https://devcenter.heroku.com/articles/ruby-support#ruby-versions


katjam (approved):
Let's go up!

PR #2526, opened 2024-07-01
chore: Update Ruby to 3.3.3

### LINKED ISSUE #2483: [Bug]: Upgrade heroku in buildpack to support Ruby 3.3.1
## Description
Renovate upgraded Ruby from 3.1.x to 3.3.x - Our buildpack does not support that so deploy is currently broken.

https://github.com/geeksforsocialchange/PlaceCal/actions/runs/9187309397/job/25268127334

If it's not a quick fix, we will roll back that PR so that we can get our partner event listing bugfix released

Docs:
https://dokku.com/docs/deployment/builders/herokuish-buildpacks/

kimadactyl:
OK, I think this will finally work now the buildpack issue is resolved.

Fixes #2483


katjam (approved):
Nice! Let's try it.

PR #2527, opened 2024-07-01
chore: Update Node.js to v20

kimadactyl:


katjam (approved):
As long as the build works, this should be fine, I'm surprised more deps weren't updated when it changed, but let's merge.

kimadactyl:
Oh I just updated that one package as it was the one moaning about the node version. I think the rest will update afterwards!

PR #2548, opened 2024-08-02
fix: Handbook link

### LINKED ISSUE #2536: [Bug]: Handbook link in calendar interface is broken
Screenshot 2024-07-12 at 15 31 55
The link in the screenshot 404s

kimadactyl:
Is it weird to move things into locale one at a time when they become an issue? Doing the whole file would take ages but as is this is a quick fix.

Fixes #2536


config/locales/en.yml
```diff
@@ -46,6 +46,9 @@ en:
           attributes:
             tag_ids:
               blank: "Select at least one Tag"
+  admin:
+    calendars:
+      handbook_hint: "The source URL for your calendar feed. See the <a href='https://handbook.placecal.org/reference/supported-calendar-sources'>PlaceCal Handbook</a> if you need help."
```
katjam, line 51: Wondering if it's better to mark html safe here by using `_html` in key. If it's in the render template, then may be forgotten if html is removed from the locale string. No strong feelings but think marginally "safer" to acknowledge the html in the place it is written out so it's obvious what it is, rather than blindly whitelisting whatever ends up here in the template.

Probably more important to follow whatever convention is set in code previously too though.

kimadactyl: thats a nice convention, i like it

kimadactyl:
can you approve pls @katjam ? :)

app/views/admin/calendars/_form.html.erb
```diff
@@ -21,7 +21,7 @@
                           hint: 'Which group organises these events?' %>
 
         <%= f.input :name, hint: 'A simple description to help rememember what this calendar is' %>
-        <%= f.input :source, label: 'URL', placeholder: 'https://your-domain.com/events.ics', hint: "The source URL for your calendar feed. See the #{link_to 'PlaceCal Handbook', 'https://handbook.placecal.org/admin-handbook'} if you need help.".html_safe %>
+        <%= f.input :source, label: 'URL', placeholder: 'https://your-domain.com/events.ics', hint: t('admin.calendars.handbook_hint_html').html_safe %>
```
katjam, line 24: I don't think you need the `.html_safe` here anymore but I may be misunderstanding the magic powers of the locale key slug `_html`

katjam (approved):
Approved with caveat comment.

app/views/admin/calendars/_form.html.erb
kimadactyl: oh i see!! i didn't know that! is this a normal ish thing for using locale files? thanks for the tip :)

katjam: I've not encountered before. Seems like a special rails thing. I _think_ it makes sense though to declare the html escape in the place where you add it.

PR #2568, opened 2024-08-19
feat: Add Appsignal

### LINKED ISSUE #2547: Add Appsignal & possibly remove Rollbar
## Use case

I've been having a look at the[ codebase for the government epetitions website](https://github.com/alphagov/e-petitions/tree/master) to learn some tricks and see they are using this service [Appsignal](https://www.appsignal.com/) which looks excellent. I've never loved rollbar, I don't think it helps with bug triage at all really, and this looks like a nice drop in alternative.

It seems like pure upside:

- Most notably, it has a free open source package so we'd no longer be fighting rate limits.
- It also has a GUI for logging, which will really help as currently you need to go into dokku for this. Also it means when we add better CRUD creation logs, we have our data log done.
- We don't have any monitoring for expensive processes atm - this looks like it has it in spades
- It's specifically targetting Rails and really understands the underlying system - rollbar is very generic

To me this just looks like a pure upgrade.

## Acceptance criteria

- [x] Get a free open source account
- [x] Add Appsignal
- [x] Remove rollbar when we've agreed it's surplus to requirements

## Implementation plan

- [x] Install gem and get token
- [x] Add env var to staging
- [x] Add env var to prod
- [x] Test on staging
- [x] Remove rollbar
- [x] Deploy to prod
- [x] Test on prod

kimadactyl:
Partially fixes #2547 

To test on staging and if we like it will open a PR to remove Rollbar.


katjam (approved):
Great. Let's try it. Is there any config that needs updating for it to work?

kimadactyl:
no i think this is it, its just an environment variable that ive now set on prod and staging.

i think it is possible to configure it after that, though, but a task for another day.

katjam:
> no i think this is it, its just an environment variable that ive now set on prod and staging.
> 
> i think it is possible to configure it after that, though, but a task for another day.

In that case, we just need to add the stub to https://github.com/geeksforsocialchange/PlaceCal/blob/main/.env.example so it doesn't get forgotten in future deployments / setups

Issue #2542, opened 2024-07-29
Block AI bots


kimadactyl:
## Use case

Promoted by this: https://about.readthedocs.com/blog/2024/07/ai-crawlers-abuse/

> Given that our Community site is only for hosting open source projects, AWS and Cloudflare do give us sponsored plans, but we only have a limited number of credits each year. The additional bandwidth costs AI crawlers are currently causing will likely mean we will run out of AWS credits early.
> 
> By blocking these crawlers, bandwidth for our downloaded files has decreased by 75% (~800GB/day to ~200GB/day). If all this traffic hit our origin servers, it would cost around $50/day, or $1,500/month, along with the increased load on our servers.
>
> Normal traffic gets cached by our CDN, and doesn't cost us anything for bandwidth. But because many of these files are not downloaded often (and they're large), the cache is usually expired and the requests hit our origin servers directly, causing substantial bandwidth charges. Zipped documentation was designed for offline consumption by users, not for crawlers.

We should look into filtering out AI requests

## Acceptance criteria

- [ ] AI requests are filtered out before hitting PlaceCal

## Implementation notes & questions

I think this can be done with `.htaccess` somehow - need to research more.

## Implementation plan

_To be written by the developer_

PR #2569, opened 2024-08-19
feat: Add robots.txt to ask AI bots not to crawl the site

### LINKED ISSUE #2566: [Bug]: Invalid date params getting through to Date constructor
## Description
Reported on rollbar. 7052 times between 2024-04-06 and 2024-08-19

https://app.rollbar.com/a/gfsc/fix/item/placecal-production/415

image

### LINKED ISSUE #2567: [Bug]: current_site sometimes undefined
## Description

Reported by rollbar. 18 times between 2024-04-13 and 2024-08-19 on productions server. 

https://app.rollbar.com/a/gfsc/fix/item/placecal-production/417

Doesn't seem like a big deal but we ought to catch and set default and find out why it is sometimes not there.

image

kimadactyl:
List from: https://github.com/ai-robots-txt/ai.robots.txt

Fixes #2567, unless we think it should stay open until the underlying issue is fixed.
Fixes #2566, unless we think it should stay open until the underlying issue is fixed.
Part of #2542


katjam (approved):
Not convinced the bots will respect, but let's see. Is it worth noting the source of the list somewhere so we can try to keep it up to date.

Happy for this to close those other 2 issues. They'll get reopened if we can't verify.

kimadactyl:
there's an RSS feed for it, i'll add to my reader!

PR #2577, opened 2024-08-26
fix: update cron command

kimadactyl:
Oops, forgot cron is set in the dokku config


katjam (approved):
Great

kimadactyl:
hmmm while we're here should we rename this or is it ok? it looks a bit half finished too ngl

kimadactyl:
eh gonna just merge it, distraction

PR #2582, opened 2024-08-27
feat: Add Appsignal deploy marker

kimadactyl:
This just lets Appsignal know that there's a new deployment so we can see if things are fixed or not as a result of updates


katjam (approved):
Great idea.

config/appsignal.yml
```diff
@@ -0,0 +1,6 @@
+# https://docs.appsignal.com/application/markers/deploy-markers.html
+
+production:
+  revision: <%= ENV['GIT_REV'] %>
```
katjam, line 4: I assume this is a Git env var set somewhere to track the latest tag / release?

kimadactyl:
Oh no. An unrelated Capybara test has stopped working. I'm having a poke and it seems to have nothing to do with this. really weird.

katjam:
> Oh no. An unrelated Capybara test has stopped working. I'm having a poke and it seems to have nothing to do with this. really weird.

Ugh.

test/system/application_system_test_case.rb
```diff
@@ -9,4 +9,11 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   Capybara.server = :puma, { Silent: true }
   Selenium::WebDriver.logger.ignore(:browser_options)
+
+  # TODO: Remove this hack
```
kimadactyl, line 13: @katjam - not the prettiest but i think if we don't get this passing we can't deploy. i've made a tech debt ticket.

katjam: Seems like a reasonable hack though.

PR #2587, opened 2024-08-28
feat: improve performance in SitesController#index

### LINKED ISSUE #2584: Slow performance in SitesController#index
**Duration**

```
  81 ms
```

View on AppSignal: https://appsignal.com/geeks-for-social-change/sites/66c5191d2bf9601632322b31/performance/incidents/6?timestamp=2024-08-28T10:41:24Z

kimadactyl:
The others are beyond me for now but this looks like a hopeful improvement. Prob a bit silly to work on this of all things but having performance benchmarking is a novelty and it only took 5 mins.

Fixes #2584


katjam (approved):
If you are 100% happy it returns the same result then I'm also happy!

kimadactyl:
It's 2 queries instead of 4 but I realised afterwards I don't _actually_ have any proof that's faster -- something to think about!

PR #2597, opened 2024-09-03
feat: Remove rollbar

### LINKED ISSUE #2547: Add Appsignal & possibly remove Rollbar

kimadactyl:
Final step once we're happy with Appsignal - I think its just a straight upgrade!

Fixes #2547


app/jobs/calendar_importer_job.rb
```diff
@@ -60,7 +60,6 @@ def report_error(e, message)
     calendar.reload
     calendar.flag_error_import_job! full_message
     # puts full_message, backtrace if Rails.env.development?
```
katjam, line 62: Also remove this line?

kimadactyl: sure thing

katjam (approved):

kimadactyl:
Confirmed removed from staging

PR #2602, opened 2024-09-05
feat: Add GET endpoint to GraphQL API

kimadactyl:
@katjam - if you want a GET endpoint this is all we need to do. feel free to merge it yourself if you want to experiment on staging. the bruno collections just work if i switch to GET! feels too easy...

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -95,6 +95,7 @@ Rails.application.routes.draw do
 
   get '/robots.txt' => 'pages#robots'
 
+  get '/api/v1/graphql', to: 'graphql#execute'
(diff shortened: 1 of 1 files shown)


katjam (approved):
Sure I guess let's do this. Better than the hack I'm setting up.

katjam:
I think we want to back this out before any release.

kimadactyl:
Oh really? I'm happy to merge I don't see the downside - it actually makes it way easier to debug as you can use browser to view it. Id rather leave in then take out again if its a problem?

Confirmed `ping` is working on both GET and POST on staging.

PR #2638, opened 2024-11-02
Migrate the AddressComponent to using ViewComponent

### LINKED ISSUE #2270: remove mountain_view gem / code

lenikadali:
Partially resolves #2270 

## Description

This PR migrates the `AddressComponent` to use `ViewComponent` instead of Mountain View. 
Updated the calls to `AddressComponent` as well to partners' and events' index pages.

### Motivation and Context
From [the issue](https://github.com/geeksforsocialchange/PlaceCal/issues/2270):

> `mountain_view` gem has not scaled at all well and has a bunch of weird issues with integrating with JavaScript.

> We therefore need to migrate to a newer solution, `view_component`

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to change)

### How Has This Been Tested?

Currently the attempt at a test in `test/components/address_component_test.rb` is failing.
Need to see how to display the event and partner pages.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -131,4 +131,5 @@ Lint/SelfAssignment:
   Enabled: false
 Lint/RedundantCopDisableDirective:
   Enabled: false
-
+Lint/MissingSuper:
+  Enabled: false
--- a/app/components/address_component.rb
+++ b/app/components/address_component.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class AddressComponent < ViewComponent::Base
+  def initialize(address, raw_location: nil)
+    @address = address
+    @raw_location = raw_location
+  end
+
+  def formatted_address
+    # rubocop:disable Rails/OutputSafety
+    if address.present?
+
+      address_lines = address.all_address_lines.map(&:strip)
+      return address_lines.join(', <br>').html_safe
+    end
+
+    uri = URI.parse(raw_location)
+    "<a href='#{uri}'>#{uri.hostname}</a>".html_safe
+
+  rescue URI::InvalidURIError
+    raw_location
+  end
+  # rubocop:enable Rails/OutputSafety
+end
--- a/app/views/partners/show.html.erb
+++ b/app/views/partners/show.html.erb
@@ -97,9 +97,10 @@
         <div class="gi gi__1-2">
           <h3 class="udl udl--fw allcaps h4">Address</h3>
           <div class="small">
-            <%= render_component "address",
-              address: place.address
-            %>
+        <%= render(
+          AddressComponent.new(address: place.address)
+        )
+        %>
           </div>
           <h3 class="udl udl--fw allcaps h4">Contact</h3>
           <div class="small">
--- a/test/components/address_component_test.rb
+++ b/test/components/address_component_test.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'view_component/test_case'
+
+class AddressComponentTest < ViewComponent::TestCase
+  setup do
+    VCR.use_cassette('import_test_calendar') do
+      @test_event = create(:event)
+      @address = @test_event.address.street_address
+      @raw_location = @test_event.raw_location_from_source
+    end
+  end
+
+  def test_component_renders_address
+    render_inline AddressComponent.new(address: @address, raw_location: @raw_location)
+    assert_text 'Unformatted Address, Ungeolocated Lane, Manchester'
+  end
+end
--- a/app/components/address_component.html.erb
+++ b/app/components/address_component.html.erb
@@ -1,0 +1,3 @@
+<p class="place_info__address" property="address" typeof="PostalAddress">
+  <%= formatted_address %>
+</p>
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -29,9 +29,9 @@
     <div class="gi gi__1-3">
       <h3 class="h4 udl">Event address</h3>
       <div class="small">
-        <%= render_component "address",
-            address: @event.address,
-	    raw_location: @event.raw_location_from_source
+        <%= render(
+          AddressComponent.new(address: @event.address, raw_location: @event.raw_location_from_source)
+        )
         %>
       </div>
     </div>


lenikadali:
Not sure what the appropriate name should be for the PR title between `fix`, `docs` and `chore`. 

Happy to update the title once that's clear 🙏🏿

kimadactyl:
Looking good!

I think let's not touch VCR for the unit tests, anything with VCR feels more like an integration test.

component tests should just test the logic inside the component - should make it simpler to read too.

there's a small thing on these too that i'm unsure about - do we like having all the new components suffixed `_component`? it makes it easier to see whats been done, and makes it ultra clear what's being called, but im not sure how annoying itll be later on lol.

.rubocop.yml
```diff
@@ -131,4 +131,5 @@ Lint/SelfAssignment:
   Enabled: false
 Lint/RedundantCopDisableDirective:
   Enabled: false
-
+Lint/MissingSuper:
+  Enabled: false
```
kimadactyl, line 135: can you explain what this is? i never really got `super` but disabling a linter makes me nervous :)

app/components/address_component.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class AddressComponent < ViewComponent::Base
+  def initialize(address, raw_location: nil)
+    @address = address
+    @raw_location = raw_location
+  end
+
+  def formatted_address
+    # rubocop:disable Rails/OutputSafety
```
kimadactyl, line 10: this also seems a good thing to remove

test/components/address_component_test.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'view_component/test_case'
+
+class AddressComponentTest < ViewComponent::TestCase
+  setup do
+    VCR.use_cassette('import_test_calendar') do
```
kimadactyl, line 8: I think let's not touch VCR for the unit tests, anything with VCR feels more like an integration test.

.rubocop.yml
lenikadali: Yeah. I think will remove this. I looked around at other codebases using ViewComponent and probably the solution is to call `super` in the method instead of disabling the cop.

app/components/address_component.rb
lenikadali: Will give it a go and see 🙌🏿

.rubocop.yml
kimadactyl: ohhh i see so otherwise it would monkeypatch it? makes sense :)

lenikadali: Resolved in [9b10173](https://github.com/geeksforsocialchange/PlaceCal/pull/2638/commits/9b1017332dcc523fbb76326bca1dec6ad9ba321f)

app/components/address_component.rb
lenikadali: Resolved in [9b10173](https://github.com/geeksforsocialchange/PlaceCal/pull/2638/commits/9b1017332dcc523fbb76326bca1dec6ad9ba321f)

test/components/address_component_test.rb
lenikadali: Resolved in [9b10173](https://github.com/geeksforsocialchange/PlaceCal/pull/2638/commits/9b1017332dcc523fbb76326bca1dec6ad9ba321f)

kimadactyl:
@lenikadali i pushed this along a bit. rough process was to load up a page that has an address on it and note the error then just work throught the errors. i think the main confusion is where files should live. right now the presenter is in the root component directory and the template is in a folder - i think this is confusing and it should all live in the same place.

i've deleted the mountain_view code but now the other references to the address component need changing. does this make sense?

app/components/address_component.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(address:, raw_location: nil)
+    super
+    @address = address
+    @raw_location = raw_location
+  end
+
+  def formatted_address
+    if @address.present?
+      address_lines = @address.all_address_lines.map(&:strip)
+      return address_lines.join(', <br>').html_safe # rubocop:disable Rails/OutputSafety
```
kimadactyl, line 13: i couldnt work out what it wanted me to do here but feels fine to leave in for now if you cant figure it out either

app/components/address_component.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AddressComponent < ViewComponent::Base
+  def initialize(address:, raw_location: nil)
```
kimadactyl, line 4: i think we should possibly change the function siganature so it just takes an address? im not sure what this raw_location thing is doing but its on the Address record anyway

app/components/address_component.rb
lenikadali: strange. I thought I had resolved in [9b10173](https://github.com/geeksforsocialchange/PlaceCal/pull/2638/commits/9b1017332dcc523fbb76326bca1dec6ad9ba321f) Looks like I forgot to change this one as well; the solution was to use `sanitize` here instead of `html_safe`. Will fix.

kimadactyl: ugh it might be prettier/rubocop interplay

kimadactyl:
Have tried the inline template style - what do you think? I think this is a lot easier to work with and reduces cognitive load but comes with the cost that unpacking it again means undoing this? But would we do that for something this simple?

app/components/address_component.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class AddressComponent < ViewComponent::Base
+  erb_template <<~ERB
```
kimadactyl, line 4: experiment - is this ok or does it cause more problems than its worth?

lenikadali: I would say more problems than it is worth; having the template be separate seems cleaner. Also in case the component gets bigger, it wouldn't be as easy to read with template code in it.

app/components/address_component.rb
lenikadali: I changed the call to this https://github.com/geeksforsocialchange/PlaceCal/blob/2270-migrate-address-component-to-view-component/app/views/events/show.html.erb#L32

and the original (on the `main` branch) is this https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/views/events/show.html.erb#L32-L36

The commit adding `raw_location` to the signature is [b6959cf](https://github.com/geeksforsocialchange/PlaceCal/commit/b6959cf2950647c5f40abca9a703f501316f7429) and from looking around `raw_location_source` is the renamed `location` field on the `Event` model https://github.com/geeksforsocialchange/PlaceCal/blob/main/db/migrate/20181121122228_rename_events_location_column.rb#L5

Going by the history of that migration, [5bab05d](https://github.com/geeksforsocialchange/PlaceCal/commit/5bab05df938cf7cee9c86511803b97392c5bab9d#diff-3bd8a9f39831c937db535d4d188966a521cab79270c5b040df0cb450e342c59aR44) and searching in the code, it seems `app/models/events/base.rb` has become [`app/jobs/calendar_importer/events/base.rb`](https://github.com/geeksforsocialchange/PlaceCal/blob/main/app/jobs/calendar_importer/events/base.rb#L88) which is the only place `raw_location_from_source` is used.

Long story short :sweat_smile: I would keep the function signature the same, unless you feel up to further digging and manual testing. This would go back to the scope of the ticket: we can make a `TODO` to harmonize the usage of this field once the migration is done or attempt it within this PR. Let me know what you think :relaxed:

lenikadali (commented):
> i've deleted the mountain_view code but now the other references to the address component need changing. does this make sense?

Yup :sunglasses: I think that is the natural next step here. Most likely my mistake was to attempt to migrate while the `mountain_view` code was still in place so that was a good thing to do.

> Edit: I've done a bit of searching and maybe its not possible to have the component .rb file in the sidecar directory, wahhh https://viewcomponent.org/guide/templates.html#subdirectory

Not sure about this. It looked like it could cause errors as [reported here](https://github.com/ViewComponent/view_component/issues/79#issuecomment-543945867)
If the test can succeed, then I would be open to that setup. It's your call :wink:

app/components/address_component.rb
kimadactyl: i was thinking about this when i was trying to sleep lol. i think one complaint we have deffo had with rails is theres too many files involved in making a change - like having these in 2 diff places is a bit infuriating! having it all in one file makes it super easy to edit! like i agree its a bit of a faff if it gets bigger but what if it doesnt? maybe gonna sleep on it a bit more lol

app/components/address_component.rb
kimadactyl: ok agreed, and also i think with hindsight this is way too in the weeds to get into if we want to whizz through this directory, more just limit testing here :)

lenikadali: Removed my original comment since it seemed to be duplicated across the review thread but apparently GitHub's review comment UI is misleading in its design. Anyway, what I recall writing was as follows:

The function signature `initialize(address:, raw_location: nil)` was added in commit [b6959cf](https://github.com/geeksforsocialchange/PlaceCal/commit/b6959cf2950647c5f40abca9a703f501316f7429) for displaying events without an address.

This commit [5bab05d](https://github.com/geeksforsocialchange/PlaceCal/commit/5bab05df938cf7cee9c86511803b97392c5bab9d) shows that the `raw_location_from_source` was a renaming of the `location` field on the `Event` model

The `raw_location` field isn't required for showing partners' info since they will most likely have filled in an address as part of onboarding.

app/views/partners/show.html.erb
```diff
@@ -46,9 +46,8 @@
         <% if @partner.has_service_areas? %>
           <p>We operate in <%= partner_service_area_text(@partner) %>.</p>
         <% end %>
-        <%= render_component "address",
-          address: @partner.address
-        %>
+
+        <%= render AddressComponent.new(address: @partner.address) %>
```
kimadactyl, line 50: @lenikadali i found one more address just fyi! otherwise happy to merge if you are?

.rubocop.yml
```diff
@@ -131,4 +131,4 @@ Lint/SelfAssignment:
   Enabled: false
 Lint/RedundantCopDisableDirective:
   Enabled: false
-
+
```
kimadactyl, line 134: cant work out how to get rid of this but i think more faff than its worth to remove from the commit now lol

app/views/partners/show.html.erb
lenikadali: Hahaha beat me to it :smile: 

Yes please merge :raised_hands:

kimadactyl: oh no theres an test fail now are you ok to fix the test?

lenikadali: Not sure why that is. Locally it doesn't fail though. Let me see :crossed_fingers:

kimadactyl (approved):

PR #2668, opened 2025-01-18
fix: add imagemagick to ci script

kimadactyl:
Unclear why or how it was working without this in there but all CI tests failing atm due to this!


katjam (approved):
Probably used to be bundled with node or other package we use

kimadactyl:
yeah weird it just started happening from nowhere tho!

katjam:
yeah weird? Maybe renovate bot updated something?

PR #2677, opened 2025-02-07
Update upload-artifact

kimadactyl:
CI won't run as v3 is deprecated.


lenikadali (approved):
Looks good to me. Hopefully we will not need to go back to version pinning :crossed_fingers: :smile:

PR #2685, opened 2025-02-12
WIP: Change on push to pull_request_target

### LINKED ISSUE #2556: PRs from forks are not running GitHub Actions properly
## Description

2 new contributors have done PRs from forks and GitHub isn't loading properly. The below action just never starts. A Google I did before hinted this is something to do with our key config, but leaving here for further investigation.

Screenshot 2024-08-13 at 21 50 26

## Motivation

We want to have people contribute to the repo without needing to add them into our org.

JessFairbairn:
Resolves #2556

### Motivation and Context

Get the Github actions running for pull requests created by people outside of the PlaceCal group

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update


kimadactyl:
oh wait does this need to be merged into main before it will work? hmm

JessFairbairn:
It doesn't seem to have triggered the test-and-deploy stage at all as far as I can tell?

wheresalice:
You'll still need the `on: push` in order to run the steps that only run on the main and production branches, but I think that adding the `pull_request_target` will indeed fix this once merged

There is a security impact in doing this though, because it means anyone's PRs can read your github secrets and token unless you also set permissions in the workflow

So generally people will use `on: pull_request` which doesn't give access to secrets and tokens, but then you'll need to manually approve the runs (at least for first time contributors) - https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

I think what you want is:

```yaml

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
```

PR #2687, opened 2025-02-12
chore: Delete unused  bus route component

### LINKED ISSUE #2686: bus_route/_bus_route.html.erb

kimadactyl:
This is markup for a physical bus that hasn't existed since about 2019, lol. Looks like the page for it got deleted but the associated components did not.

Resolves #2686


kimadactyl:
yeah good to do both sides of the job! cheers :)

PR #2666, opened 2025-01-09
WIP: chore: Migrate AdminIndexComponent to use ViewComponent

lenikadali:
Migrates the AdminIndexComponent to use ViewComponent instead of MountainView.

# Checklist:

- [x] I have performed a self-review of my own code,
- [x] I have commented my code, particularly in hard-to-understand areas,
- [ ] I have added tests that prove my fix is effective or that my feature works,
- [x] Title include "WIP" if work is in progress.

Resolves #2646 

### Motivation and Context

From [this issue](https://github.com/geeksforsocialchange/PlaceCal/issues/2270):

> mountain_view gem has not scaled at all well and has a bunch of weird issues with integrating with JavaScript.
> We therefore need to migrate to a newer solution, view_component

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### How Has This Been Tested?

Unfortunately this hasn't been tested :see_no_evil: 
The logic in `app/components/admin_index_component/admin_index_component.html.erb` is a bit complex.

PR #2718, opened 2025-04-17
chore: convert admin_edit and contact_details to viewcomponent

kerimhudson:
This PR attempt to tick off some bits from #2270, specifically the `admin_edit` and `contact_details` components. Thought I'd share and get this up before going too deep in the rabbit hole to make sure I'm not making any obvious mistakes as it's my first time working with Ruby.
# Checklist:

-->

### Type of change


kimadactyl:
This looks awesome Kerim, you've smashed it. I'll add a few notes which are mostly our fault but I can't believe you just picked up Rails!

app/components/admin_edit/admin_edit.html.erb
```diff
@@ -0,0 +1,8 @@
+<% content_for @title do %><%= @title %><% end %>
```
kimadactyl, line 1: we eventually decided just to put the components and .rb files in the root components directory - im actually a little suprised this is working as previously the rb file needed to be in the root and the template in its own directory, whcih is why we moved. puzzling!

app/components/contact_details/contact_details.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# app/components/contact_details/contact_details_component.rb
+class ContactDetails < ViewComponent::Base
+  # rubocop:disable Metrics/ParameterLists
+  def initialize(name: nil, phone: nil, url: nil, email: nil, facebook_link: nil, twitter_handle: nil, instagram_handle: nil, is_valid_phone: nil)
```
kimadactyl, line 6: this function signiture was very naive of me back in 2017 when i was still learning rails - i think lets use the opp to refactor this to just take the @partner as the function input?

(may also be as places used to be a sep model)

kimadactyl:
Remaining notes and issues:

- Due to https://github.com/geeksforsocialchange/PlaceCal/issues/2556, we can't run CI on forks - have added you to the org team so if you can redo this after changes that'd be ace
- Our testing leaves a lot to be desired but a very simple presence test would be great, noting that @lenikadali has had [some issues with this](https://github.com/geeksforsocialchange/PlaceCal/pull/2666#pullrequestreview-2733815916) - so i think if this happens lets just leave it, we are no worse off than before
- just a note on our process, please make a ticket and assign yourself before working on things - you prob had to be added to the repo for this, just noting going forwards, i shoulda added you before :). theres a handy button on the master ticket to convert the bullet items into issues

kerimhudson:
Will close this ticket and open a new one with the changes!

Issue #2704, opened 2025-03-13
Flaky tests are getting more unreliable


kimadactyl:
## Description

We have had an occasionally flaky test go from what feels like a 1/10 failure rate to a 4/10 and it's getting annoying. They look like this.

Probably we want to think about replacing this rather than doing too much work to make it work again. #2573 maybe.

```
 Done in 0.25s.
Running 17 tests in a single process (parallelization threshold is 50)

# Running tests with run options --seed 12317:

.......[Screenshot Image]: /home/runner/work/PlaceCal/PlaceCal/tmp/capybara/failures_test_visiting_the_index.png
E.........

Finished tests in 105.388306s, 0.1613 tests/s, 1.3759 assertions/s.


Error:
CreateAdminUsersTest#test_visiting_the_index:
Capybara::ElementNotFound: Unable to find field "New password" that is not disabled
    test/system/create_admin_users_test.rb:55:in `block in <class:CreateAdminUsersTest>'

17 tests, 145 assertions, 0 failures, 1 errors, 0 skips

== Command ["bin/rails test:system"] failed ==
```

## Motivation

Why is it important?

Issue #2573, opened 2024-08-22
Add regression tests for existing frontend functionality


kimadactyl:
## User story

As a developer or project owner, I need to understand what the frontend is meant to do, in order that I can reason about changes to it and prevent things that work now from breaking.

## Acceptance criteria

- [ ] There is a BDD test suite that describes the desirable functionality
- [ ] Events have their behaviour described in tests
- [ ] Partners have their behaviour described in tests
- [ ] Sites have their behaviours described in tests
- [ ] Project README is updated to reflect the testing idiom

## Implementation notes & questions

Proposal is to set up a [Spinach](https://github.com/codegram/spinach) test suite that describes the functions. @pedanticgit's https://github.com/fishpercolator/name.pn/ repo has an example how to do this,

## Implementation plan

_To be written by the developer_

pedantic-git:
LMK if you need any specific Spinach or Capybara tips when this work gets underway.

kimadactyl:
> LMK if you need any specific Spinach or Capybara tips when this work gets underway.

Ahh thankyou! The main thing I could maybe use your eyes on is our Capybara config. This does not look like a happy test suite to me - a pragmatic solution to something that perhaps shouldnt be there.

https://github.com/geeksforsocialchange/PlaceCal/blob/main/test/system_test_helper.rb#L24-L44

I think a lot of this has to do with Select2 still loading from JQuery. Do you have a gut feeling if this a Select2 problem, or Selenium, or other?

pedantic-git:
Oh yeah I think this is a classic Capybara gotcha that catches everyone when they first start. The rule with Capybara is "if you think you have to write your own waiting code, you don't".

Where it usually catches people is when they do something like this:

```ruby
input = find('input.blah') # This blows up if input.blah hasn't appeared yet
# ... eg. input.send_keys 'Hello'
```

Here you don't need to write your own waiting code to wait for the input to appear - you can just add another assertion before it:

```ruby
expect(page).to have_css('input.blah') # This will wait for the default timeout - usually 5 seconds - for the element to appear before it considers the assertion to have failed
input = find('input.blah')
# ...
```

You can increase the timeout on assertions that are expected to take a long time (e.g. if there's a whole page with millions of images loading!), with the `:wait` parameter:

```ruby
expect(page).to have_css('input.blah', wait: 10)
```

[...]

PR #2719, opened 2025-04-20
chore: view component for admin edit & contact details

### LINKED ISSUE #2720: admin_edit/admin_edit_component.rb

### LINKED ISSUE #2721: contact_details/contact_details_component.rb

kerimhudson:
# Checklist:

Resolves #2720 
Resolves #2721 

## Description

Wasn't sure how to create a ticket that tackles two at once from the main issue? Is there a way to do that? Nevertheless, this converts the `admin_edit` and `contact_details` components from MountainView to ViewComponent.

### Type of change

- [ ] Documentation update


app/views/partners/show.html.erb
```diff
@@ -100,11 +100,12 @@
           </div>
           <h3 class="udl udl--fw allcaps h4">Contact</h3>
           <div class="small">
-            <%= render_component "contact_details",
+            <%= render ContactDetails.new(
               email: place.public_email,
```
kerimhudson, line 104: I guess I haven't tackled this one, but I guess what's the desired approach here. Should it just take the partner still, or have the option to override email, phone & url as this is doing?

I guess it's not quite right too that the `is_valid_phone` checks the partner valid phone, when the place phone is being used too.

kimadactyl: Yeah a bit confusing - we used to have sep partner and place models but it got too messy so we merged into one. Now it's messy in a diff way lol. A place is now a partner that represents a physical location for the event. So yes - this should be partner - but this whole logic needs a bit of a review and to make explicit more.

kimadactyl: Given our priority is to remove mountain_view its prob best to not mess with this as much as possible!

kerimhudson: How's this for a fix? I've pushed the changes, basically takes the partner model, but allows some optional params?

app/assets/stylesheets/application.scss
```diff
@@ -49,3 +49,4 @@
 @import "components/partnership_card_component";
 @import "components/pull_quote_component";
 @import "components/steps_component";
+@import "components/contact_details_component";
```
kimadactyl, line 52: alphabetize pls :)

app/views/partners/show.html.erb
kimadactyl: so the easiest way to fix the tests is to drop a `byebug` in the offending place then run the specific test - it should kick you into a console where you can poke around. im sure theres more elegant ways but it works for me!

this is looking great tho!!

kimadactyl: fyi when youre working on views you can also put a console in a view with `<% console %>`, or chuck some garbage in the controller to make it error and itll throw up a console too

kerimhudson: Seems like the tests are system tests now failing on this, I'm not sure this is code I touched, so it's interesting. I see a ticket regarding flaky tests, is this what it's regarding?

Failing test is `CreateAdminUsersTest#test_visiting_the_index:`. Attached the corresponding image.

I don't see where this is getting called from though? I see this string in the devise yml file, but I don't see it called from anywhere in the flash array.

failures_test_visiting_the_index

kimadactyl: and welcome to hopefully the last bit of annoyance right now - at some point our system tests (i.e. ones that do the browser emulation and use js) got flaky. this is one. you prob have a good sense of the current milestone now lol. ill rerun it!

kerimhudson: I did re-run it 3 times, and it kept flaking, so brought it back here, but fingers crossed this time eh!

kimadactyl: tbh i think we should prob comment them out from running on CI. theyre SUCH a pain and are slowing down develoment a ton and dont cover _that_ much functionality. feel free to do it as part of this (change line 55 in the Makefile to just be `test` not `test:system`) and ill document in the ticket to handle this.

it started about 1 in 20 times and now it feels like its 50/50 or worse, big cry

kimadactyl: yep that works good shout!

kimadactyl: issue here for the record: #2704. really tho i think we want to address this with this: #2573

kimadactyl (approved):
This looks great to me! Could optionally add a simple presence test in `tests/components` - but as mentioned the testing strategy does need a rethink and we prob have good coverage of these features already.

kerimhudson:
> This looks great to me! Could optionally add a simple presence test in `tests/components` - but as mentioned the testing strategy does need a rethink and we prob have good coverage of these features already.

I'll look to review the testing as a whole maybe once this big ticket goes through?

PR #2730, opened 2025-04-23
chore: move components from mv to viewcomponent

### LINKED ISSUE #2731: Site homepage components (can be done in one PR)

### LINKED ISSUE #2732: hero/hero_component.rb

### LINKED ISSUE #2733: footer/_footer.html.erb

### LINKED ISSUE #2734: meta/_meta.html.erb

kerimhudson:
Resolves #2731
Resolves #2732
Resolves #2733
Resolves #2734

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update


app/components/meta.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Meta < ViewComponent::Base
+  renders_one :link
```
kimadactyl, line 4: what does this do?

app/components/hero_section_component.html.erb
```diff
@@ -1,4 +1,5 @@
 <div class="hero_section">
+  This is a hero section
```
kimadactyl, line 2: possibly a test string to remove?

app/components/hero_component.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class HeroComponent < ViewComponent::Base
+  def initialize(title, subtitle = '', schema = nil)
+    super
+    @title = title.length > 32 ? title.split.in_groups(2, false).map { |g| g.join(' ') }.join('<br> ').html_safe : title # rubocop:disable Rails/OutputSafety
```
kimadactyl, line 6: oh ruby. this prob needs moving into a method now i look at it again, lol

its basically to even out the line lengths on longer titles and avoid widows

app/components/computer_access.html.erb
```diff
@@ -17,9 +17,9 @@
 
   <h3 class="h2--alt"><span>Computer access</span></h3>
   <p>Places with computers you can use</p>
-  <% if places&.any? %>
+  <% if @places %>
```
kimadactyl, line 20: do you think its confusing to still have references to `places` in the code? should we switch these all to `partners` or is it good to denote that its only places with a fixed address this can be at (even if we dont have that enforced rn)?

app/assets/stylesheets/components/footer.scss
```diff
@@ -0,0 +1,149 @@
+@import "variables_mixins";
+
+.footer {
+	background-color: $base-text;
```
kimadactyl, line 4: indents have gone mega big, they are 2 spaces elsewhere - if theres a way we can enforce this that'd be great (prettier?)

app/components/meta.rb
kerimhudson: This is the way to do a slot. We insert html elsewhere using this

kimadactyl: nice simple syntax just its the only place ive seen it. is it confusing to have both this and the initialise block? they are both doing the same thing right?

kerimhudson: I don't believe so. We still pass in variables i.e. the permalink here, so that's still needed. Initialise is just a way of passing anything to the construction of a class, at least in js that's what it is :D

kimadactyl: i was making a case for either having both variables passed in through the initaliser, or both passed in as `renders_one`. i think i dont quite get this and prob need to have a poke tho - happy to leave for now!

app/components/hero_component.rb
```diff
@@ -0,0 +1,14 @@
...
+class HeroComponent < ViewComponent::Base
+  def initialize(title, subtitle = '', schema = nil)
+    super
+    @title = clean_title(title)
+    @subtitle = subtitle
+    @schema = schema
+  end
+
+  def clean_title(title)
+    return title.length > 32 ? title.split.in_groups(2, false).map { |g| g.join(' ') }.join('<br> ').html_safe : title # rubocop:disable Rails/OutputSafety
```
kimadactyl, line 12: return is implicit in ruby - the last line of any function auto returns.

perhaps this ternary should be converted into an if so we can remove that rubocop:disable statement?

kimadactyl (approved):
couple potential comments still open but happy to ship this when you are now looking good!

app/components/hero_component.rb
kerimhudson: the rubocop is unhappy with the labelling of it as html_safe. I could probably rewrite this to generate the html code rather than just copy the existing code though.

app/components/computer_access.html.erb
kerimhudson: I think this is a bigger question we can tackle at a later date? Perhaps an issue dedicated to the convergence of these.

app/components/meta.rb
kerimhudson: I think it would be quite a lot to pass in just a link through another slot. I think having them as this is better because the permalink is consistent, whereas the other is dynamic.

app/components/hero_component.rb
kimadactyl: ohh i think this might be one of those "lots of hours have been wasted here" places lol, might be best to just leave it

app/components/computer_access.html.erb
kimadactyl: good idea

app/components/meta.rb
kimadactyl: AH ok i understand now!

PR #2757, opened 2025-05-28
fix: Implement new Eventbrite description format

james-collapse:
Resolves # 2672

## Description

- `event.description` was deprecated in 2021 - see [this blog post](https://www.eventbrite.com/platform/docs/changelog#:~:text=event.&text=description%20will%20be%20deprecated%20on,content%20to%20your%20event%20description.).
- The full event description can be retrieved with `get 'https://www.eventbriteapi.com/v3/events/{event_id}/description/'`.
- This change adds a new method, `get_event_description`, which makes the above API call.
- The full description HTML is then set on the `event` object.
- In theory, this should get sanitised by the `EventResolver`.

### Motivation and Context

It was observed that the full event description as it appears on Eventbrite was missing from events on PlaceCal (see # 2672).

### Type of change

### How Has This Been Tested?

Ran the `import_calendar` task for an existing Eventbrite calendar (ID 241).


kimadactyl:
Oh wow that's non trivial and really annoying huh!

Great catch and good work fixing. My brain is goo and I can barely remember the internals but will have another look tomorrow.

james-collapse:
I figured out how to set the description, although there's some failing tests I need to sort out.

kimadactyl:
oh - can you just refer back to the original bug and check that this is getting output via the API?

james-collapse:
Doesn't seem to work for me...

james-collapse:
I've managed to log in to admin.placecal-staging.org but I don't see Social Refuge & Queer Lit as a partner, or their calendar - do you?

kimadactyl:
its not merged!

Issue #2751, opened 2025-05-26
event/event_component.rb


rixatron:

rixatron:
Ah that's my fault, I think I pressed 'create sub issue' rather than create issue which removed it from the list and put it further down the page in a different box.

PR #2778, opened 2025-06-25
WIP: Upgrade event component

rixatron:
Resolves #[2751](https://github.com/geeksforsocialchange/PlaceCal/issues/2751)

## Description

I've had a bit of trouble with this one.  The two main issues were that it wouldn't compile after the change because of too many parameters in the init function.  I got around this by passing site in.

The other problem is that I'm having trouble getting the test (not checked in at the moment) to work.  The code is as below: 

```ruby
require 'test_helper'
require 'view_component/test_case'

class EventComponentTest < ViewComponent::TestCase
  setup do
    @site = Site
    @context = 'SomeContext'
    @event = Event.new
    @primary_neighbourhood = Neighbourhood.new
    @show_neighbourhoods = false
  end

  def test_component_renders_event_without_badge
    with_request_url('/events/11234', host: 'mossley.mossley.localhost') do
      render_inline(EventComponent.new(site: @site, context: @context, event: @event, primary_neighbourhood: @primary_neighbourhood, show_neighbourhoods: @show_neighbourhoods))
      assert_text 'SomeContext'
      assert_text 'EventName'
      assert_text 'Second'
    end
  end
end
```

From the code above I get the error: "ActionController::UrlGenerationError: No route matches {:action=>"show", :controller=>"events", :format=>"json", :id=>nil}, possible unmatched constraints: [:id]"

the url is meant to supply the id parameter to the controller so I'm not sure why that's not working.  I've done some manual testing but not sure if we want to wait for me to figure out the testing.

### Motivation and Context

Resolves #[2751](https://github.com/geeksforsocialchange/PlaceCal/issues/2751)

Include anything else we should know about.-->

### Type of change

- [ ] Documentation update

### How Has This Been Tested?
I've just manually visited the screens involved, I can put some screenshots involved.

Normal CI pipeline


rixatron:
> this is off the top of my head but maybe gets you going again?
That's great thanks I'll give it a go : )

app/models/neighbourhood.rb
```diff
@@ -5,6 +5,7 @@ class Neighbourhood < ApplicationRecord
   #    see /lib/tasks/neighbourhoods.rake
   LATEST_RELEASE_DATE = DateTime.new(2023, 5).freeze
 
+  self.primary_key = 'id'
```
rixatron, line 8: I actually didn't add this manually so not sure where these model changes have come from but the tests don't seem to work without them? Could use some advice on this I think

app/components/event_component.rb
```diff
@@ -34,7 +39,7 @@ def duration
   end
 
   def formatted_date(date)
-    if date.year == Time.now.year
+    if date.year == Time.zone.now.year
```
kimadactyl, line 42: what does the extra `zone` method do? :)

test/components/event_component_test.rb
```diff
@@ -0,0 +1,28 @@
...
+      @site = create(:site)
+      @context = 'SomeContext'
+      @event = create(:event)
+      @primary_neighbourhood = create(:neighbourhood)
+      @show_neighbourhoods = false
+    end
+  end
+
+  def test_component_renders_event_without_badge
+    with_request_url('/events/11234.json', host: 'mossley.mossley.localhost') do
```
kimadactyl, line 18: this looks a little odd - normally you'd visit `event_path(@event)`? how come we're loading json?

app/components/event_component.rb
rixatron: This was something I seem to remember having to add for a linting violation

```
app/components/event_component.rb:42:26: C: [Correctable] Rails/TimeZone: Do not use Time.now without zone. Use one of Time.zone.now, Time.current, Time.now.in_time_zone, Time.now.utc, Time.now.getlocal, Time.now.xmlschema, Time.now.iso8601, Time.now.jisx0301, Time.now.rfc3339, Time.now.httpdate, Time.now.to_i, Time.now.to_f instead.
    if date.year == Time.now.year
```
adding the Time.zone just explicitly links it to the current time zone which I think can either be set by environment and just defaults to whatever timezone the server is currently using.  For example if I run `bin/rails console` and then `Time.zone` I get

```
#<ActiveSupport::TimeZone:0x00007bbfc7f66960
 @name="Europe/London",
 @tzinfo=#<TZInfo::DataTimezone: Europe/London>,
 @utc_offset=nil>
```

and Time.zone.now returns `Fri, 04 Jul 2025 20:59:27.440072070 BST +01:00`.  I'm not 100% familar with ruby but my understanding from the [docs](https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html) is that basically this is the same behaviour we got before but now we're effectively declaring it explicitly rather than relying on implicit behaviour.  If you wanted to start using UTC then we could just declare that instead but I think what's there is correct as far as I can tell but let me know if you think differently or want me to dig deeper on it.

test/components/event_component_test.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'view_component/test_case'
+
+class EventComponentTest < ViewComponent::TestCase
+  setup do
+    VCR.use_cassette(:import_test_calendar) do
+      @site = create(:site)
+      @context = 'SomeContext'
```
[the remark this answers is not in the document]
rixatron: Yes sorry this was something I meant to come back to and then completely forgot about it.  I've now added a few more tests that more completely test the different ways the page can render depending on the state but let me know if you think I've missed anything.

app/models/neighbourhood.rb
rixatron: I've removed this now, I think it's from before I got the hint about the factories so I think it should work without it 🤞

kimadactyl:
hellloo sorry this has taken me so long last couple weeks have been mad

i think overall this looks good but it now needs a bit of a spring clean

- check the files changed tab - i think a lot of these shouldnt be in this pr
- on reflection this shouldnt be using VCR gem at all. this is more to simulate browser response which is WAY too integrated / high level for this ticket. we should just be making an @event model i think manually using plain ruby
- nono to hard coded page IDs - these will change based on test environment and make for flaky tests. you just need to visit `@event_path(@event)` or similar

the basic logic is sound though! just needing a little work to make this a bit more of a self contained PR

app/assets/stylesheets/application.scss
```diff
@@ -57,3 +57,4 @@
 @import "components/pull_quote_component";
 @import "components/resource_not_found";
 @import "components/steps_component";
+@import "components/event_component";
```
kimadactyl, line 60: remember to alphebetise these

rixatron: Sorry for dropping off the face of the earth after this comment. It's a long story but basically I accidentally messed up my computer and had to reinstall everything but I should be able to come back and finish this this week all being well.

kimadactyl: all good and no hurry we are deffo all at low ebb atm!

PR #2800, opened 2025-09-07
build: fix the users seed so the extra create_user action isn't needed

ghost:
Resolves # 

## Description

The users created in the `001_users.rb` db seed aren't usable, so an extra makefile task `create_user` was added. This change fixes the users in the seed and removes the now unnecessary makefile action

### Motivation and Context

Improves developer onboarding process, removing the need to perform an undocumented step during initial setup

### Type of change

### How Has This Been Tested?

In Alice's local Docker development environment

In development environments this gets run as part of either `./bin/setup` or `bundle exec rails db:seed` or `bundle exec rails db:seed:replant`

In production this shouldn't need any deployment, but I am unaware of how this might impact the production environment

### DIFF
--- a/db/seeds/001_users.rb
+++ b/db/seeds/001_users.rb
@@ -1,47 +1,17 @@
 # frozen_string_literal: true
 
 module SeedUsers
-  # role: root, editor, citizen
-
-  USER_INFO = [
-    {
-      first_name: 'root',
-      last_name: 'placecal',
-      role: 'root',
-      phone: '301-688-6311',
-      email: 'root@placecal.org'
-    },
-    {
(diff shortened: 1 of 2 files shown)


db/seeds/001_users.rb
```diff
@@ -1,47 +1,17 @@
...
 
-    USER_INFO.each do |user_info|
-      user = User.create!(
-        user_info.merge(
-          password: 'password',
-          password_confirmation: 'password'
-        )
-      )
-
-      user.skip_invitation = true
```
kimadactyl, line 41: hmm do we really not need these 3 lines?

PR #2817, opened 2025-10-28
docs: document security policy

ghost:
Add a basic security policy to appear on the Security section of GitHub

## Description

Use GitHub's template to add basic information

### Motivation and Context

Makes it easier for people to report security issues by making the information very visible within GitHub

### Type of change

- [x] Documentation update

### How Has This Been Tested?

vale lint passes

No additional steps needed, it will be automatically shown on GitHub under the Security tab

### Screenshots

Before:

Screenshot_2025-10-28_19-20-24

After:

Screenshot_2025-10-28_19-22-00

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,7 +4,7 @@
 
 If you discover a security vulnerability in this code please report it to:
 
-Email: [kim@gfsc.studio](mailto:kim@gfsc.studio)
+Email: [security@gfsc.studio](mailto:security@gfsc.studio)
 
 Please include the following information in your report:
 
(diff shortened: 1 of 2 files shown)


kimadactyl:
Thanks for this. Something went supremely screwy with my push that managed to bypass branch protection but this should be going to `security@gfsc.studio` (an alias of my email for now), and the security.txt is aso meant to be in `public/.well-known`. All a bit of a mess but if you have a sec to update all this to be in the right place with the right email that would be super appreciated.

I'll investigate how I was able to bypass branch protection soon as I'm still a bit clueless how that happened and its left this work in this limbo state until I have a mo to get back to it.

kimadactyl (approved):
great ty!

PR #2819, opened 2025-10-31
feat: add support for webp images

### LINKED ISSUE #2539: Support webp images where appropriate
## Use case

Webp is a more "modern" image format - we should support it

## Acceptance criteria

- [ ] Partner images can be webp
- [ ] Site headers can be webp
- [ ] Site logos can be webp

## Implementation plan

ghost:
Add support for webp images wherever images can be uploaded

Update tests to reflect this change

Resolves #2539 

## Description

- Update all uploaders to additionally support webp file extension
- Update tests to expect this new file extension in error messages

### Motivation and Context

https://caniuse.com/webp

> WebP generally has better compression than JPEG, PNG and GIF and is designed to supersede them. [...] Widely available across major browsers

Additionally, many other websites use WebP format and site administrators are likely to have images in this format that they want to use

Whilst jpegxl is expected to replace webp in the future, there is very little support for jpegxl as of late 2025

### Type of change

- [x] New feature (non-breaking change which adds functionality)

### How Has This Been Tested?

Manual testing by uploading a webp image and seeing it being used in the development ad-astra site

Assuming the imagemagick libraries being used already support webp, there should be no special requirements for this deployment


pedantic-git:
@kimadactyl yeah and that's what a class tree is really good at - a parent class with the common requirements and subclasses to override only the bits that are different.

kimadactyl:
image

confirmed working on staging

PR #2823, opened 2025-11-05
chore: Bump versions of Puma and Rack

pedantic-git:
## Description

This PR bumps the versions of Puma and Rack in the `Gemfile.lock`.

It also updates `schema.rb` to 7.2 which happened automatically for me - it must have just failed to be checked in before now.

### Motivation and Context

Spinnning up a dev env from scratch was impossible with the current `main` version of `Gemfile.lock`. The reason is:

1. `rackup` was pinned to v1.0.1 which I think [is a broken release as it doesn't even contain the rackup/server.rb file](https://github.com/rack/rackup/tree/v1.0.1/lib/rackup).
2. `Rackup::Server` is needed and expected by yard.

Prior and subsequent versions of `rackup` do have the necessary file, but bumping `rackup` by itself isn't possible since its 2.x series depends on Rack >= 3.x which in turn depends on Puma >= 6.x.

Bumping both Rack and Puma to their latest versions allowed my dev env to compile but that's as far as I've tested it so it will need a verification on staging.

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

Bumping both Rack and Puma to their latest versions allowed my dev env to compile but that's as far as I've tested it so it will need a verification on staging.

Normal process.


kimadactyl (approved):
lgtm thankyou!

PR #2827, opened 2025-11-11
chore: Bump dependencies suggested by renovate

pedantic-git:
@kimadactyl This is a PR to bump the deps that renovate raised PRs for. I don't actually know what PlaceCal's policy is on dep management - why renovate chose those deps specifically when they had fixed versions listed. But hopefully this is fine for this release.

I've given it a cursory check to make sure basic functionality works but nothing more than that.


pedantic-git:
Oh I seeeeee! Yeah sure

PR #2828, opened 2025-11-11
Get Rubocop clean

pedantic-git:
# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas,

## Description

This PR cleans up Rubocop so it generates a clean output.

It also changes the behaviour of a couple of methods that I believe were not behaving as they should:

* `Join#submit` was written so it would return `true` even if `#deliver` fails. I assumed this was an error but it's easy to make it work this way by changing its implementation to use `Object#tap`.
* It deletes a subclass method in `CalendarImporter::Events::ResidentAdvisor` that was only really comments and could inherit from the base class.

### Motivation and Context

Just a good excuse to keep the code style clean.

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

Just run the normal test suite.

### DIFF
--- a/app/jobs/calendar_importer/event_resolver.rb
+++ b/app/jobs/calendar_importer/event_resolver.rb
@@ -26,7 +26,7 @@ class CalendarImporter::EventResolver
   end
 
   def determine_online_location
-    data.online_address_id = data.online_event?
+    data.online_address_id = data.online_event_id
   end
 
   def determine_location_for_strategy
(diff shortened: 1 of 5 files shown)


kimadactyl (approved):
wonderful, thankyou this is doing  a ton for readability!

funnily enough that method naming has been confusing me a whole bunch its way clearer now!

PR #2840, opened 2025-11-25
deps: Wait 7 days before accepting updates

ghost:
Resolves Discord discussion

## Description

Add minimumReleaseAge to rennovate in order to enforce a cooldown so that we don't auto-merge vulnerable libraries

### Motivation and Context

https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

By waiting 7 days the theory is that security scanners and other organisations will already have identified malware in published libraries.

This PR applies that to both npm and ruby (docker isn't supported)

### Type of change

### How Has This Been Tested?

The rennovate.json configuration here is a lot more complicated than I've seen on other projects, so this is a bit of a guess as to whether it will work as expected. I'm hoping that rennovate bot will see the PR and tell us if it looks sensible.

Rennovate should just pick this up and start using it for any future dependency updates. And hopefully will reflect this in the dashboard.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -40,5 +40,11 @@
 			"matchCurrentVersion": "13.4",
 			"allowedVersions": "13.4"
 		}
-	]
+	],
+	"npm": {
+		"minimumReleaseAge": "7 days"
+	},
+	"rubygems": {
+		"minimumReleaseAge": "7 days"
+	}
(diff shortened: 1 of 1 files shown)


kimadactyl:
awesome thankyou!

PR #2852, opened 2025-12-14
WIP fix: use web ical exporter rather than api

ghost:
## Description

The URL currently used for meetup events appears to no longer work. But this new URL does, at least for the one group I've tried it on.

This needs feedback from users to see if meetup events are importing for anyone else, and whether this new URL is going to work for regular imports.

### Motivation and Context

Originally spotted when looking at how to import [Queer in Tech](www.meetup.com/queer-in-tech-inclusive-tech-meetup/) meetups, the url `https://www.meetup.com/queer-in-tech-inclusive-tech-meetup/events/ical` provides an ical file for download.

### Type of change

### How Has This Been Tested?

I've not tested this at all yet, I'm opening this up for discussion

standard deploy process


kimadactyl:
from a very quick glance it might make more sense to just fall back to the default ical importer for these (or inherit from it) and update the handbook to describe how to get the link?

PR #2860, opened 2025-12-18
Modernise robots.txt

ghost:
## Description

Updates production robots.txt to use latest version from ai.robots.txt to block AI crawlers

Tells Bing and Amazon to not use the data for AI training purposes

### Motivation and Context

- [ai.robots.txt](https://github.com/ai-robots-txt/ai.robots.txt)
- [extra steps for Bing](https://github.com/ai-robots-txt/ai.robots.txt/blob/main/docs/additional-steps/bing.md)

This is the polite way of telling AI crawlers to stop it, and builds upon the earlier attempts to achieve this

### Type of change

### How Has This Been Tested?

The contents of ai.robots.txt and the new meta tag are pretty well tested by the users of ai.robots.txt

Standard deploy

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -22,6 +22,11 @@ class PagesController < ApplicationController
   def terms_of_use; end
 
   def robots
-    render plain: current_site.robots
+    if current_site
+      render plain: current_site.robots
+    else
+      # Admin subdomain or no site found - disallow all indexing
+      render plain: "User-agent: *\nDisallow: /"
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


kimadactyl:
I am in two minds what to do about AI bots long term.

On the one hand, they chew up server resources, perhaps mitigated if we ever got Cloudflare proxy turned on properly. While there are plans to give micropayments for scrapes I don't think this is up and running yet and the volume to make it worth it is prob a long way off.

On the other hand, if people are asking AI bots for events near them it's nice if they can get accurate data linked to our site which they then might use directly in the future. As a kind of "gold standard" data site we are a positive for the many people using ChatGPT et al as a daily driver.

Agree having it in the middle is prob a bit of a nothingness but it would be nice to express what we are trying to achieve here in terms of a business rule (which we can define more easily when my cucumber branch is added)

kimadactyl:
ah sorry the test failiure is because ive moved to matrix testing on my new PR and that requires changing the branch settings in the github interface itself to run properly - itll work again when thats merged in

Issue #2770, opened 2025-05-30
[Bug]: Incorrect times for OutSavvy events


james-collapse:
## Description

This [event](https://fruitsaladevents.placecal.org/events/401426) appears as 12PM - 7PM:

![Image](https://github.com/user-attachments/assets/6d6f01b8-3617-4c2f-9405-5c5add04544c)

However, the [original event](https://www.outsavvy.com/event/25977/ldn-queer-mart-lgbtqia-art-market) is from 11AM - 6PM:

![Image](https://github.com/user-attachments/assets/a1c6218e-9726-4622-a80b-52257f268cd6)

## Steps to reproduce

1. What you did first
1. What you did next

## What you expected to happen

What would you like to happen instead?

## Platform (if relevant)

What device and browser were you using?

kimadactyl:
assume this is a timezone issue again

kimadactyl:
they seem pretty on it i would love it if they can just fix this their end! great detective work

kimadactyl:
did you ever get a reply @james-collapse ? do you know if this is still an issue?

PR #2909, opened 2026-01-28
Fix OutSavvy address and time issues

### LINKED ISSUE #2767: [Bug]: GraphQL query not returning address for OutSavvy event
## Description

The address for [this event](https://fruitsaladevents.placecal.org/events/401426) is appearing in PlaceCal:

Image

However, it is not appearing when queried:

Image

What were you doing?

## Steps to reproduce

1. What you did first
1. What you did next

I would expect the same address to appear in the `address` field.

## Platform (if relevant)

kimadactyl:
## Summary
Fixes #2767 and #2770

## Problem
1. **#2767**: OutSavvy events had address data but it wasn't appearing in the GraphQL API. The address was visible in the UI but the `address` field returned null.

2. **#2770**: OutSavvy events showed times 1 hour off. An event listed as "11AM - 6PM" on OutSavvy was showing as "12PM - 7PM" in PlaceCal.

## Root Causes
1. **Address issue**: The JSON-LD context was missing mappings for `addressLocality`, `addressRegion`, `postalCode`, and `addressCountry`. Also, `LinkedDataEvent#extract_location` was only using `street_address`, discarding the postcode and other components.

2. **Time issue**: OutSavvy produces malformed ISO 8601 timestamps like `2025-06-01T11:00:00:00+01:00` (note the extra `:00` before the timezone). This causes `DateTime.parse` to ignore the timezone offset.

## Solution
1. Added missing address field mappings to the JSON-LD `CONTEXT`
2. Updated `LinkedDataEvent#extract_location` to combine all address components
3. Created `OutsavvyEvent` class (extends `LinkedDataEvent`) that fixes the malformed timestamps
4. Updated `Outsavvy` parser to use a custom `EventConsumer` that creates `OutsavvyEvent` instances

## Testing
- Added `spec/jobs/calendar_importer/events/linked_data_event_spec.rb`
- Added `spec/jobs/calendar_importer/events/outsavvy_event_spec.rb`
- All 163 calendar importer tests pass

### DIFF
--- a/app/jobs/calendar_importer/events/linked_data_event.rb
+++ b/app/jobs/calendar_importer/events/linked_data_event.rb
@@ -38,8 +38,17 @@ module CalendarImporter::Events
       return if loc.blank?
 
       address = loc['address']
-      address = address['street_address'] if address.is_a?(Hash)
-      address
+      return loc['name'] unless address.is_a?(Hash)
+
+      # Build full address from components
+      components = [
+        address['street_address'],
+        address['address_locality'],
+        address['address_region'],
+        address['postal_code']
+      ].compact_blank
+
+      components.join(', ').presence || loc['name']
     end
 
     def attributes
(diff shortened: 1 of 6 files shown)


james-collapse (approved):
Looks great to me, thanks for fixing this!

Issue #1812, opened 2023-04-27
Implement refreshed PlaceCal design system


katjam:
XD walk through of everything:
https://xd.adobe.com/view/d737e480-ff3a-4e07-8d32-339f902ddf4b-8264/

Dropbox folder containing assets:
https://www.dropbox.com/sh/v4ifllv0tki83tu/AAClxFi1k6NJ4FAWXNwuoAdBa?dl=0
Contains a few new SVGs, the master XD file, and PNG backings of XD file.

## Acceptance criteria

- [x] #1988
- [x] Site home page text is pushed into user control (#1699)
- [ ] Existing sites have descriptions either migrated from what it is now or added manually (maybe a good chance to cull dead sites)
- [ ] #1943
- [ ] [Update with breakdown of components to implement]

## Implementation plan

- #3000 
- #3010 
- Attempt automigrate of Sass -> Tailwind

katjam:
This may or may not be a thing we want to do - to keep consistency between new homepage / info site & the sub-sites

studiosquid:
Designing in consistency of the sub-sites was something that was originally requested (but I realise PlaceCal is pretty fluid) reason was those are all working on the very original 2017 designs and I believe we've run into quite a lot of issues since they first began. 

But also there feels like quite a lot of push for either more generic or more customisable versions. (Or both!)

I did do a draft where I took out all the PlaceCal colours and took it a more neutral place except for an accent colour. I parked this for now as it tended to look very cold and unwelcoming, and the accent colour would have to be either manually checked, or picked from a preset list of colours so that accessibility isn't scrambled.

katjam:
Removing from priorities - but is a good idea & has designs

r-ferrier:
I've also added another item to the acceptance criteria on #1699 related to this

kimadactyl:
just have a read of the other ones. doesnt have to be fancy - just a record of the problem definition and what options you considered. if you open a pr with just that then we can discuss before geetting stuck in :)

PR #2996, opened 2026-02-26
Decide future design system implementation

soyaEnjoyer:
References: https://github.com/geeksforsocialchange/PlaceCal/issues/1812


kimadactyl:
As mentioned i'm [happy to proceed](https://github.com/geeksforsocialchange/PlaceCal/pull/3000#issuecomment-3976721924) with Phlex + Tailwind if you are. Just update the ADR first to outline why and have a bit of a think what theming will look like :) i think itll actually be much easier as everything is in ruby now so we can do a lot of conditional stuff or have some theme tooling and so on.

kimadactyl (approved):

PR #2997, opened 2026-02-26
Add Y offset to leaflet_controller for marker height

### LINKED ISSUE #2033: [Bug]: Pin locations on the map aren't accurate
## Description

When looking at the map on the Partners page of norwich.placecal.org the location pins are inaccurate enough to possibly cause confusion when locating partners. 

I have included a screenshot below which shows the map as is, with X in the approximate correct location for the addresses given. 
Screenshot 2023-09-04 at 09 07 06

This could cause issues if event attendees rely on the PlaceCal map to find the entrance of a site or venue.

soyaEnjoyer:
Resolves #2033

### Type of change

### How Has This Been Tested?

Updated coords on and address in my local db to match the coords from the example in the linked issue
Panned and zoomed and all looks ok
image

image

image

### DIFF
--- a/app/javascript/controllers/leaflet_controller.js
+++ b/app/javascript/controllers/leaflet_controller.js
@@ -43,9 +43,13 @@ export default class extends Controller {
 
 			iconSize: [46, 43], // size of the icon
 			shadowSize: [46, 43], // size of the shadow
-			iconAnchor: [17, 43], // point of the icon which will correspond to marker's location
-			shadowAnchor: [17, 43], // the same for the shadow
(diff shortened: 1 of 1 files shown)


idkidk000:
Yes, it looks like it. I didn't even know that map was there actually
Firefox:
image
image
image
image

Chromium:
image
image

FF mobile layout:
image
image

Chromium mobile layout:
image
image

---
EDIT: hang on. Clicking the pointer shows a label and that also needs an offset. I'll fix and add notes

PR #3000, opened 2026-02-27
Replace all ViewComponents with Phlex

kimadactyl:
## Summary

- Migrate all ViewComponents from ViewComponent+ERB to **Phlex 2** with inline Ruby templates
- Add `phlex-rails` gem with Zeitwerk autoloading under `Components::` namespace with `Phlex::Kit`
- Delete all `.html.erb` template files — each component is now a single `.rb` file
- Remove `view_component` gem dependency
- Update all render calls across admin and public ERB templates
- Remove all `rubocop:disable Metrics` comments by extracting private render methods, moving SVG heredocs to constants, and deleting unnecessary disables
- Consolidate and inline trivial components, reducing total count from 68 to 61
- Net result: **-1,152 lines** (4,676 added / 5,828 removed) across 360 files

### Why Phlex?

- **No template files** — each component is a single `.rb` file instead of `.rb` + `.html.erb`
- **Ruby-native DSL** — HTML is built with method calls, enabling better composition and refactoring
- **Kit syntax** — components render as `Breadcrumb(...)` instead of `render(BreadcrumbComponent.new(...))`

### What changed

| Area | Details |
|---|---|
| New gem | `phlex-rails ~> 2.3` (installed 2.4.0) |
| Removed gem | `view_component` |
| Initializer | `config/initializers/phlex.rb` — Zeitwerk `Components::` namespace with `Phlex::Kit` |
| Base classes | `app/components/base.rb`, `app/components/admin/base.rb` |
| 31 admin components | Migrated to `app/components/admin/` |
| 30 public components | Migrated in `app/components/` (renamed from `*_component.rb` to `*.rb`) |
| Views | Updated all `render` calls to use Kit syntax |
| Tailwind | Added `@source` for class scanning in `.rb` files |
| Rubocop | Zero `rubocop:disable Metrics` comments — all methods under 10 lines via extraction |

### Component cleanup

As part of the migration, trivial wrapper components were inlined and similar components were consolidated:

| Change | Components |
|---|---|
| **Consolidated into `HelpCard`** | `ComputerAccess`, `FreePublicWifi`, `HelpGettingHelp`, `HelpAddingEvents` → single component with variant pattern |
| **Inlined as plain HTML/CSS** | `MaxWidth`, `HomeStrapline`, `LinkBtnLrg`, `OpeningTimes`, `ContainerWithHeader`, `FourColGrid`, `FeaturedPartnerships`, `ImpactStory` |
| **Simplified** | `ContactDetails` — removed 8 unnecessary instance variables |

### raw/safe patterns

Established correct Phlex 2 patterns for rendering HTML content:

- **Rails form helpers** (return `SafeBuffer`) → `raw(@form.check_box(...))`
- **Plain strings with HTML** (SVGs, i18n) → `safe(SVG_CONST)` as last expression, or `raw(safe(SVG_CONST))` if not last
- **Plain strings without HTML** → just return them (Phlex escapes safely)

## Test plan

- [ ] Admin component specs pass (`spec/components/admin/`)
- [ ] Public component specs pass (`spec/components/`)
- [ ] System/integration tests pass
- [ ] Rubocop clean (61 component files, 0 offenses)
- [ ] Dev server boots and renders admin + public pages without errors
- [ ] Visual smoke test of admin forms, datatables, flash messages, and public site

### DIFF
--- a/spec/components/details_component_spec.rb
+++ b/spec/components/details_component_spec.rb
@@ -15,37 +15,109 @@ RSpec.describe Components::Details, type: :phlex do
     }
   end
 
-  it "renders header" do
+  it "renders a details element with summary" do
     render_inline(described_class.new(**attrs))
 
-    expect(page).to have_selector("h1")
+    expect(page).to have_selector("details.details")
+    expect(page).to have_selector("details summary")
+  end
+
+  it "renders header at the specified level" do
+    render_inline(described_class.new(**attrs))
+
+    expect(page).to have_selector("h1.details__header")
     expect(page).to have_text("The header value")
   end
 
-  it "renders summary" do
+  it "applies custom header class" do
     render_inline(described_class.new(**attrs))
 
+    expect(page).to have_selector("h1.the-header-class")
+  end
+
+  it "defaults to h3 when no header_level specified" do
+    render_inline(described_class.new(header: "Test", summary: "Summary"))
+
+    expect(page).to have_selector("h3.details__header")
+  end
+
+  it "does not render header when nil" do
+    render_inline(described_class.new(summary: "Summary only"))
+
+    expect(page).to have_no_selector(".details__header")
     expect(page).to have_selector(".details__summary")
   end
 
-  it "renders image alt text" do
+  it "renders plain text summary in a paragraph" do
+    render_inline(described_class.new(summary: "Plain text summary"))
+
+    expect(page).to have_selector(".details__summary p", text: "Plain text summary")
+  end
+
+  it "renders html_safe summary without extra wrapper" do
+    render_inline(described_class.new(summary: "<p>Already wrapped</p>".html_safe))
+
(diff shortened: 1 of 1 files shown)


kimadactyl:
alrighty wow this is looking good. i think i am a convert. this feels way nicer than view component ever did and yeah the type indicators are very good - ive not lived in this kinda ecosystem before but i see why you like it now.

- helper warnings fixed
- ive got it to make the types more specific where it can (can prob be improved as we go too)
- locale strings fixed
- collections code is preexisting (this is currently unsupported)
- ditto supporters code

im extremely happy with how this looks and feels. if youre happy to go ahead i suggest following course of action:

1. update your ADR docmenting this course of action and lets get that merged
2. lets get your icons fix merge in
3. Consider if we should get a release out here in case we start finding more issues we dont want on prod yet
4. merge this pr
5. make tickets for the rest of the migration: converting views and templates and then per-component tailwind migrations + updates to new design system

what do you think?

Issue #3015, opened 2026-03-04
Site header SVG logos not rendering (inline_svg path issue)


kimadactyl:
## Problem

Site header logos that are uploaded SVGs don't render. Instead, the HTML shows:

```html
<svg><!-- SVG file not found: 'site/logo/35/manchester_libraries.svg' --></svg>
```

Visible on e.g. https://manchesterlibraries.placecal.org/

## Cause

In `app/components/navigation.rb:9`, the logo path is built by stripping `/uploads/`:

```ruby
@logo_path = @site&.logo&.to_s&.sub(%r{^/uploads/}, '').presence
```

This gives `site/logo/35/manchester_libraries.svg`, which is then passed to `svg_image` → `inline_svg_tag`. But `inline_svg` looks in `app/assets/images/`, not `public/uploads/`, so the file is never found.

Non-SVG logos work fine because they go through `image_tag(@site.logo.url)` instead.

## Fix

For SVG uploads, read the file directly from `public/uploads/` instead of using `inline_svg_tag`. Something like:

```ruby
def render_logo
  if @logo_path
    if /\.svg$/.match?(@logo_path)
      file_path = Rails.root.join('public', 'uploads', @logo_path)
      if File.exist?(file_path)
        raw(File.read(file_path))
      else
        image_tag(@site.logo.url, alt: @site.name)
      end
    else
      image_tag(@site.logo.url, alt: @site.name)
    end
  else
    raw(view_context.svg_image('home/icons/logo.svg', alt_text: 'PlaceCal'))
  end
end
```

## Affected files

- `app/components/navigation.rb`

PR #3032, opened 2026-03-05
Correct nav spacing and logo behaviour

soyaEnjoyer:
I previously got the nav spacing quite mixed up between the default and partner sites
Additionally, my logic was incorrect for the nav logo and the documentation on SvgImagesHelper was unclear. 

This PR corrects both

# Checklist:

- [x] Title include "WIP" if work is in progress.

Resolves # 

## Description

Nav spacing is not pixel perfect at every breakpoint for the default site and partner sites because I've tried to use clean(er) multiples and be a bit more consistent with values as much as possible, but it's very close. I can try harder to make it pixel perfect if you like.

Nav logo logic is now tidied up and commented. Default site uses the home logo, partner sites will inline an svg if they have an svg, add an image tag if they have an image, and fallback to the default header logo (different colours to the home logo). Previously this was falling back to the home logo, which was incorrect. Behaviour between this branch and main prior to the SVG images helper is consistent for me in dev.

Note: there was an issue https://github.com/geeksforsocialchange/PlaceCal/issues/3015 where the Manchester Libraries SVG logo was not inlining in prod. I've checked out main prior to Kim's fix but **I cannot replicate locally with a recent dump of the prod database and a copy of prod uploads**. The documentation for `svg_image` in `SvgImagesHelper` was incorrect (it did not mention that I had added `Rails.root.join(Rails.public_path, 'uploads')` to the asset path) so I have fixed that. **I'm a bit concerned about this and wonder if dev/prod asset behaviour is different somehow?** I *could* use Kim's code which reads the SVG file directly, but it means we then have two ways of inlining SVG images.

### Type of change

### How Has This Been Tested?

Tested at 450, 650, 950, and 1250px breakpoints using modern and old Firefox and Chromium
Tested the default site and various partner sites with none, SVG, and PNG images
Tested this branch against a main prior to the SVG merge

### Screenshots

I don't think this works very well on a GitHub ticket but here we go. "Pre" is main prior to the SVG commit. "Post" is this branch right now. Both are using the same recent copy of the prod DB and a current copy of prod uploads. 
I only have latest Firefox screenshots because this would get out of hand very quickly.
These are 450, 650, 950, 1250 px breakpoints in their collapsed and expanded states.
I'll keep these images locally for a while in case you want a copy directly.

Default site pre
home_pre_1250
home_pre_950
home_pre_650_exp
home_pre_650_col
home_pre_450_exp
home_pre_450_col

Default site post
home_post_1250
home_post_950
home_post_650_exp
home_post_650_col
home_post_450_exp
home_post_450_col

Manchester libraries pre
mlib_pre_1250
mlib_pre_950
mlib_pre_650
mlib_pre_450_exp
mlib_pre_450_col

Manchester libraries post
mlib_post_1250
mlib_post_950
mlib_post_650
mlib_post_450_exp
mlib_post_450_col


idkidk000:
OK go for it then. It'll deploy to staging first right, so I have a chance to fix it before prod if I have broken it again?

PR #3074, opened 2026-03-24
Add visual regression testing for CSS migration

### LINKED ISSUE #3066: Add visual regression testing for CSS migration
## Context

PR #3020 (Tailwind migration + Sprockets→Propshaft) is too large to review visually by hand. @idkidk000 suggested adding visual regression testing that snapshots every public page at multiple viewports, before and after changes, and diffs the results. We should merge this as a separate PR first, then use it to validate #3020.

## What we already have

- Capybara + Selenium + headless Chrome (`spec/rails_helper.rb`)
- `lvh.me` subdomain routing for partner site theming
- System specs running against a real browser

## Proposed approach

A spec file (`spec/system/visual_regression_spec.rb`) that:

1. Seeds test data (events, partners, a partner-themed site)
2. Visits every public route: homepage, find-placecal, partners index/show, events index/show, news, privacy, terms, our-story, get-in-touch, collections, plus a partner-themed site
3. At each of 450px, 650px, 950px, 1250px
4. Saves screenshots to `tmp/screenshots/{page}_{width}.png`

### Workflow

```bash
# On main branch
git checkout main
bundle exec rspec spec/system/visual_regression_spec.rb
mv tmp/screenshots tmp/screenshots_before

# On PR branch
git checkout the-css-branch
bundle exec rspec spec/system/visual_regression_spec.rb
mv tmp/screenshots tmp/screenshots_after

# Diff with ImageMagick
mkdir -p tmp/screenshots_diff
for f in tmp/screenshots_before/*.png; do
  name=$(basename "$f")
  compare "$f" "tmp/screenshots_after/$name" "tmp/screenshots_diff/$name" 2>/dev/null
done
```

### Screenshot diffing approach

- **Option A: Manual diff workflow (above)** — just a spec that takes screenshots, diff with ImageMagick manually. Simple, no new dependencies beyond ImageMagick. Good enough for a one-off migration validation.
- **Option B: `capybara-screenshot-diff` gem** — automates baseline storage and fails specs if diff exceeds a threshold. More infrastructure but reusable for future CSS changes. Might be overkill if we only need this for the migration.

### Browser coverage

- **Chrome only** — already configured, zero setup. Covers the majority of users.
- **Chrome + Firefox** — Selenium supports both, just need to register a second driver. Catches flexbox/grid rendering differences.
- **Chrome + Firefox + Safari** — Safari requires macOS + `safaridriver` (no headless mode). Could do this manually or skip.
- **Legacy browsers via BrowserStack/Sauce Labs** — significant setup, ongoing cost. Probably overkill unless we have known legacy browser users.

### Threshold

- **Pixel-perfect** — any diff is a failure. Strict but noisy (font rendering differences between runs can cause false positives).
- **Layout-level** — allow small anti-aliasing/font differences, flag structural changes. More practical.

## Questions for @idkidk000

1. Which screenshot approach (A or B)?
2. Which browser coverage level? Chrome-only seems fine for validating the migration, but open to more.
3. Any pages I've missed that should be covered?
4. Is ImageMagick `compare` good enough for diffing or do you have a preferred tool?

kimadactyl:
## Summary
- Adds `spec/system/visual_regression_spec.rb` that screenshots all public pages at 4 viewports (450/650/950/1250px)
- Adds `bin/visual-regression` runner that captures baseline from main, captures current branch, and diffs with ImageMagick
- Baselines cached by SHA — only regenerate when main moves forward

## Test plan
- [x] Spec passes: 14 examples, 56 screenshots
- [x] Two consecutive runs produce pixel-perfect results (52/56 identical, 4 have <1% timing noise)
- [x] Cached baseline skips recapture on second run
- [ ] Test on an actual CSS-change branch to validate diffs show real changes

Closes #3066


idkidk000 (commented):
Mostly looks good. I added a few little comments. [This cheat sheet](https://devhints.io/bash) is my go-to for all things Bash.
- Can you capture a full page screenshot so we also get the page content and footer?
- `find_placecal_*` is missing content
  from tmp/visual_regression: Image
  from browser: Image
- not now, but in the future it would be helpful if we can get Safari/WebKit and Firefox captures too. [Playwright](https://playwright.dev/docs/) looks promising

PR #3118, opened 2026-04-02
Add API rate limiting

### LINKED ISSUE #2607: [Bug]: API requests can take down whole site
## Description

I was messing around with the TransDim repo and noticed that the PlaceCal.org main site was down due to our AppSignal monitoring. On further investigation the requests being sent by TransDim locally effectivly denial-of-service'd the whole PlaceCal site.

It should not be this easy to take down the whole PlaceCal site and this needs mitigating somehow.

## Steps to reproduce

1. Load up TransDim repo
2. Set `PLACECAL_API=https://placecal.org/api/v1/graphql`
3. Muck about a bit
4. Note downtime from PlaceCal.org

## Platform (if relevant)

kimadactyl:
## Summary
- Adds `rack-attack` gem for IP-based request throttling (100 req/min on `/api/*`)
- Adds GraphQL `max_depth` (10) and `max_complexity` (1500) limits to prevent expensive queries
- Returns JSON 429 responses for throttled API requests

Closes #2607

## Test plan
- [x] Rate limiting specs pass (`spec/requests/rate_limiting_spec.rb`)
- [x] All existing GraphQL specs pass (34 examples, 0 failures)
- [x] Verified Trans Dimension queries are well within complexity limits (~100 each vs 1500 cap)


kimadactyl:
ah interesting thought - yeah lets start with this and see how we get on tho. appsignal gives reporting on slow running queries so we can be a bit targetted if we need to improve it further.

Issue #3067, opened 2026-03-23
Migrate public components


kimadactyl:
## Context

We are migrating from SCSS to Tailwind. Each component needs manually migrating and reviewing as we go:

1. **Check visual correctness** at 450, 650, 950, 1250px (pending visual regression testing from #3066)
2. **Replace remaining arbitrary values** with standard Tailwind sizes or theme variables where appropriate
3. **Unify link/anchor styling** — currently each component styles `<a>` differently
4. **Clean up code smells** — `<br/>` for spacing, inconsistent patterns

### Design references

- **XD walkthrough**: https://xd.adobe.com/view/d737e480-ff3a-4e07-8d32-339f902ddf4b-8264/
- **Dropbox assets** (SVGs, master XD, PNG backings): https://www.dropbox.com/sh/v4ifllv0tki83tu/AAClxFi1k6NJ4FAWXNwuoAdBa?dl=0

### Related issues

- #3020 — Tailwind/Propshaft migration PR
- #3066 — Visual regression testing
- #1943 — Allow a wide range of logos to be uploaded
- Supersedes #1812

## Components

- [ ] **Breadcrumb** — arbitrary `text-[0.8rem]`
- [ ] **CaseStudy** — arbitrary font sizes, button-style links need unifying
- [ ] **ContactDetails** — arbitrary grid cols `grid-cols-[auto_1fr]`
- [ ] **Flash**
- [ ] **Footer** — heavy use of descendant selectors `[&_a]`, `[&_ul]`, `[&_li]`; arbitrary `text-[0.9rem]`, `border-[3px]`, `max-w-[187px]`; link styles
- [ ] **FullWidthAction** — arbitrary `text-[1.3rem]`, `text-[1.8rem]`, `text-[2.2rem]`, `transition-[300ms]`; button-style links
- [ ] **HelpCard** — `border-b-[3px]` descendant selector
- [ ] **Hero** — arbitrary `w-[4.5rem]`
- [ ] **HeroSection** — arbitrary `gap-[1.55rem]`, `p-[1.55rem]`, font sizes `text-[1.77778rem]`, `text-[0.8rem]`
- [ ] **HomeFooter** — `border-t-[3px]`, `text-[0.66rem]`, `size-[0.9rem]`
- [ ] **Meta**
[...]

PR #3138, opened 2026-04-07
Enable Tailwind on public site and port Navigation component

### LINKED ISSUE #3025: Clean up `fonts.scss`
## Description

I have never seen so many font files. Check compat and standardise on probably woff/woff2

It's probably fine to still declare faces for font weights and styles we're not using, but we probably shouldn't preload them (not sure if we're doing that already - verify)

Will need to clean up anyway as part of the Tailwind migration, but I just want to make sure that I don't run into too many sidequests

## Motivation

Helps with maintainability and potentially reduces amount of data to transfer on page load for new users etc

soyaEnjoyer:
This PR (or more likely, series of PRs) will migrate the public pages and components from Scss to Tailwind

Resolves some or all of #3067 

## Description

So far, this
- undoes parts of the Tailwind reset where browser defaults were implicitly part of the existing Scss theme
- Begins the initial Tailwind and ports some of the core Scss over (this includes cleaning up font formats which we definitely don't need and updating Trocchi to a modern format)
- ports the Navigation (header) component (default and partner sites. The Footer component is already split in two - should I do the same for this?)

There are currently plenty of FIXMEs and TODOs - the draft PR is just to get feedback early. My plan is to work through the page on `/` from top to bottom, repeat for each nav item, them repeat for the partner sites

The current lengths are based on a mixture of fractions/multiples of of 16px, 18px, 10px, and percentages. Currently I'm porting them as-is, but I'd like to normalise them all to be based on the root element size, which is 18px. This will cause layout drift but should improve visual cohesion

I'm using some temporary utility classes, `with-reset` and `with-no-sass`, which will need to be cleaned up later

I haven't run any of the tests yet but will update them as needed

### How Has This Been Tested?

https://github.com/idkidk000/visual-diff (you need to run it locally) in current Firefox and Chromium

### Screenshots

TODO


soyaEnjoyer (commented):
Thanks for the reply

> hey sorry not sure what you mean here?

ah my mistake - i didn't submit the review so nothing would have shown up for you. try now please:  https://github.com/geeksforsocialchange/PlaceCal/pull/3138/changes/BASE..869a6576463d8757c4da3defc3a9c0ff4f0d8ef8#diff-85002d4e3441a096fb493ed3adbb4add79e7503e3554da8b44330c273df02954
image


> normalising units

yeah i figured as much. so basically my issue is that tailwind has a base measurement unit, which by default is 4px (1/4 of the default rem of 16px). we have to choose *something* and the least bad option is .25rem, which is 18px/4 = 4.5px. Multiples of 4.5px (well, 2.25px actually) can now be expressed cleanly as a tailwind class, i.e. `mx-6` means `inline margin of 6 * 4.5px`, but we have some sizes which are not multiples of 4.5px (we would have had more of these to deal with if i'd kept the default of 4px). Those currently are expressed as `mx-[1.222rem]` or similar and have no relation to the base measurement unit (which means that overriding it via a partner theme will do nothing, which is a misbehaviour). I'd like to round them to `mx-5`, for example, so that overriding theme variables will work consistently. It will cause slight layout drift but should look more consistent.

app/components/navigation.rb
```diff
@@ -38,43 +52,55 @@ def render_logo
...
-      ul do
-        li { active_link_to('Home', root_path) }
+    # FIXME: arbitrary values
+    li_css_class = [
+      'text-center',
+      ('max-md:bg-tertiary max-md:py-3' if @site&.default_site?),
+      ('py-4 max-md:bg-text transition-none max-md:transition-colors duration-300 max-md:hover:bg-tertiary md:max-lg:justify-evenly md:max-lg:grow' unless @site&.default_site?)
+    ]
+    # active_link_to is defined in app/helpers/application_helper.rb and wraps link_to, which is not a phlex component
+    a_css_class = [
```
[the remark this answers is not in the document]
kimadactyl: the safe nav thing is bothering me too, i wonder if we want to just have @site always set and add in some other simple methods here. this can also be `directory_site` (soon) and `admin_site` i think.

kimadactyl (approved):
this looks awesome!! -110k LOC, lol, is that all font files?

feel free to merge when youre happy!

app/tailwind/public_tailwind.css
```diff
@@ -1,5 +1,247 @@
...
-   existing SCSS styles (strips margins, heading sizes, etc). */
+/* Public site Tailwind CSS  */
+@import "tailwindcss";
+@import "./public/fonts.css";
+
+/* https://tailwindcss.com/docs/detecting-classes-in-source-files#ignoring-specific-paths
+   TODO: verify and maybe rework as `@import "tailwindcss" source(none);` then @source explicit roots only
+   TODO: also exclude non-admin from admin_tailwind.css
+*/
+@source not "../../app/assets";
```
kimadactyl, line 9: hmm my one concern is there really no other way to do this? this doesnt look very maintainable and i think mgiht start causing problems immediatley

soyaEnjoyer: Nah, it scans the entire project unless you tell it otherwise. Ideally the project would be broken up into (admin / home / partner / common) but I have no idea what that might look like (or if it's even possible) with Rails

idkidk000:
> this looks awesome!! -110k LOC, lol, is that all font files?
> 
> feel free to merge when youre happy!

Thanks. Yeah, mostly old fonts. Did you test it?

kimadactyl:
yeah i did i cant really tell the diff, great job :)

app/tailwind/public_tailwind.css
kimadactyl: we could deffo do that! do you want me to look into it before merging this or shall we merge first?

PR #3141, opened 2026-04-09
Add missing stimulus target to filter options. Fix edge cases in partner_filter_component_controller.js

soyaEnjoyer:
The reset buttons at the bottom of the Filter component dropdowns were broken. This fixes them

Part of https://github.com/geeksforsocialchange/PlaceCal/issues/3140

### Type of change

### How Has This Been Tested?

Firefox and Chromium, ran unit tests

### DIFF
--- a/app/components/filter.rb
+++ b/app/components/filter.rb
@@ -49,7 +49,10 @@ class Components::Filter < Components::Base
         @name,
         item[:id],
         selected?(item[:id]),
-        data: { action: submit_action_value },
+        data: {
+          action: submit_action_value,
+          "#{target_key}": target_value
+        },
         class: 'tag__button'
       )
       label_tag(
@@ -93,5 +96,13 @@ class Components::Filter < Components::Base
 
   def reset_action_value
     "click->#{@controller}##{@reset_action}"
+  end
+
+  def target_key
+    "#{@controller}-target"
+  end
+
+  def target_value
+    @name
   end
 end
--- a/app/javascript/controllers/partner_filter_component_controller.js
+++ b/app/javascript/controllers/partner_filter_component_controller.js
@@ -12,82 +12,93 @@ export default class extends Controller {
 		"neighbourhoodDropdown",
 	];
 
+	/** Workaround for Stimulus's lack of typing
+	 * @returns {{
+	 * 		categoryDropdownTarget: HTMLElement | undefined,
+	 * 		categoryTargets: HTMLInputElement[],
+	 * 		categoryTextTarget: HTMLElement | undefined,
+	 * 		formTarget: HTMLFormElement,
+	 * 		neighbourhoodDropdownTarget: HTMLElement | undefined,
+	 * 		neighbourhoodTargets: HTMLInputElement[],
+	 * 		neighbourhoodTextTarget: HTMLElement | undefined,
+	 * }} */
+	get typedThis() {
+		// @ts-ignore
+		return this;
+	}
 	connect() {
 		this.updateLabels();
 	}
 
 	submitCategory() {
 		this.updateLabels();
-		this.hideDropdown(this.categoryDropdownTarget);
+		this.toggleDropdownHidden(this.typedThis.categoryDropdownTarget, true);
 		this.submitForm();
 	}
 
 	submitNeighbourhood() {
 		this.updateLabels();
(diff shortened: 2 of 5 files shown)


kimadactyl (commented):
Hey! The core bug fix here is solid — the missing Stimulus target on the filter checkboxes and the null guards are clearly needed. But I'd like to split out the infrastructure changes before merging.

## The fix itself ✅

The `filter.rb` changes adding `data-{controller}-target` to the checkbox inputs, plus the `if (this.selectedCategory)` guards — these make sense and should fix the broken reset buttons.

## Stimulus in package.json ❌

Stimulus is already provided by the `stimulus-rails` gem and pinned via importmap:

```ruby
# Gemfile
gem 'stimulus-rails'

# config/importmap.rb
pin '@hotwired/stimulus', to: 'stimulus.min.js'
pin '@hotwired/stimulus-loading', to: 'stimulus-loading.js'
```

Adding `@hotwired/stimulus` to `package.json` creates a second copy in `node_modules/` that nothing uses at runtime — the browser loads the gem version via importmap. It's only there so the tsconfig type checker can resolve the import, which is the tail wagging the dog.

## tsconfig.json + `typedThis` pattern ❌

This is a significant paradigm shift bundled into a bug fix:

- **`tsconfig.json` with `checkJs: true, strict: true`** turns on TypeScript's type checker across *all* JS files project-wide — that's a huge scope change
- **The `get typedThis()` indirection** means every target access becomes `this.typedThis.formTarget` instead of `this.formTarget`, which every contributor now needs to understand
- **It fights Stimulus's design** — Stimulus is intentionally dynamic with its target properties. The idiomatic way to handle optional targets is `if (this.hasCategoryDropdownTarget)`, which the old code already used

We don't want to introduce "TypeScript by stealth" — if we decide to add type checking to our JS it should be a deliberate architecture decision with its own discussion, not something that comes in with a bug fix.

## Suggestion

Could you split this into two?

1. **Merge now**: the `filter.rb` target data attribute fix + null guards using idiomatic Stimulus patterns (`if (this.hasXTarget)` / `if (this.selectedCategory)`)
2. **Separate discussion**: the tsconfig/typedThis/package.json typing infrastructure — happy to discuss this as its own proposal

idkidk000:
You already OK'd me adding a [jt]sconfig.json, a `stimulus-rails` `devDependency`, and adding jsdoc types for the password input view toggle pr. As I said at the time: this just allows the IDE to perform basic type checking. It makes no runtime difference and needs no additional build step

There is no typescript. You can tell by the lack of `.ts` files, lack of `typescript` dependency, and lack of typescript compiler in the build process

Your llm is right that I'm "fighting" stimulus’s design (i.e. doing it properly instead), and that's because "check thing a to determine the validity of thing b" is quite literally an antipattern. But "thing a" wasn't even being checked consistently, hence the bug. Stimulus is completely untyped so you will get nowhere trying to work with it. My pattern corrects the behaviours so your IDE will now actually tell you if you're doing something silly. You're welcome

You can of course still write bad JS if you like - it won't fail any checks since we have no js linter

`stimulus-rails` should have been in `devDependencies` rather than `dependencies`, and I can fix that. I disagree with everything else

kimadactyl:
happy to say javascript is not my forte. the principle im defending here though is having one consistent style through the whole app. the biggest issue we have as an open source project is every new developer has their preferred way of working that they think is "correct". the project as a whole though needs to only have one style. all the messes we have got into before have been from this: just different people working in parallel using subtly different styles that make the overall thing a mess. 

> There is no typescript. You can tell by the lack of .ts files, lack of typescript dependency, and lack of typescript compiler in the build process

so yes i get this - im saying stimulus is untyped. we dont have a type checking system in js rn. this PR looks a lot to me like trying to squeeze static types into js in a bug fix - im not against the principle of thinking about how we can add it into the codebase, its just not the place to introduce it.

> You can of course still write bad JS if you like
> I'm "fighting" stimulus’s design (i.e. doing it properly instead)

please be careful with making value judgements like this - people have different preferences, things have different affordances, its pros and cons. for better or worse, rails + stimulus is dynamically typed. we've been adding in better hinting as you've seen but all i'm saying is the place to start changing code base idioms is not in a bug fix. the rails author is a shitlord but theres a ton written about this and big opinion on both sides.

i hope that makes sense?

Issue #2879, opened 2026-01-15
Display abbreviated neighbourhood names in public frontend


kimadactyl:
## Description

Reinstate the display of abbreviated neighbourhood names (`name_abbr`) in the public frontend where space is limited.

## Context

The `name_abbr` field exists on neighbourhoods and can be edited in the admin interface, but it's not currently being used in the public-facing site. This field was intended to provide shorter versions of neighbourhood names for display in compact UI elements.

## Suggested Implementation

Use `name_abbr` (falling back to `name`) in places where:
- Space is constrained (mobile views, badges, compact lists)
- Multiple neighbourhood names appear together
- Breadcrumbs or navigation elements

## Related

- #2393 - Bug about abbreviated names showing in admin ward pickers (separate issue)

kimadactyl:
Superseded by #3240, which rethinks neighbourhood name display holistically (condensed names everywhere, admin vs. public context rule, manual vs. algorithmic abbreviation).

Issue #2380, opened 2024-04-10
[Bug]: Counties do not display when choosing Neighbourhood information 


honor-gfsc:
## Description

When choosing a Neighbourhood either for a neighbourhood_admin or a site, counties are missing from the list of options. This means one has to manually add each District within a County to effectively grant the same permissions. 

## What you expected to happen

~~- [ ] Counties (or all neighbourhoods?) should be choosable from the list of site neighbourhoods~~
~~- [ ] Counties (or all neighbourhoods?) should be choosable from the list of user neighbourhoods~~

- [ ] Admin dropdowns show name not abbreviated name fields
- [ ] Improve neighbourhood admin text if this isn't good enough as is

aaaaargZombies:
> As noted at standup today this is hopefully an `e` fix and is just missing options from the dropdown.
> 
> If not then this is prob an `eee`. Boo!

oh boy - wish the reasons for this had been added to the ticket  - wish me luck

aaaaargZombies:
Struggling to reproduce this bug as a `root`

![2024-04-12-11:51:39-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/eab18006-44e1-4f0d-9fde-fa6285c7f664)
![2024-04-12-11:51:56-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/6937492d-18b9-4ab6-a808-1ac4a10d3836)
![2024-04-12-11:52:04-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/7f4666bd-cad4-4f1a-b0b8-361639a7cb44)

Will move on to other types of admin - Good case for adding the "how to reproduce" section to a bug report.

aaaaargZombies:
Created a `partnership_admin` and `site_admin` for tran dimension manchester, `neighbourhood_admin` for england

- only `root` users can add new neighborhoods to sites I believe so not applicable.
- again I don't have permission to assign neighbourhoods to users when not a `root`.

I can still access counties when setting the service area for partners.

![2024-04-12-12:59:14-screenshot](https://github.com/geeksforsocialchange/PlaceCal/assets/6824891/93eba070-8d5e-4242-9379-8592e4bcdfd1)

katjam:
[...]

PR #3194, opened 2026-05-28
Fix #2393: show full neighbourhood names in admin ward pickers

### LINKED ISSUE #2393: [Bug]: abbriviated neighbourhood names can make it hard to find a neighbourhood
## Description

We use to the abbreviated name in the select2 dialog for adding neighbourhoods to other enterties

This lead some people to think counties were missing from our neighbourhood dataset see #2380

We may need to think of a way to include both the full name and abbreviation or remove the option to abbreviate if it going to muddy the waters.

## Steps to reproduce

1. What you did first
1. What you did next

## Platform (if relevant)

kimadactyl:
## What & why

Admin neighbourhood/ward pickers were showing the **abbreviated** name (`name_abbr`) instead of the full ward name. This caused confusion — e.g. "Greater Manchester" appeared as "Gtr Mcr" and some users thought counties were missing from the dataset (see #2380).

Per the issue discussion, the fix is to use the **full name in the admin interface** and keep abbreviations on the public frontend (that redesign is tracked separately in #2879).

## Changes

- Add `Neighbourhood#contextual_fullname`, mirroring the existing `#contextual_name` but using `#fullname` (full `name`) instead of `#shortname` (`name_abbr`).
- Switch the three admin picker helpers to it:
  - `PartnersHelper#options_for_service_area_neighbourhoods`
  - `SitesHelper#options_for_sites_neighbourhoods`
  - `UsersHelper#options_for_user_neighbourhoods`
- Public-facing code (`partner_card.rb`, datatables, hierarchy badges) is **deliberately untouched** — it keeps using `shortname`/`contextual_name`.

## Tests

- Model spec for `#contextual_fullname` (full name vs abbreviation, parent and no-parent branches).
- Helper spec for `UsersHelper#options_for_user_neighbourhoods` asserting the rendered picker label uses the full name and not the abbreviation. Verified it fails before the change and passes after.

Closes #2393

### DIFF
--- a/app/helpers/partners_helper.rb
+++ b/app/helpers/partners_helper.rb
@@ -19,7 +19,7 @@ module PartnersHelper
       .order(:name)
       .all
       .collect do |ward|
-        name = ward.contextual_name
+        name = ward.contextual_fullname
         name += " - #{ward.release_date.year}/#{ward.release_date.month}" if ward.legacy_neighbourhood?
 
         { name: name, id: ward.id }
--- a/app/helpers/sites_helper.rb
+++ b/app/helpers/sites_helper.rb
@@ -10,7 +10,7 @@ module SitesHelper
       .order(:name)
       .all
       .collect do |ward|
-        name = ward.contextual_name
+        name = ward.contextual_fullname
         name += " - #{ward.release_date.year}/#{ward.release_date.month}" if ward.legacy_neighbourhood?
 
         { name: name, id: ward.id }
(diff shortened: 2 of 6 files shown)


ivanELEC (commented):
Can't see any implementation issues but I am wondering whether these options_for... methods reference old code that isn't currently in use, I can't see the helper functions referenced anywhere unless i'm missing some framework magic or usage of send that calls them via a variable name that holds a string/symbol.

ivanELEC:
> Can't see any implementation issues but I am wondering whether these options_for... methods reference old code that isn't currently in use, I can't see the helper functions referenced anywhere unless i'm missing some framework magic or usage of send that calls them via a variable name that holds a string/symbol.

TLDR should these methods be removed instead of enhanced?

PR #3198, opened 2026-05-28
Fix #1433: Validate partner opening times

### LINKED ISSUE #1433: Validate opening times
## User story

As a PlaceCal admin I need be warned or stopped adding impossible opening times so that members of the public who want to visit me don't get confused. 

## Acceptance criteria

- [ ] Prevents user adding overlapping opening times
- [ ] End should not begin before start

I'm not sure if this is best done via the model in rails or the stimulus controller in the page.

## Implementation plan

kimadactyl:
## Summary

Adds server-side validation on the `Partner` model so admins cannot save impossible opening times, per the acceptance criteria in #1433:

- **End must not be before start** — rejects any specification whose closing time is not strictly after its opening time (also catches equal open/close).
- **No overlapping times** — rejects ranges that overlap on the same day of the week.

Validation lives on the model (the issue asked "model or Stimulus" — the model is authoritative since data can be set via API/import as well as the form).

## Implementation notes

- Opening times are stored as a JSON string of schema.org `OpeningHoursSpecification` objects (`dayOfWeek`, `opens`, `closes`).
- Times are parsed defensively into minutes-since-midnight via a regex, so both `"HH:MM"` and `"HH:MM:SS"` are handled. Blank/malformed entries, non-array JSON, and missing keys are skipped without error (JSON validity is already covered by the existing `opening_times_is_json_or_nil` check).
- Adjacent ranges that touch at a boundary (e.g. 09:00–12:00 and 12:00–17:00) are treated as valid, non-overlapping.
- Validation messages added to `config/locales/en.yml` under `activerecord.errors.models.partner.attributes.opening_times` (i18n, no hardcoded strings).

## Testing

Added model specs covering: end-before-start invalid, close-equals-open invalid, same-day overlap invalid, and valid non-overlapping cases (same day and across days). The three "invalid" specs fail before the change and pass after. Full `spec/models/partner_spec.rb` (59 examples) passes; RuboCop clean.

Closes #1433


app/models/partner.rb
```diff
@@ -488,6 +489,67 @@ def opening_times_is_json_or_nil
...
+  # that ranges on the same day do not overlap. Skips when the value is blank
+  # or not parseable (handled by #opening_times_is_json_or_nil).
+  def opening_times_are_valid
+    return if opening_times.blank?
+    return unless valid_json? opening_times
+
+    specs = JSON.parse(opening_times)
+    return unless specs.is_a?(Array)
+
+    by_day = Hash.new { |hash, key| hash[key] = [] }
```
ivanELEC: Not sure the need for such a verbose line, could just be

by_day = {} or by_day = Hash.new

app/models/partner.rb
```diff
@@ -488,6 +489,67 @@ def opening_times_is_json_or_nil
...
+
+    specs.each do |spec|
+      next unless spec.is_a?(Hash)
+
+      opens = parse_opening_time(spec['opens'])
+      closes = parse_opening_time(spec['closes'])
+      next if opens.nil? || closes.nil?
+
+      if closes <= opens
+        errors.add :opening_times, :end_before_start
```
ivanELEC: More of a nitpick than a showstopper, this adds multiple instances of the same error to the errors array. This doesn't have an effect on the UI as that correctly only shows the error message once, but could be good to protect against this if there could be other knock on effects e.g with 
adding ` if errors.filter { |error| error.type == :end_before_start }.blank?` to this line

ivanELEC (approved):
Approved with comments. Works but could be slightly cleaner

app/models/partner.rb
kimadactyl: lol god nice catch

kimadactyl: Simplified to `by_day = {}` with `||=` at the append site.

app/models/partner.rb
kimadactyl: Fixed — errors are now collected as booleans and added once after the loop. Also fixed the same issue in `flag_overlapping_opening_times`.

Issue #3156, opened 2026-05-06
Implement new join.placecal.org site


kimadactyl:
## Use case

The stuff about joining should be spun out elsewhere so our homepage can focus on the community stuff

Draft design: https://c34c044f-2cc7-468c-9d78-b719c3e49cf4.claudeusercontent.com/v1/design/projects/c34c044f-2cc7-468c-9d78-b719c3e49cf4/serve/Join.html?t=9987aabc16e51c43230b5a9623e6b79ec24a66dd595c89503ea3b91187689ec4.89df25d4-08c9-4695-8409-afdbab869255.3d7bbb02-0714-4be8-b331-6271cc1253dc.1778091974

## Acceptance criteria

- [ ] criteria one
- [ ] criteria two

## Implementation plan

_To be written by the developer_

PR #3201, opened 2026-05-28
Fix #2463: Hide homepage-only URLs (e.g. /find-placecal) from site subdomains

### LINKED ISSUE #2463: [Bug]: Visiting /find-placecal on a subdomain has bad styling
## Description
e.g. https://exeter.placecal-staging.org/find-placecal

It's not explicitly linked but also easy to find. We should check routes on subdomains generally and lock them down. See also #2324 

image

kimadactyl:
## What

Homepage-only pages such as \`/find-placecal\` were reachable on any site subdomain (e.g. \`exeter.placecal-staging.org/find-placecal\`), where they don't belong. As noted in the issue, this let main-homepage URLs be "spoofed" as part of areas/partnerships they aren't part of — the routing didn't match where these pages legitimately exist.

## How

Added a \`before_action\` guard in \`PagesController\` scoped to exactly the homepage-only actions:

\`\`\`
find_placecal, our_story, community_groups, metropolitan_areas,
vcses, housing_providers, social_prescribers, culture_tourism
\`\`\`

On a non-default site (i.e. \`current_site\` present and not the default site) the guard redirects to the site root; on the default site these pages render as before. The guard reuses the existing \`default_site?\` helper and mirrors the naming of the existing \`SitemapsController#require_default_site\` convention (that one returns 404; here we redirect, since these requests land on a real sub-site with its own root).

Actions deliberately **not** guarded:
- \`home\` — serves both the default directory home and per-site homepages
- \`privacy\`, \`terms_of_use\` — render \`Views::Directory::*\`, valid per-site
- \`robots\` — per-site robots.txt

## Tests

Added request specs in \`spec/requests/public/pages_spec.rb\`: on a non-default host every homepage-only path redirects to the site root; on the default host \`/find-placecal\` still returns 200 (existing test retained).

Verified before/after: the 8 new non-default-site redirect specs fail without the guard and pass with it. (Other pre-existing failures in that file are unrelated environmental asset-build errors — missing \`public_tailwind\`/\`application.css\` — affecting all render-path tests regardless of this change.)

Closes #2463

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
 class PagesController < ApplicationController
+  # Actions that belong only to the main placecal.org homepage. They must not be
+  # reachable on a sub-site subdomain, where they don't belong (see #2463).
+  HOMEPAGE_ONLY_ACTIONS = %i[
+    find_placecal our_story community_groups metropolitan_areas
+    vcses housing_providers social_prescribers culture_tourism
+  ].freeze
+
   before_action :set_primary_neighbourhood, only: [:site]
   before_action :set_site
+  before_action :require_default_site, only: HOMEPAGE_ONLY_ACTIONS
 
   def home
     if default_site?
@@ -74,6 +82,17 @@ class PagesController < ApplicationController
   DIRECTORY_CACHE_TTL = 1.day
 
   private
+
+  # Homepage-only actions are not valid on sub-sites. When such an action is
+  # requested on a real, non-default site, redirect to that site's root rather
+  # than letting the page render where it doesn't belong (see #2463).
+  # (SitemapsController#require_default_site does the same for its routes but
+  # returns 404; here we redirect because these pages have a sub-site equivalent.)
+  def require_default_site
+    return if current_site.nil? || default_site?
+
+    redirect_to root_path # site root on the current (sub-site) host
+  end
 
   def render_directory_home
     @stats = Rails.cache.fetch('directory/stats', expires_in: DIRECTORY_CACHE_TTL) do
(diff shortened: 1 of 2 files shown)


kimadactyl:
Most of these are going to be moved to join.placecal.org (see #3156). Can we do that now or do we need to wait? I think most are now not linked from the homepage. It feels like this should be in a subsite block like the admin routes?

kimadactyl:
Two things bundled here — happy to go either way:

**The guard mechanism (your "subsite block like admin routes" idea):** agreed that's cleaner in principle. Admin uses `constraints: { subdomain: 'admin' }`, but these homepage-only pages can't use a plain subdomain constraint — "is this the directory site?" isn't a fixed subdomain (sub-sites are arbitrary subdomains; the default site is resolved via `Site.find_by_request`, see `app/constraints/sites.rb`). So the routes-level version would be a small custom constraint, e.g. `constraints ->(req) { Site.find_by_request(req)&.default_site? }` wrapping these routes — definitely doable, just a bit more than the admin one-liner.

**Timing vs #3156:** since these are moving to join.placecal.org anyway, this PR is really a stopgap. The `before_action` I used is deliberately low-risk and a one-line delete once #3156 lands. So:
- If you want the cleaner version **now**, say the word and I'll convert it to the routes-constraint approach above.
- If #3156 is imminent, it may be best to **hold** and fold this into that work rather than churn routing twice.

I didn't rewrite the routing speculatively since it's entangled with your #3156 roadmap call. Your steer?

kimadactyl:
(For when this is picked up in the join refactor #3156.) Recommended approach: a **routes-level constraint** rather than the controller `before_action` in this PR — analogous to admin's `constraints: { subdomain: 'admin' }`, but since "directory site" isn't a fixed subdomain, wrap the homepage-only routes in `constraints ->(req) { Site.find_by_request(req)&.default_site? }` (reusing `app/constraints/sites.rb`). The homepage-only action list is in this branch's `HOMEPAGE_ONLY_ACTIONS` constant; branch `worktree-agent-a64f15082b9893e78` can be cherry-picked.

PR #3204, opened 2026-05-28
Fix #3123: backfill city on addresses from postcodes.io

### LINKED ISSUE #3123: Backfill city field on existing addresses from postcodes.io
## Problem

The `city` field on the Address model is not populated for existing addresses. This means `addressLocality` is missing from JSON-LD structured data, causing a non-critical warning in Google Rich Results Test.

As of #3122, new addresses will have `city` auto-populated from postcodes.io `admin_district` during geocoding. But existing addresses need backfilling.

## Proposed fix

Write a migration or rake task that:
1. Iterates over addresses where `city` is NULL
2. Looks up the postcode via postcodes.io
3. Sets `city` from the `admin_district` field in the response
4. Respects postcodes.io rate limits

## Related

- #3121 — JSON-LD structured data fixes
- #3122 — PR implementing the fix for new addresses

kimadactyl:
Closes #3123

Follow-up to #3122 (which populates `city` for new addresses during geocoding): backfills the `city` field on pre-existing addresses where it is NULL, so `addressLocality` appears in JSON-LD.

- `Address#backfill_city!` — looks up the postcode via the existing `Geocoder` path and writes `admin_district` to `city` via `update_columns` (no re-geocode/validation), returning a status symbol.
- `Address.needs_city_backfill` scope — NULL-city rows that have a postcode.
- `addresses:backfill_city` rake task — iterates the scope with a polite `POSTCODES_IO_SLEEP` pause; prints a summary.
- Specs stub postcodes.io via WebMock (never hits the live API); the `around` Geocoder reconfigure is wrapped in `ensure` for test isolation.

Deliverable is the task + specs; the actual prod backfill is run separately by a maintainer.


kimadactyl:
Is it the right approach to bake this into the model forever vs doing it as a one-off task? Has this been fixed on prod already?

kimadactyl:
Good call — agreed it shouldn't live on the model forever. Pushed `adb859d9`:

- Removed `Address#backfill_city!` and the `needs_city_backfill` scope from the model.
- Inlined the postcode lookup directly in the `addresses:backfill_city` rake task, matching the sibling `addresses:regeocode` task in the same file (which does the same inline `Geocoder.search` + `update_columns`).
- Swapped the model specs for a rake-task spec that drives the task with postcodes.io stubbed via WebMock.

On **"has this been fixed on prod already?"** — I can't see prod state from here, so worth a quick check before running. But the task is **idempotent**: it only touches rows where `city IS NULL` (and skips when postcodes.io returns no `admin_district`), so running it is safe regardless of current state. #3122 only auto-fills *new* addresses during geocoding, so any pre-#3122 rows would still need this one-off pass.

PR #3206, opened 2026-05-28
Fix #2567: set default site when current_site is nil

### LINKED ISSUE #2567: [Bug]: current_site sometimes undefined

kimadactyl:
## What

`current_site` is deliberately `nil` on the admin subdomain (it provides a global, site-less view; see `app/controllers/application_controller.rb`). Public resource pages such as `/events` and `/events/:id` are **not** behind a subdomain constraint, so crawlers (e.g. Bytespider) reach them at `admin.placecal.org`. There the `nil` site flowed into the non-nilable `Site`-typed Phlex view props (`Views::Events::Index`, `Views::Events::Show`) and raised `Literal::TypeError` — the Rollbar error in this issue.

## Fix

`set_site` now falls back to the default site when `current_site` is `nil` and no redirect is already in flight:

```ruby
def set_site
  @site = current_site
  return if @site || response.redirect?

  @site = Site.find_by(slug: 'default-site') || Site.new
end
```

This is the single shared seam used by all site-scoped public controllers (events, partners, news, partnerships, collections, pages), so it fixes the root nil-deref uniformly rather than guarding each view.

- **Site present (normal traffic):** unchanged — `current_site` returns the site and we return early.
- **Unknown subdomain:** unchanged — `current_site` already issues a redirect to the base domain; we detect `response.redirect?` and leave `@site` nil so that redirect halts the request.
- **Admin subdomain / no resolvable site:** previously crashed; now renders the default (directory) site view.

`current_site` itself is untouched, preserving its documented "nil on admin" contract used by `default_site?`, navigation, etc. The admin interface and Devise views do not use `@site`, so they are unaffected (verified by the existing devise request specs).

## Tests

Added request specs in `spec/requests/public/events_spec.rb` for the no-current-site (admin subdomain) path on both the events index and an event show page. These fail with `Literal::TypeError` before the change and pass after. Existing public-controller and devise request specs continue to pass.

Closes #2567

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -192,8 +192,21 @@ class ApplicationController < ActionController::Base
     @calendar = Calendar.find(params[:id])
   end
 
+  # Assigns @site for site-scoped public controllers (events, partners, etc.).
+  #
+  # current_site is deliberately nil on the admin subdomain (it provides a
+  # global, site-less view). Public resource pages such as /events/:id are not
+  # behind a subdomain constraint, so crawlers can reach them on the admin
+  # subdomain where current_site is nil — which previously blew up the
+  # Site-typed Phlex view props (issue #2567). Fall back to the default site so
+  # those pages render the directory view instead of raising. When current_site
+  # has already issued a redirect (e.g. an unknown subdomain) we leave @site nil
+  # and let that redirect halt the request.
   def set_site
     @site = current_site
+    return if @site || response.redirect?
+
+    @site = Site.find_by(slug: 'default-site') || Site.new
   end
 
   def redirect_from_default_site
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -198,15 +198,15 @@ class ApplicationController < ActionController::Base
   # global, site-less view). Public resource pages such as /events/:id are not
   # behind a subdomain constraint, so crawlers can reach them on the admin
   # subdomain where current_site is nil — which previously blew up the
-  # Site-typed Phlex view props (issue #2567). Fall back to the default site so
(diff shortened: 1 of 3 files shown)


kimadactyl:
is Site.new a meaningful think to pass? or is it acutally just fine for it to be nil?

i think also we should be referring to this as "directory site" now if thats what we mean - or have the default set elsewhere.

kimadactyl:
All three points well taken — pushed `fc9b4e67`:

- **`Site.new` vs `nil`:** you're right that `Site.new` isn't meaningful (a blank, nameless, unsaved site). But `nil` isn't safe *here* either — that was the actual #2567 crash: the Phlex view props on `Views::Events::Index/Show` are non-nilable `Site`. (Interesting wrinkle: the **query** layer already treats `@site.nil?` as the directory site — see `EventsQuery#base_scope` and `PartnersQuery`. So nil is fine for data, just not for the typed view props.) So the meaningful fallback is the real **directory site**, which is what it now passes.
- **"directory site" naming + set elsewhere:** added a named `Site.directory` accessor (in the existing `class << self` block) and use it from `set_site`, replacing the inline `Site.find_by(slug: 'default-site')`. Reuses the existing `directory_site?` terminology and gives us one place for the lookup.

Alternative if you'd rather lean into "nil is fine": make the `Views::Events::*` Site props nilable and let `nil` flow through (the queries already cope). That's a bigger change touching the view layer — happy to switch to that direction if you prefer it over a fallback. Left as the contained fix for now.

PR #3208, opened 2026-05-28
Fix #2991: load NormalIsland module in geocoder; link skip-admin checkbox to its label

### LINKED ISSUE #2991: [Bug]: Problems adding/editing Partners using seeded dev db
## Description

Trying to add a new partner using seed address data. I tried with a postcode copied from an existing seed partner and got an "Uninitialized constant NormalIsland" error:

Image

Using a different postcode which matches the regex at `lib/normal_island/geocoder_lookup.rb#L33` gives the same error.

A real postcode give some form validation errors:
- Postcode has been found but could not be mapped to a neighbourhood at this time (probably expected)
- Address is invalid

Image

(also, the `Partner Name` field always shows a field validation error that the name isn't long enough. editing the field resolves)

Trying to edit an existing Partner (doesn't matter what field you edit - in this I edited the name) gives an "Address is invalid" form validation error

Image

Also, if i'm being pedantic, this checkbox and label are unlinked so I can't toggle the checkbox by clicking the label. But more importantly, it's bad for screenreader support. The checkbox needs an `id="somevalue"` and the label needs an `for="somevalue"`

Image

Image

kimadactyl:
## Summary

Fixes two defects from #2991 hit when adding/editing partners against the seeded dev DB.

### (a) `uninitialized constant NormalIsland`
`config/initializers/geocoder.rb` required the custom `Geocoder::Lookup::NormalIsland` lookup but never loaded the `NormalIsland` **data** module (`POSTCODES`, `WARDS`, `ADDRESSES`, …) that the lookup references. That module is only `require_relative`'d by the seed scripts (and, in the test suite, by factories), so it was undefined in the running dev/seed app. Geocoding a `ZZ`-prefix postcode during a partner save then raised `uninitialized constant NormalIsland`.

Fix: `require 'normal_island'` in the initializer (before the lookup) so the constant resolves at app boot. As a bonus this also loads the `UKPostcode` Normal Island monkey-patch in dev/test app boot, not just during seeding. The require is idempotent and the patch is prepended exactly once.

### (b) Unlinked checkbox + label (accessibility)
In the new-partner wizard's "Invite a Partner Admin" step (`app/views/admin/partners/new.rb`), the "Skip this step — I'll add admins later" checkbox had no `id` and its label no `for`, so clicking the label did nothing and screen readers couldn't associate them. Added `id="partner_skip_admin"` to the input and a matching `for` on the label.

## Tests
- `spec/lib/normal_island/geocoder_lookup_spec.rb` — asserts the geocoder initializer loads `normal_island` (regression guard, since factories otherwise mask the constant in-suite) and that a Normal Island postcode lookup resolves without `NameError` and returns the seeded result.
- `spec/components/admin/partners/new_invite_step_spec.rb` — renders the wizard's invite step in isolation and asserts the checkbox has an `id` and its label a matching `for`.

Both specs were confirmed to fail before the change and pass after. Targeted specs green; rubocop clean on changed files.

## Notes / out of scope
- I did **not** chase the "Address is invalid" / "postcode could not be mapped to a neighbourhood" validation messages, nor the "Partner Name too short" transient error — per the issue these may be expected; they need separate investigation.
- The "Skip this step…" label text is a hardcoded string (an i18n violation per project rules) but it is pre-existing and unrelated to either defect, so I left it untouched to keep this change focused.

Closes #2991

### DIFF
--- a/app/views/admin/partners/new.rb
+++ b/app/views/admin/partners/new.rb
@@ -325,10 +325,10 @@ class Views::Admin::Partners::New < Views::Admin::Base
                            "Optionally invite someone to manage this partner. They'll receive an email invitation.")
         div(class: 'space-y-6') do
           div(class: 'flex items-center gap-3 p-4 bg-base-200/50 rounded-xl') do
-            input(type: 'checkbox', class: 'checkbox checkbox-sm',
+            input(type: 'checkbox', id: 'partner_skip_admin', class: 'checkbox checkbox-sm',
                   data: { 'partner-wizard-target': 'skipAdminCheckbox',
                           action: 'change->partner-wizard#toggleAdminFields' })
-            label(class: 'text-sm') { "Skip this step - I'll add admins later" }
+            label(for: 'partner_skip_admin', class: 'text-sm') { "Skip this step - I'll add admins later" }
           end
           div(data: { 'partner-wizard-target': 'adminFields' }) do
             render_admin_details_card(form)
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 if Rails.env.local?
+  # Load the NormalIsland data module (POSTCODES, WARDS, etc.) referenced by
+  # the custom geocoder lookup. Without this the constant is undefined outside
+  # the seed scripts, raising "uninitialized constant NormalIsland" when the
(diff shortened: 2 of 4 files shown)


kimadactyl:
Superseded by #3212 — identical fix on a clean, independent worktree branch. #3208 was accidentally opened from the session branch via a stale upstream tracking ref.

PR #3212, opened 2026-05-28
Fix #2991: load NormalIsland in geocoder + link skip-admin checkbox

### LINKED ISSUE #2991: [Bug]: Problems adding/editing Partners using seeded dev db

kimadactyl:
Closes #2991

Two defects when adding/editing partners on a seeded dev DB:

1. **`uninitialized constant NormalIsland`** — `config/initializers/geocoder.rb` required the custom `NormalIsland` lookup but never loaded the `NormalIsland` data module it references (only the seed scripts/factories did), so geocoding a `ZZ`-prefix postcode raised at runtime. Added `require 'normal_island'` in the initializer.
2. **Unlinked checkbox/label** — the new-partner "Skip this step" checkbox had no `id` and its label no `for`; added matching `id`/`for` for accessibility.

Specs: a guard that the initializer loads the module + a behavioral lookup; a component spec asserting the checkbox `id`/label `for` match. Out of scope (noted): the "Address is invalid" / unmapped-neighbourhood messages.


ivanELEC (approved):
LGTM - was able to create a new partner without error (with a valid address) and accessibility issues look to be fixed

PR #3357, opened 2026-07-23
Fix partner wizard failing on blank service area rows

### LINKED ISSUE #3356: [Bug]: Partner creation wizard not progressing properly due to service area issues
## Description

When adding a new partner, all steps appeared complete but I was presented with errors at the final stage. Errors included:
1. "Service areas neighbourhood must exist" - I had added 4 service areas all within Leeds as I usually do.
2. "Service area invalid" - I've used these same service areas for other partners without issue
3. "Partners must have an address or a service area inside your neighbourhood" - again, I did nothing different to usual and all service areas are within Leeds.

## Steps to reproduce

1. What you did first
 Entered the data as usual

2. What you did next
tried to edit the data from step 1. and it threw up an error on every page e.g. "name must be 5 characters long" - but the name entered was "Common" which it accepted again once I re-typed. same with every stage, it showed an error that didn't align but re-typing e.g. the email address worked up to the final step.

I would like to add this partner as usual

## Platform (if relevant)

Placecal - Google chrome, from a desktop (mac)

Image
Image
Image
Image
Image
Image
Image

kimadactyl:
Resolves #3356

## Description

Fixes a family of bugs in the admin "service area" / neighbourhood pickers, found while diagnosing #3356 (reproduced live in the browser as a non-root neighbourhood admin).

### 1. Blank service-area rows break the save (the reported bug)

An untouched "New Service Area" picker row submits a blank `neighbourhood_id`. With no `reject_if`, that built an invalid `ServiceArea` producing three errors at once — and the nil id also failed `check_neighbourhood_access` for non-root admins even when valid service areas were present. Fixed with a `reject_if` that drops blank rows (casting `_destroy`, which the form always submits as `"false"`).

### 2. Failed-save re-render dropped newly-added rows

`NeighbourhoodCard` rendered no hidden `neighbourhood_id`, so after a validation failure a row shown as a card was dropped from the retry submission. Fixed by rendering the hidden field in the card; the same shared card fix also covers the **site** form (secondary neighbourhoods) — verified by a request spec. The sites primary card branch also now preserves `relation_type`.

### 3. Wizard showed stale errors after a failed save

Name min-length / availability / admin-email state only updated on input events, so a re-rendered wizard showed stale errors and a disabled Continue until every field was re-typed. Fixed by re-running the checks on `connect()`.

### 4. Cascading picker dead-ended on single-country data

When a level has one option the picker auto-selects it, then loaded the next level from the parent scope — which is null at the top level, so the endpoint returned roots instead of the country's children and every level below died. The picker was unusable with a single country. Fixed by falling back to the auto-selected id at the root. Also derives neighbourhood `level` from `unit` in the Normal Island factory (the picker filters by `level`, which the real seeds set but the factory left nil) — this is what previously blocked driving the picker in system tests.

### Type of change

### How Has This Been Tested?

- Reproduced the exact three-error failure and the picker dead-end live in Chrome, then confirmed each fix
- Model + request specs: blank rows rejected; added rows survive a failed-save re-render (partners and sites); `_destroy` removal still works
- System specs: full wizard failed-save→recovery run; full auto-select cascade down to the ward dropdown (fails without the JS fix)
- Broad regression run across neighbourhood-using specs (models, queries, requests, graphql, datatables, system) after the global factory `level` change — all green

Standard deploy, no infrastructure or config changes.

### DIFF
--- a/app/components/admin/neighbourhood_card.rb
+++ b/app/components/admin/neighbourhood_card.rb
@@ -11,6 +11,10 @@ class Components::Admin::NeighbourhoodCard < Components::Admin::Base
 
   def view_template
     div(class: 'nested-fields card bg-base-200/50 border border-base-300') do
+      # Without this hidden field, unpersisted rows rendered as cards (e.g. after
+      # a failed save re-renders the form) would submit nothing and be lost (#3356)
+      raw(@form.hidden_field(:neighbourhood_id)) if @form
+
       div(class: 'card-body p-4 gap-3') do
         if @show_header
           h3(class: 'font-semibold flex items-center gap-2') do
--- a/app/javascript/controllers/cascading_neighbourhood_controller.js
+++ b/app/javascript/controllers/cascading_neighbourhood_controller.js
@@ -153,7 +153,13 @@ export default class extends Controller {
 				// descendants of the auto-selected item (e.g., Manchester is a direct
 				// child of North West, not Lancashire, but Lancashire is auto-selected
 				// as the only county).
-				await this.loadLevel(level - 1, parentId);
+				//
+				// At the top level there is no parent scope (parentId is null): the
+				// children endpoint would then return top-level roots instead of the
+				// auto-selected country's children, dead-ending every level below it.
+				// Fall back to the auto-selected id so the cascade can continue — this
+				// is what left the picker unusable whenever there was a single country.
+				await this.loadLevel(level - 1, parentId || data[0].id);
 			} else {
 				// Multiple options: show the dropdown
 				this.showSelect(target);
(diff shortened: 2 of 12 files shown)


kimadactyl:
Update: live-browser verification against the real wizard (as a non-root neighbourhood admin) caught a hole in the first version of this fix — the form submits `_destroy: "false"` on every service-area row via `nested_form_remove_link`, so the original `reject_if`'s `_destroy.blank?` guard never held and blank rows still produced the exact three errors from the issue. The request specs had passed because they omitted the `_destroy` key the real form always sends.

Fixed in 7462f60 by casting the flag with `ActiveRecord::Type::Boolean`, and the specs now submit `_destroy: "false"` like the real form. Re-verified end-to-end in Chrome: four ward service areas + one untouched blank row → partner created with all four wards, resubmitted from a failed-save re-render (which also exercises the `NeighbourhoodCard` hidden-field round-trip and the wizard `connect()` re-validation).

One repro note: this couldn't be reproduced on dev seed data because with a single country the picker auto-selects it and writes its id — only with multiple countries (as on prod: England/Wales/Scotland/NI) does an untouched row stay blank.

Issue #3337, opened 2026-07-15
Creating a new user with a partner attached failed with "partner invalid" on prod


kimadactyl:
## Report

On production: created a new user via the admin invite wizard with **Trinity House** attached as a partner, and it failed with an error along the lines of "partner invalid". Workaround that succeeded: create the user as a plain citizen first, then attach the partner from the user edit page afterwards.

## Investigation so far (2026-07-15)

Could **not** reproduce locally on current code, logged in as root:

- Ran the full invite wizard (personal details → permissions → select partner → Invite User) with a valid partner: user created, partner attached correctly.
- Repeated with a partner record made deliberately **invalid** (failing current Partner validations, simulating legacy prod data): user still created fine and the partner attached. Rails does not validate unchanged existing HABTM records when saving a new user, so "legacy invalid partner blocks user creation" is ruled out as the mechanism on current code.
- The server accepts `user[partner_ids][]` on create for root (`UserPolicy#permitted_attributes_for_create`) and `Admin::UsersController#create` attaches them via `invite!`.
- AppSignal: no error incidents or retained traces for `Admin::UsersController#create`, and log retention (24h) has already rolled past the attempt, so the exact prod error text is unrecoverable.

## Leads for next time it happens

1. **Capture the exact error message** (screenshot ideally) — "partner invalid" vs "Partners is invalid" vs the non-root guard message point at different code paths.
2. Check the Trinity House record in prod console: `Partner.find_by(name: 'Trinity House').tap(&:valid?).errors.full_messages` — and whether anything about it is unusual (service areas, tags, hidden state).
3. Confirm which account/role did the creating. Non-root creators hit `validate_partner_relation` in `Admin::UsersController` ("You must assign a Partner to the User…"), a different path from root.
4. Check which release was deployed at the time of the attempt — the failure may predate a recent admin users redesign deploy.

PR #3359, opened 2026-07-23
Fix log retention and ship logs to AppSignal (production + staging)

kimadactyl:
## Description

Production logs were effectively undiagnosable after the fact. The app logged at `:debug`, so the per-query SQL firehose filled the container's default Docker `json-file` buffer (`max-size=10m`) within **~47 seconds** — by the time anyone looked, the relevant request had already rolled off. This is why the recent partner-wizard incident (#3356) and the earlier user-invite one (#3337) both hit a wall: "log retention has already rolled past."

This makes production logs both leaner and durable:

- **Default the production log level to `:info`** — keeps the request-lifecycle lines that actually matter for diagnosis (`Started POST …`, `Processing by …`, `Parameters: {…}`, `Completed 200 OK`) while dropping the per-query SQL `DEBUG` noise that was ~90% of the volume. Overridable via `RAILS_LOG_LEVEL` when deeper detail is needed temporarily.
- **Broadcast logs to AppSignal** (`Appsignal::Logger`) alongside stdout, using `ActiveSupport::BroadcastLogger` (the Rails 8 / AppSignal-recommended pattern). This gives searchable logs with real retention, independent of the container buffer — the account already has a log source and the push key is already provisioned. Guarded on `APPSIGNAL_PUSH_API_KEY`, so development, test, and any unconfigured environment keep stdout-only logging unchanged.
- **Raise the Kamal `json-file` cap to `100m × 5`** so `docker logs` also keeps a useful window (AppSignal holds the durable copy).

### Type of change

- [x] This change requires a change to the deployment process (Kamal `logging:` block; picked up on next deploy)

### How Has This Been Tested?

Booted the production environment locally (`RAILS_ENV=production RAILS_LOG_TO_STDOUT=true`) and inspected the constructed logger:

- With the push key set: `Rails.logger` is an `ActiveSupport::BroadcastLogger` over `[ActiveSupport::Logger (stdout), Appsignal::Logger]`, at level `INFO`, with `request_id` tags still rendering and `DEBUG` filtered.
- Without the key: stdout-only (`[ActiveSupport::Logger]`), no AppSignal sink.
- `RAILS_LOG_LEVEL=warn` override respected (level → `WARN`).
- Dev/test are untouched — the whole block is gated on `RAILS_LOG_TO_STDOUT`, which they don't set.

Standard deploy. The `logging:` block applies to all containers (web + job_worker) on the next Kamal deploy. `APPSIGNAL_PUSH_API_KEY` is already in the deploy secrets, so the AppSignal broadcast activates automatically; no new config needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,6 +8,10 @@ This file configures AI coding assistants for PlaceCal. Review the context file 
 - **When the user says something is wrong, believe them** - Trust the user over your assumptions.
 - **Debug properly** - If a fix doesn't work, investigate why rather than assuming it did.
 
+## Environment Config
+
+Staging runs its own `config/environments/staging.rb`, a near-duplicate of `production.rb`. Changes to production env config (logging, caching, mailers, etc.) must be mirrored in `staging.rb` or they won't reach staging.
+
 ## Development URLs
 
 The admin interface uses a subdomain. Access it at:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,7 +10,7 @@ This file configures AI coding assistants for PlaceCal. Review the context file 
 
 ## Environment Config
 
-Staging runs its own `config/environments/staging.rb`, a near-duplicate of `production.rb`. Changes to production env config (logging, caching, mailers, etc.) must be mirrored in `staging.rb` or they won't reach staging.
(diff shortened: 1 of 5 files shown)


kimadactyl:
Updated to also cover staging: `config/environments/staging.rb` is a separate near-duplicate of production and had the same `:debug` level + stdout-only logger, so the fix is mirrored there. Verified staging boots to a `BroadcastLogger` over [stdout, AppSignal] at INFO. Added a note to AGENTS.md that staging.rb must be kept in sync with production.rb.

PR #3394, opened 2026-08-07
Allow manual dispatch of the Discord release notification

kimadactyl:
Follow-up to #3392. Adds a `workflow_dispatch` trigger (optional `tag` input, defaults to the latest published release) to the Discord release workflow, so the announcement can be re-run by hand from the Actions tab, or triggered as a one-off test, without waiting for a real release. On manual runs the release is fetched from the GitHub API.

Lets us fire a test post now that the `DISCORD_RELEASE_WEBHOOK` secret is set, and gives a permanent "re-announce a release" button.

## Verification

YAML parses, embedded Python compiles, and the manual-dispatch path (API fetch then Discord post, both mocked) builds the same green embed with the maintenance section collapsed.


kimadactyl:
Closing as overkill per discussion. The manual-test need is covered by a local command, and the real fix (missing User-Agent header) will go in a focused PR.

Issue #3368, opened 2026-07-25
Retire the standalone Trans Dimension site and serve it as an extension-themed PlaceCal Site (PRD)


kimadactyl:
**This is a PRD-style epic.** It is written to be fed to AI coding sessions as a working brief, work package by work package, with as little human attention as possible. Section 9 is the agent operating model and is not optional reading for whoever runs it.

Mapped against `main` at 5f92cb55 on 2026-09-02. Implementers must verify against the code; paths below were checked on that commit.

## 1. Why we're doing this

The Trans Dimension (transdimension.uk) is a ~9.7k line elm-pages static site that consumes PlaceCal's GraphQL API, rebuilt hourly by a cron so that data is at most an hour stale. Maintaining a second frontend, a second design system, a second deploy pipeline and a keepalive cron for one partnership is a drag on a small team, and it gives TD none of the things PlaceCal already has (live data, iCal, CSV, maps, tests, monitoring).

The fix is not to move TD's code into PlaceCal. It is to give PlaceCal one generic capability it should have had anyway: **extensions**. An extension is a Rails engine that can register a theme (stylesheet, homepage view, map style, fonts, copy) and content for a Site, without core knowing the extension exists. TD becomes the first extension. It also sets up the model GFSC needs next: a public core that anyone can self-host, plus GFSC-specific things (the join site, a future billing portal, client themes) living outside core as extensions.

**The goal, stated once.** The new transdimension.uk matches the current one exactly, page for page, at every viewport, except for the deliberate deviations listed in section 4a. Anything else that differs is a bug. This is the acceptance criterion for the whole epic and the reason cheap agents can do most of the work: "matches the screenshot" is a decidable gate, "looks good" is not.

**The one rule that decides where code goes:** if a second partnership could want it, it is a core feature and gets built generically. If only this site wants it, it is views, CSS, assets and copy in an extension. Extensions contain no models, no migrations, no business logic.

## 2. Repos and boundaries (the target shape)

| Repo | Visibility | Contains | Exists today? |
|---|---|---|---|
| `geeksforsocialchange/PlaceCal` | public | the deployable app, the extension mechanism, every generic feature | yes |
| `geeksforsocialchange/placecal-theme-transdimension` | public, Hippocratic licence, assets © Gendered Intelligence | the TD extension: theme, homepage, static page content, copy, fonts, illustrations | no, created in Phase 2 |
[...]

PR #3437, opened 2026-09-02
WP 0.4: engine loading proof (#3368)

kimadactyl:
WP 0.4 of #3368. Throwaway engine under spec/fixtures/extensions/example_theme, required from rails_helper only.

Proven, with no core config edits:
- Engine autoloads ExampleTheme::Views::Home and ExampleTheme::Components::Head via Rails.autoloaders.main.push_dir in its own engine.rb (nested root dirs work: app/components stays Object-namespaced, app/components/example_theme is ExampleTheme::Components).
- Core layout renders the engine view (title, chrome, body).
- Propshaft serves the engine's committed app/assets/builds/example_theme/theme.css (asset path is fingerprinted, GET returns text/css with the file's bytes).
- Engine config/locales/en.yml loads.
- Engine config/routes.rb draws into the host routes (fixture-only proof route; real extensions have no controllers or routes).

Finding for WP 2.5: the engine's locale file is placed BEFORE core's in I18n.load_path (index 363 vs core en.yml at 367), so core wins on shared keys. Overriding core strings from an extension needs the engine to append its override file after boot (engine-side, no core change).

Gate:
```
RAILS_ENV=test bundle exec rspec spec/requests/extensions/example_theme_spec.rb
6 examples, 0 failures
RAILS_ENV=test bundle exec rspec spec/i18n_keys_spec.rb spec/requests/public/sites_spec.rb spec/requests/public/custom_theme_spec.rb spec/requests/extensions
20 examples, 0 failures
bundle exec rubocop spec/fixtures/extensions spec/requests/extensions spec/rails_helper.rb
8 files inspected, no offenses (after -A)
```

### DIFF
--- a/spec/fixtures/extensions/example_theme/app/assets/builds/example_theme/theme.css
+++ b/spec/fixtures/extensions/example_theme/app/assets/builds/example_theme/theme.css
@@ -1,0 +1,7 @@
+/* Committed, prebuilt stylesheet for the fixture extension (D4). */
+:root {
+	--color-primary: #123456;
+}
+.example-theme-home {
+	border: 1px solid var(--color-primary);
+}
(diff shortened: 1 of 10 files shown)


kimadactyl:
Merged into feature/td-extension locally (merge commit on the integration branch).

PR #3480, opened 2026-09-03
Tag-only site scoping and SQL-side partner sets in the queries

kimadactyl:
Extracted from #3436 so the SQL change every site runs can be reviewed on its own (#3368 D7 and D24).

- A site defined only by partnership tags, with no neighbourhoods, used to resolve to no partners and no events. PartnersQuery now scopes by tags alone in that case; sites with neighbourhoods keep the tag-and-neighbourhood behaviour.
- EventsQuery keeps the tagged-site partner set in SQL (organiser, place, and the 2024 legacy venue match as subqueries) instead of loading every partner with its address on each events request. A spec captures SQL and fails on the old code.
- EventsQuery gains an optional tag_id filter, unused until the region filter lands.

Specs: queries plus the public events and partners request specs, 190 examples, 0 failures.

Worth timing against production-sized data before deploy: the venue match is an EXISTS over partners joined to addresses.

### DIFF
--- a/app/queries/events_query.rb
+++ b/app/queries/events_query.rb
@@ -55,18 +55,21 @@ class EventsQuery
   # @param place [Partner] filter to events at this place
   # @param organiser_or_place [Partner] filter to events by OR at this organiser
   # @param neighbourhood_id [Integer] filter to events in this neighbourhood
+  # @param tag_id [Integer] filter to events whose organiser or place carries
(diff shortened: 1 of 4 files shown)


kimadactyl:
Superseded: everything here is in #3436, which has since moved on (hosted-at-partner events on neighbourhood sites, subquery tidy-up). Review the query changes there.

Issue #3341, opened 2026-07-15
Flaky system tests (Selenium/headless-Chrome) on CI — tracking + investigation log


kimadactyl:
## Summary

Intermittent **system-test** (Selenium/headless-Chrome) failures on CI. They surface as different tests failing on different runs, are timing/resource-sensitive, and have **not** been reproducible locally or in a controlled CI batch. This issue is a running log — **append new occurrences** (test + seed + run link) as they happen so we can spot a pattern.

## Observed occurrences

| Date (UTC) | Run | Commit | Seed | Test | Symptom |
|---|---|---|---|---|---|
| 2026-07-15 13:08 | [29417996280](https://github.com/geeksforsocialchange/PlaceCal/actions/runs/29417996280) | `a0262d7a` (main) | 32142 | `Partner Save Bar › shows Back, Save, and Continue buttons on middle tabs` | `input.tab[aria-label='📍 Location']:checked` not found within 5s after clicking the tab |
| 2026-07-15 13:58 | [29421522430](https://github.com/geeksforsocialchange/PlaceCal/actions/runs/29421522430) | `6106548c` (main) | 60049 | `User Invitation Flow › allows admin to invite a user who can then set their password` | `Net::ReadTimeout with #<TCPSocket:(closed)>` while fetching `Capybara.current_session.server.port` on the **first** action |
| 2026-07-21 14:39 | [29839785424](https://github.com/geeksforsocialchange/PlaceCal/actions/runs/29839785424) | `3ae41b3` (PR #3351) | 48876 | `Admin Users › stacked list selectors on users form › allows selecting partners and partnerships` | `Capybara::ElementNotFound: Unable to find link "Admin User"` at `click_link full_name` (users_spec.rb:46, inside `find_element_with_retry`) |
| 2026-07-21 01:19 | [29792599760](https://github.com/geeksforsocialchange/PlaceCal/actions/runs/29792599760) | `renovate/actions-checkout-7.x` (PR #3350) | 8356 | `Admin Tags › creating and editing FacilityTag › allows creating and editing a Facility tag` | `assert_has_flash`: expected css `[role='alert'].alert-success, .flashes .alert-success` — no matches (tags_spec.rb:106) |
[...]

PR #3482, opened 2026-09-03
Site pages: per-site static content at /:slug with admin CRUD

kimadactyl:
Extracted from #3436 so the catch-all route and the new model can be reviewed on their own (#3368 D5 and D14).

- `Page` belongs to a Site: title, slug, markdown body rendered through the existing HtmlRenderCache sanitiser, position, show in nav, published flag. Unique per site and slug.
- Admin CRUD under the admin subdomain, scoped by PagePolicy: site admins manage pages on their own sites, root sees everything. A site admin of several sites can move a page between them; other sites are refused.
- Public route `/:slug` appended after all other routes in an initializer (so reload does not double-define it), html only, 404 for drafts and for the nationwide directory. Reserved slugs are derived from the live route set; `privacy` is the one core route a page may replace.
- Migration creates `pages` with a unique index on site and slug.

Specs: page model, policy, admin and public request specs; requests, models and policies suites 1168 examples, 0 failures on this branch. Two admin Cucumber scenarios are tagged wip (a Turbo click is swallowed when they follow another scenario, #3341); the same paths are covered by request specs.

Nothing here depends on the theme system. The derived site navigation that lists in-nav pages arrives with the rest of #3436.

### DIFF
--- a/app/controllers/admin/pages_controller.rb
+++ b/app/controllers/admin/pages_controller.rb
@@ -1,10 +1,16 @@
 # frozen_string_literal: true
 
 module Admin
+  # Serves both the admin dashboard (#home, the admin root) and CRUD for a
+  # Site's static content pages (#3368, WP 1.1). The dashboard actions predate
+  # the Page model and are exempt from Pundit's verification callbacks.
   class PagesController < Admin::ApplicationController
-    skip_after_action :verify_authorized
-    skip_after_action :verify_policy_scoped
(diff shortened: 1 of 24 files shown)


kimadactyl:
Closing: Kim decided static page content stays in the theme repo rather than a core Page model. The theme will render its own markdown through views, with a small core slot for page paths. The branch is kept for reference.

